PR #2, opened 2019-11-05
Creation of initial classes

wbclark:


spec/classes/config_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
```
ekohl: I have moved away from testing all private classes. This is a nice idea, but with the current development practices it simply doesn't make sense. PDK has some stupid defaults.

ekohl:
Oh, and because of https://puppet.com/docs/pdk/1.x/pdk_install.html#pdk-analytics I refuse to use it, even if it has an opt out.

spec/default_facts.yml
```diff
@@ -0,0 +1,8 @@
+# Use default_module_facts.yml for module specific facts.
```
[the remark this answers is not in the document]
wbclark: Removed, thanks

manifests/init.pp
```diff
@@ -0,0 +1,22 @@
...
+
+  class { selinux:
+    mode => 'permissive',
+  }
+
+  include ::redis
+
+  class { '::redis':
+    bind => '127.0.0.1',
+    manage_repo => true,
```
ehelms: We will want to set this to false, and configure to use the Redis in the SCL eventually, for now EPEL is fine:

http://mirror.centos.org/centos/7/sclo/x86_64/rh/rh-redis5/

manifests/init.pp
```diff
@@ -0,0 +1,22 @@
...
+#   include pulpcore
+class pulpcore {
+
+  include selinux
+
+  class { selinux:
+    mode => 'permissive',
+  }
+
+  include ::redis
```
ehelms: I think we'll want to move this to a `redis.pp` most likely within this module just for organization purposes.

manifests/init.pp
```diff
@@ -0,0 +1,22 @@
+# @summary A short summary of the purpose of this class
+#
+# A description of what this class does
+#
+# @example
+#   include pulpcore
+class pulpcore {
+
+  include selinux
```
ehelms: Something currently failing if SELinux is on?

manifests/database.pp
```diff
@@ -0,0 +1,15 @@
...
+  class { 'postgresql::globals':
+    version => '10',
+    client_package_name => 'rh-postgresql10-postgresql-syspaths',
+    server_package_name => 'rh-postgresql10-postgresql-server-syspaths',
+    contrib_package_name => 'rh-postgresql10-postgresql-contrib-syspaths',
+    service_name => 'postgresql_pulpcore',
+    datadir => '/var/lib/pgsql-pulpcore/data',
+    confdir => '/var/lib/pgsql-pulpcore/data',
+    bindir => '/usr/bin',
+
```
ehelms: I think this is missing a closing brace

manifests/firewalld.pp
```diff
@@ -0,0 +1,14 @@
...
+
+  include ::firewalld
+
+  firewalld_service { 'allow http service':
+    ensure => 'present',
+    service => 'http',
+    zone => 'public',
+  }
+
+}
```
ehelms: I'm not sure this is needed. We've not had to use firewalld rules in any other modules.

manifests/preinstall.pp
```diff
@@ -0,0 +1,14 @@
...
+# A description of what this class does
+#
+# @example
+#   include pulpcore::preinstall
+class pulpcore::preinstall {
+
+  include pulpcore::selinux
+  include pulpcore::redis
+  include pulpcore::database
+  include pulpcore::firewalld
```
ehelms: I assume this will drop away before its all said and done.

manifests/preinstall.pp
```diff
@@ -0,0 +1,14 @@
+# @summary A short summary of the purpose of this class
+#
+# A description of what this class does
+#
+# @example
+#   include pulpcore::preinstall
+class pulpcore::preinstall {
+
+  include pulpcore::selinux
```
ehelms: I assume this will drop away before its all said and done.

manifests/database.pp
```diff
@@ -0,0 +1,15 @@
+# sets up separate postgresql instance for pulpcore
+# @api private
+class pulpcore::database {
+
+  class { 'postgresql::globals':
```
ekohl: Generally this is not a good idea. This will break others if they set the globals. In the installer we set this via Hiera.

manifests/database.pp
```diff
@@ -0,0 +1,15 @@
...
+    version => '10',
+    client_package_name => 'rh-postgresql10-postgresql-syspaths',
+    server_package_name => 'rh-postgresql10-postgresql-server-syspaths',
+    contrib_package_name => 'rh-postgresql10-postgresql-contrib-syspaths',
+    service_name => 'postgresql_pulpcore',
```
ekohl: This implies you want to run a postgres instance just for pulp, but the module doesn't support that.

manifests/redis.pp
```diff
@@ -0,0 +1,13 @@
+# deploys redis server on 127.0.0.1
+# should use https://github.com/voxpupuli/puppet-redis for now
+# @api private
+class pulpcore::redis {
+
+  include ::redis
+
+  class { '::redis':
```
ekohl: This doesn't work. You can either `include redis` or define the class here.

manifests/redis.pp
```diff
@@ -0,0 +1,13 @@
...
+    bind => '127.0.0.1',
+    manage_repo => true,
```
ekohl: In general we don't want to manage repos or at least make it a parameter so we can disable it for users who manage their own repos, for example via Pulp but also downstream where we can't enable EPEL.

metadata.json
```diff
@@ -0,0 +1,64 @@
...
+      ]
+    }
+  ],
+  "requirements": [
+    {
+      "name": "puppet",
+      "version_requirement": ">= 4.10.0 < 7.0.0"
+    }
+  ],
+  "pdk-version": "1.14.1",
```
ekohl: You can drop PDK here and the template parts.

metadata.json
```diff
@@ -0,0 +1,64 @@
...
+      "operatingsystem": "Fedora",
+      "operatingsystemrelease": [
+        "29"
+      ]
+    }
+  ],
+  "requirements": [
+    {
+      "name": "puppet",
+      "version_requirement": ">= 4.10.0 < 7.0.0"
```
ekohl: Our modules only support `>= 5.5.8`

manifests/firewalld.pp
wbclark: I included it because the ansible installer does:

   https://github.com/pulp/ansible-pulp/blob/master/roles/pulp-webserver/tasks/firewalld.yml

But we can remove it if it's not needed here

manifests/preinstall.pp
wbclark: I included this based on the following task from the ansible-pulp installer:

https://github.com/pulp/ansible-pulp/blob/86a19f7bf4869cbe55d4850c4d2d53d33e1ee86b/roles/pulp/tasks/install.yml#L49-L53

manifests/install.pp
```diff
@@ -0,0 +1,22 @@
...
+#   include pulpcore::install
+class pulpcore::install {
+
+  ensure_packages(['package1', 'package2', 'package3'], {
+    ensure   => present,
+    provider => 'pip3',
+    require  => [ Package['python3-pip'], ],
+  })
+
+  file { '/etc/httpd/conf.d/pulp-vhost.conf':
```
ekohl: We always use the [apache](https://github.com/puppetlabs/puppetlabs-apache) module. It provides [apache::vhost](https://github.com/puppetlabs/puppetlabs-apache/blob/master/REFERENCE.md#apachevhost). https://github.com/theforeman/puppet-pulp/blob/master/manifests/crane/apache.pp is probably a good minimal example though https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/reverse_proxy.pp is probably closer to how you'll deploy this.

I'd recommend putting the apache configuration in its own file `manifests/apache.pp`.

ehelms: I think this as a first pass, especially as @wbclark is learning puppet at the same time is good for just getting things to work. We discussed then layering in the next level of enhancements as he goes.

I will add that we will more likely be tackling this via something similar to (https://github.com/theforeman/puppet-katello/blob/master/manifests/application.pp#L75-L77) given we will need to add reverse proxy statements for Pulp to the Foreman vhost definition since it controls 443.

manifests/service.pp
```diff
@@ -0,0 +1,31 @@
+# @summary A short summary of the purpose of this class
+#
+# A description of what this class does
+#
+# @example
+#   include pulpcore::service
+class pulpcore::service {
+  file { '/lib/systemd/system/pulpcore-api.service':
```
ekohl: We use the [systemd module](https://github.com/camptocamp/puppet-systemd) in various repositories to define these. It automatically handles all the service related parts: defining the unit file, reloading system and managing the service itself:

```puppet
systemd::unit_file { 'pulpcore-api.service':
  content => template('pulpcore/pulpcore-api.service.erb'),
  active  => true,
  enable  => true,
}
```

metadata.json
```diff
@@ -0,0 +1,71 @@
...
+    }
+  ],
+  "operatingsystem_support": [
+    {
+      "operatingsystem": "CentOS",
+      "operatingsystemrelease": [
+        "7"
+      ]
+    },
+    {
```
ekohl: I'd not be too pro-active with listing support. I'd just list Red Hat 7 and CentOS 7.

templates/pulpcore-api.service.erb
```diff
@@ -0,0 +1,35 @@
+[Unit]
+Description=Pulp WSGI Server
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Environment="DJANGO_SETTINGS_MODULE=pulpcore.app.settings"
+Environment="PULP_SETTINGS=/etc/pulp/settings.py"
+Environment="PATH=/usr/local/lib/pulp/bin:{'msg': 'skipped, running in check mode', 'skipped': True, 'failed': False, 'changed': False}"
```
ekohl: This doesn't look correct

manifests/install.pp
ekohl:
```puppet
$pulp_api_path = '/pulp/api'
$pulp_api_url = "http://${pulp::pulp_api_host}:%{pulp::pulp_api_port}/pulp/api"
$pulp_content_path = '/pulp/content'
$pulp_content_url = "http://${pulp::pulp_content_host}:%{pulp::pulp_content_port}/pulp/content"

include apache
apache::vhost { 'pulp':
  servername => $facts['fqdn'],
  port       => 80,
  docroot    => $pulp::pulp_webserver_static_dir,
  proxy_pass => [        
    {
      'path'         => $pulp_api_path,
      'url'          => $pulp_api_url,
      'reverse_urls' => [$pulp_api_path, $pulp_api_url],
    },
    {
      'path'         => $pulp_content_path,
      'url'          => $pulp_content_url,
      'reverse_urls' => [$pulp_content_path, $pulp_content_url],
    },
  ],
}

```

spec/classes/pulpcore_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'spec_helper'
+
+describe 'pulpcore' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+
+      it { is_expected.to compile }
```
ekohl:
```suggestion
      it { is_expected.to compile.with_all_deps }
```

manifests/install.pp
ehelms: Assume this is just for early testing purposes? @ekohl

ekohl: Bit of a lack of context, but I explained how other modules do this to @wbclark and I needed to write it down somewhere. Now it's a convenient reference. It's untested, but this should be the equivalent of the vhost file with the benefit that it can be used together with the rest of the installer modules. In the future it should also be easy to add SSL to it, but that's for a later session to explain :)

templates/pulpcore-api.service.erb
```diff
@@ -0,0 +1,35 @@
...
+Description=Pulp WSGI Server
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+Environment="DJANGO_SETTINGS_MODULE=pulpcore.app.settings"
+Environment="PULP_SETTINGS=/etc/pulp/settings.py"
+Environment="PATH=/usr/local/lib/pulp/bin:"
+User=pulp
+PIDFile=/run/pulpcore-api.pid
```
ekohl: I suspect this is a copy of what the ansible installer has and that's fine but https://docs.gunicorn.org/en/stable/deploy.html#systemd suggests deploying on a unix socket. It also suggests pid files aren't needed when using systemd. Probably best to keep this for the next iteration.

templates/pulpcore-resource-manager.service.erb
```diff
@@ -0,0 +1,27 @@
...
+[Service]
+Environment="DJANGO_SETTINGS_MODULE=pulpcore.app.settings"
+Environment="PULP_SETTINGS=/etc/pulp/settings.py"
+Environment="PATH=/usr/local/lib/pulp/bin:"
+User=pulp
+WorkingDirectory=/var/run/pulpcore-resource-manager/
+RuntimeDirectory=pulpcore-resource-manager
+ExecStart=/usr/local/lib/pulp/bin/rq worker \
+          -w pulpcore.tasking.worker.PulpWorker -n resource-manager@%%h \
+          --pid=/var/run/pulpcore-resource-manager/resource-manager.pid \
```
ekohl: https://python-rq.org/patterns/systemd/ also suggests to rely on systemd's `$MAINPID`.

manifests/install.pp
```diff
@@ -19,4 +19,9 @@
     content => template('pulpcore/pulp-vhost.conf.erb'),
   } 
 
+  file { '/etc/pulp/settings.py':
```
ekohl: Usually this is considered configuration and goes in config.pp. The benefit is that you know the package is installed which usually means things like the config dir exist. In this case you probably need to ensure `/etc/pulp` is created as well.

ekohl: Now that there's automated tests, you can see this actually breaks. https://travis-ci.org/theforeman/puppet-pulpcore/jobs/614129962#L1000

```
  Error: Could not set 'file' on ensure: No such file or directory @ dir_s_mkdir - /etc/pulp/settings.py20191119-974-1mqifpv.lock (file: /etc/puppetlabs/code/modules/pulpcore/manifests/install.pp, line: 23)

  Error: Could not set 'file' on ensure: No such file or directory @ dir_s_mkdir - /etc/pulp/settings.py20191119-974-1mqifpv.lock (file: /etc/puppetlabs/code/modules/pulpcore/manifests/install.pp, line: 23)
```

The reason is that `/etc/pulp` doesn't exist. In `config.pp` I already ensure this is all set up using the correct paths.

manifests/redis.pp
```diff
@@ -0,0 +1,11 @@
+# deploys redis server on 127.0.0.1
```
ekohl: In my fixes PR I merged this into `pulpcore::database` so can be dropped again.

manifests/install.pp
```diff
@@ -0,0 +1,27 @@
+# @summary Install pulpcore packages
+# Currently this is done with pip3. This should have the option to also install via RPMs.
+# @api private
+class pulpcore::install {
+
+  $system_packages = ['gcc', 'postgresql-server', 'postgresql-devel', 'python3-pip', 'python3-devel']
```
ekohl: `postgresql-server` should already be done by the postgresql server class in database.pp.

manifests/service.pp
```diff
@@ -1,19 +1,29 @@
...
+
   systemd::unit_file { 'pulpcore-resource-manager.service':
     content => template('pulpcore/pulpcore-resource-manager.service.erb'),
     active  => true,
     enable  => true,
   }
+
+  systemd::unit_file { 'pulpcore-worker@.service':
+    content => template('pulpcore/pulpcore-worker@.service.erb'),
+    active  => true,
```
ekohl: I don't think this works here because the unit is just a template. I think you need to drop the active and enable. Then you can create actual instances with the raw `service` type.

manifests/database.pp
```diff
@@ -0,0 +1,33 @@
...
+    bindir               => '/usr/bin',
+  }
+
+  include postgresql::server
+  postgresql::server::db { $pulpcore::database_name:
+    user     => $pulpcore::pulp_user,
+    password => postgresql_password($pulpcore::pulp_user, 'secret'),
+  }
+
+  exec { 'django-admin migrate --noinput':
```
ekohl: This needs an explicit `require => Postgresql::Server::Db[$pulpcore::database_name]` because it looks like it's now ran before the postgres server is up.

manifests/database.pp
```diff
@@ -0,0 +1,33 @@
...
+    path        => ['/usr/local/bin', '/usr/bin'],
+    environment => [
+      'DJANGO_SETTINGS_MODULE=pulpcore.app.settings',
+      "PULP_SETTINGS=${pulpcore::settings_file}",
+    ],
+    refreshonly => true,
```
ekohl: We should find a better way of determining whether a migration is needed but this will work for a first version.

manifests/service.pp
```diff
@@ -0,0 +1,31 @@
...
+
+  systemd::unit_file { 'pulpcore-resource-manager.service':
+    content => template('pulpcore/pulpcore-resource-manager.service.erb'),
+    active  => true,
+    enable  => true,
+  }
+
+  systemd::unit_file { 'pulpcore-worker@.service':
+    content => template('pulpcore/pulpcore-worker@.service.erb'),
+  } ->
```
ekohl: Linting doesn't like this style, but in most modules we ignore this. You can ignore it, but you can also add an explicit require as a property to the service. I think that needs to happen anyway: `require => [Systemd::Unit_file['pulpcore-worker@.service'], Class['systemd::systemctl::daemon_reload']]`

ekohl: With the require you can now drop the `->` here. I'd also add a newline after it.

manifests/database.pp
```diff
@@ -0,0 +1,34 @@
...
+    require => Postgresql::Server::Db[$pulpcore::database_name]
```
ekohl:
```suggestion
    require => Postgresql::Server::Db[$pulpcore::database_name],
```

manifests/service.pp
```diff
@@ -0,0 +1,32 @@
...
+  service { ['pulpcore-worker@1', 'pulpcore-worker@2']:
+    ensure  => running,
+    enable  => true,
+    require => [Systemd::Unit_file['pulpcore-worker@.service'], Class['systemd::systemctl::daemon_reload']]
```
ekohl:
```suggestion
    require => [Systemd::Unit_file['pulpcore-worker@.service'], Class['systemd::systemctl::daemon_reload']],
```

spec/acceptance/basic_spec.rb
```diff
@@ -28,4 +28,10 @@
   describe port(80) do
     it { is_expected.to be_listening }
   end
+
+  describe command("curl -s http://centos7.mercury.example.com/pulp/api/v3/status/ -w '%{response_code}' -o /dev/null") do
```
ekohl: You can use `localhost` instead of the hostname (which will differ for every system). The alternative is to use `fact('fqdn')` but for now `localhost` should do as well.

ehelms:
Looks like this needs a rebase. @ekohl @wbclark is this no longer a WIP and ready to be merged once conflicts are addressed?

metadata.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "puppet-pulpcore",
+  "version": "0.1.0",
+  "author": "William Clark",
+  "summary": "",
```
ekohl: While rebasing, could you add a short summary?

metadata.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "puppet-pulpcore",
```
ekohl:
```suggestion
  "name": "theforeman-pulpcore",
```

metadata.json
```diff
@@ -0,0 +1,50 @@
...
+  "version": "0.1.0",
+  "author": "William Clark",
```
ekohl: This should match the user we're using to publish:
```suggestion
  "author": "theforeman",
```

metadata.json
```diff
@@ -0,0 +1,50 @@
...
+  "summary": "",
+  "license": "GPL-3.0-or-later",
+  "source": "",
```
ekohl:
```suggestion
  "source": "https://github.com/theforeman/puppet-pulpcore",
```

metadata.json
```diff
@@ -0,0 +1,50 @@
...
+    {
+      "operatingsystem": "CentOS",
+      "operatingsystemrelease": [
+        "7"
+      ]
+    },
+    {
+      "operatingsystem": "RedHat",
+      "operatingsystemrelease": [
+        "8"
```
ekohl: Probably better to also make this 7 to match CentOS.
```suggestion
        "7"
```

ekohl:
You probably need to update your master branch before rebasing because there's still conflicts.

manifests/database.pp
```diff
@@ -0,0 +1,34 @@
+# sets up separate postgresql instance for pulpcore
```
ekohl:
```suggestion
# Set up the PostgreSQL and Redis databases
```

manifests/apache.pp
```diff
@@ -0,0 +1,26 @@
+class pulpcore::apache {
```
ekohl:
```suggestion
# Configure an Apache vhost
# @api private
class pulpcore::apache {
```

.fixtures.yml
```diff
@@ -0,0 +1,11 @@
+fixtures:
+  repositories:
+    apache:           'https://github.com/puppetlabs/puppetlabs-apache'
+    concat:           'https://github.com/puppetlabs/puppetlabs-concat'
+    postgresql:       'https://github.com/puppetlabs/puppetlabs-postgresql'
+    redis:            'https://github.com/voxpupuli/puppet-redis'
+    stdlib:           'https://github.com/puppetlabs/puppetlabs-stdlib'
+    systemd:          'https://github.com/camptocamp/puppet-systemd'
+    augeas_core:
```
ekohl: Nit: I like to keep lists sorted so ideally this would be between `apache` and `concat`.

PR #3, opened 2019-11-15
Add acceptance test skeleton

ekohl:
This runs beaker acceptance tests on CentOS 7. It also updates the Gemfile per the latest modulesync.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,0 +1,4 @@
+---
+.travis.yml:
+  beaker_sets:
+    - centos7-64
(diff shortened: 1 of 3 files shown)


ehelms:
Should this pass out of the gate?

ekohl:
No, technically this should be part of https://github.com/theforeman/puppet-pulpcore/pull/2 and https://gist.github.com/ekohl/add82c066b56fcbfcd0ea981d45c8490 should be there as well.

wbclark:
I've updated #2 to include this and https://gist.github.com/ekohl/add82c066b56fcbfcd0ea981d45c8490

Thanks @ekohl

PR #8, opened 2019-11-21
fixed default params

wbclark:
In this PR, I have:

1. Changed default user, group, and paths to use 'pulp' instead of 'pulpcore' as required  by the migration tool

2. Shorted parameter names by removing redundant 'pulp_' prefixes

3. Commented parameters in manifests/init.pp

4. Added random generation of DB password

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -14,9 +14,9 @@ class pulpcore::database {
   }
 
   include postgresql::server
-  postgresql::server::db { $pulpcore::database_name:
-    user     => $pulpcore::pulp_user,
-    password => postgresql_password($pulpcore::pulp_user, 'secret'),
+  postgresql::server::db { $pulpcore::postgresql_db_name:
+    user     => $pulpcore::user,
+    password => postgresql_password($pulpcore::user, $pulpcore::postgresql_db_password),
   }
 
   exec { 'django-admin migrate --noinput':
@@ -26,7 +26,7 @@ class pulpcore::database {
       "PULP_SETTINGS=${pulpcore::settings_file}",
     ],
     refreshonly => true,
-    require     => Postgresql::Server::Db[$pulpcore::database_name],
+    require     => Postgresql::Server::Db[$pulpcore::postgresql_db_name],
   }
 
   include redis
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -1,24 +1,59 @@
-# @summary A short summary of the purpose of this class
+# Manage your next generation Pulp server
 #
-# A description of what this class does
+# @param cache_dir
+#   Pulp cache directory
+#
+# @param config_dir
+#   Pulp configuration directory
+#
+# @param user
+#   Pulp user
+#
+# @param group
+#   Pulp user group
+#
+# @param user_home
+#   Pulp user home directory
+#
+# @param api_host
+#   API service host
+#
+# @param api_port
+#   API service port
+#
+# @param content_host
+#   Content service host
+#
+# @param content_port
+#   Content service port
+#
+# @param webserver_static_dir
+#   Directory for Pulp webserver static content
+#
+# @param postgresql_db_name
+#   Name of Pulp database
+#
+# @param postgresql_db_password
+#   Password of Pulp database
 #
 # @example
 #   include pulpcore
 class pulpcore (
-  Stdlib::Absolutepath $pulp_cache_dir = '/var/lib/pulpcore/tmp',
-  Stdlib::Absolutepath $pulp_config_dir = '/etc/pulpcore',
-  String $pulp_user = 'pulpcore',
-  String $pulp_group = 'pulpcore',
-  Stdlib::Absolutepath $pulp_user_home = '/var/lib/pulpcore',
-  Stdlib::Host $pulp_api_host = '127.0.0.1',
-  Stdlib::Port $pulp_api_port = 24817,
-  Stdlib::Host $pulp_content_host = '127.0.0.1',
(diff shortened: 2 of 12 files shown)


manifests/init.pp
```diff
@@ -1,24 +1,49 @@
...
+  String $user = 'pulp',
+  String $group = 'pulp',
+  Stdlib::Absolutepath $user_home = '/var/lib/pulp',
+  Stdlib::Host $api_host = '127.0.0.1',
+  Stdlib::Port $api_port = 24817,
+  Stdlib::Host $content_host = '127.0.0.1',
+  Stdlib::Port $content_port = 24816,
+  Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $database_name = 'pulpcore',
+  String $database_password = extlib::cache_data('foreman_cache_data', 'db_password', extlib::random_password(32)),
```
ekohl: `foreman_cache_data` in this case is the namespace. I think this is a direct copy of foreman which means you'll be using the same password. I'd recommend using `pulpcore_cache_data` as a namespace.
```suggestion
  String $database_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
```

manifests/init.pp
```diff
@@ -1,24 +1,49 @@
...
+  Stdlib::Absolutepath $config_dir = '/etc/pulp',
+  String $user = 'pulp',
+  String $group = 'pulp',
+  Stdlib::Absolutepath $user_home = '/var/lib/pulp',
+  Stdlib::Host $api_host = '127.0.0.1',
+  Stdlib::Port $api_port = 24817,
+  Stdlib::Host $content_host = '127.0.0.1',
+  Stdlib::Port $content_port = 24816,
+  Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $database_name = 'pulpcore',
```
ekohl: I'm wondering if we should prefix it with `postgresql_` as well. I suspect we'll also have redis equivalents. Then `postgresql_database_name` and `redis_database` will be clearer

manifests/init.pp
```diff
@@ -1,24 +1,49 @@
-# @summary A short summary of the purpose of this class
+# Manage your next generation Pulp server
 #
-# A description of what this class does
+# === Parameters:
```
ekohl: You are now converting it to rdoc rather than puppet-strings. That will provide the option to hide parameters if we expose it in the installer, but let's start with puppet-strings style documentation:

```
# @param cache_dir
#   Pulp cache directory
```

You can also leave out the defaults. Try running `bundle exec rake strings:generate:reference` and what `REFERENCE.md` contains. That will be the documentation we'll add on release (but keep out of git).

manifests/init.pp
wbclark: Done, thanks.

manifests/init.pp
wbclark: Agreed and done.

manifests/database.pp
```diff
@@ -14,9 +14,9 @@
   }
 
   include postgresql::server
-  postgresql::server::db { $pulpcore::database_name:
-    user     => $pulpcore::pulp_user,
-    password => postgresql_password($pulpcore::pulp_user, 'secret'),
+  postgresql::server::db { $pulpcore::postgresql_database_name:
+    user     => $pulpcore::user,
```
ekohl: For remote DB support we probably need a `postgresql_database_user` parameter. Fine to keep for a later PR though

manifests/init.pp
```diff
@@ -1,24 +1,49 @@
...
+  Stdlib::Absolutepath $config_dir = '/etc/pulp',
+  String $user = 'pulp',
+  String $group = 'pulp',
+  Stdlib::Absolutepath $user_home = '/var/lib/pulp',
+  Stdlib::Host $api_host = '127.0.0.1',
+  Stdlib::Port $api_port = 24817,
+  Stdlib::Host $content_host = '127.0.0.1',
+  Stdlib::Port $content_port = 24816,
+  Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
+  String $postgresql_database_name = 'pulpcore',
```
ekohl: Bikeshedding: is `postgresql_db_name` clear enough?

templates/settings.py.erb
```diff
@@ -6,16 +6,16 @@ SECRET_KEY = "TODO"
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql',
-        'NAME': '<%= scope['pulpcore::database_name'] %>',
-        'USER': '<%= scope['pulpcore::pulp_user'] %>',
-        'PASSWORD': 'secret',
+        'NAME': '<%= scope['pulpcore::postgresql_database_name'] %>',
+        'USER': '<%= scope['pulpcore::user'] %>',
+	'PASSWORD': '<%= scope['pulpcore::postgresql_database_password'] %>',
```
ekohl: Looks like this uses a tab for indenting while the rest uses spaces.
```suggestion
        'PASSWORD': '<%= scope['pulpcore::postgresql_database_password'] %>',
```

manifests/init.pp
wbclark: Done. Thanks for the explanation.

manifests/init.pp
wbclark: I think so. I'll use the same convention for `postgresql_db_password` as well.

manifests/database.pp
wbclark: Opened issue #10 for this

PR #30, opened 2019-12-11
Add file and container plugins

ekohl:


ekohl:
I honestly don't know. Looks like it's true for [pulp-file](https://pulp-file.readthedocs.io/en/latest/installation.html) but not [pulp-container](https://pulp-container.readthedocs.io/en/latest/installation.html#configure-required-settings) which needs setup for the tokens. I think we can initially skip that so I've modified it to just disable it. Not entirely sure if the way I deal with the concat resource  name and path works. Should probably write some acceptance test for that.

PR #43, opened 2019-12-11
Introduce a pulpcore::admin define

ekohl:
This is essentially a wrapper around django-admin with the right settings.

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -1,0 +1,38 @@
+# @summary Run a django-admin command
+#
+# @param command
+#   The command to run
+#
+# @param refreshonly
+#   The command should only be run as a refresh mechanism for when a dependent
+#   object is changed.
+#
+# @param unless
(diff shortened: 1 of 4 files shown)


manifests/admin.pp
```diff
@@ -0,0 +1,38 @@
...
+# @see exec
+define pulpcore::admin(
+  String $command = $title,
+  Boolean $refreshonly = false,
+  Optional[String] $unless = undef,
+  Array[Stdlib::Absolutepath] $path = ['/usr/bin'],
+  Stdlib::Absolutepath $pulp_settings = $pulpcore::settings_file,
+) {
+  File <| title == $pulp_settings |>
+  -> exec { "django-admin ${title}":
```
wbclark: So once @evgeni has modified the RPM design so that django 3's `django-admin` does not conflict with django 2, we can update the `exec` statement in a single place and have it be correct for the whole module.

I really like this approach. LGTM.

PR #44, opened 2019-12-11
Improve settings.

### LINKED ISSUE #31: Secret key handling
We need to generate a secure key:
https://github.com/theforeman/puppet-pulpcore/blob/fc9d5d05bf39d0f05713aca8500f6ad960f13c0b/templates/settings.py.erb#L5

### LINKED ISSUE #32: Drop Ansible settings

### LINKED ISSUE #34: Update Redis database number to '8'

wbclark:
supporting random SECRET_KEY, dropping ansible settings in settings.py, and setting redis database number to '8'

fixes #31
fixes #32
fixes #34

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -39,6 +39,12 @@
 # @param postgresql_db_password
 #   Password of Pulp database
 #
+# @param django_secret_key
+#   SECRET_KEY for Django
+#
+# @param redis_db
+#   Redis DB number to use. By default, Redis supports a DB number of 0 through 15.
+#
 # @example
 #   include pulpcore
 class pulpcore (
@@ -55,6 +61,8 @@ class pulpcore (
   Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
+  String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
+  Integer[0] $redis_db = 8,
 ) {
 
   $settings_file = "${config_dir}/settings.py"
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -55,6 +61,8 @@
   Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
+  String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
+  Integer[0, 15] $redis_db_number = 8,
```
ekohl: I'd remove the upper limit since Redis can be configured to support more than 16 databases. The warning in the documentation is fine to keep and IMHO sufficient.

templates/settings.py.erb
```diff
@@ -1,8 +1,6 @@
-CONTENT_HOST = "<%= @fqdn %>"
-ANSIBLE_API_HOSTNAME = "<%= @fqdn %>"
-ANSIBLE_CONTENT_HOSTNAME = "https://<%= @fqdn %>/pulp_ansible/content/"
+CONTENT_HOST = "<%= scope['pulpcore::servername'] %>"
```
ekohl: Just a nit: can we get `CONTENT_HOST` and `CONTENT_ORIGIN` together?

manifests/init.pp
```diff
@@ -39,6 +39,12 @@
 # @param postgresql_db_password
 #   Password of Pulp database
 #
+# @param django_secret_key
+#   SECRET_KEY for Django
+#
+# @param redis_db_number
```
ekohl: I'd drop the `_number` part since a DB is always a number in Redis. With the datatype this is also enforced.

manifests/init.pp
wbclark: @ekohl ok, that works for me.

templates/settings.py.erb
wbclark: sure, this makes sense and I'm happy to do it

manifests/init.pp
wbclark: Point taken and done.

ekohl:
General note: we try to keep the commit title and PR title as a useful summary. In this case something as vague as `Improve settings` is already useful. This is to keep `git log` useful. In our modules we also generate the changelog from the PR titles.

ekohl:
I've modified the PR message to automatically close the issues when this is merged. Currently Travis is still running but :+1: once it passes.

Issue #39, opened 2019-12-11
Update to use django-admin solution for pulpcore


ehelms:
The `django-admin` from python3-django will likely be renamed or put somewhere different.

wbclark:
With https://github.com/theforeman/puppet-pulpcore/pull/43 merged we now have a common wrapper for all `django-admin` commands. When it is updated in the RPMs, it will only require a single update to the `admin.pp` module here.

ekohl:
There's actually 2:
https://github.com/theforeman/puppet-pulpcore/blob/f4f177c601d63c22fd0e2fbb992a9b327cc5dbe1/manifests/admin.pp#L29
https://github.com/theforeman/puppet-pulpcore/blob/f4f177c601d63c22fd0e2fbb992a9b327cc5dbe1/manifests/database.pp#L12

PR #46, opened 2019-12-13
use python3-django-admin instead of unversioned django-admin

evgeni:
Fixes: #39

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -1,4 +1,4 @@
-# @summary Run a django-admin command
+# @summary Run a python3-django-admin command
 #
 # @param command
 #   The command to run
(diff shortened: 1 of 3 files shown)


ehelms:
From discussions this https://github.com/theforeman/puppet-pulpcore/blob/master/spec/spec_helper_acceptance.rb#L36 can get updated to http://koji.katello.org/releases/yum/katello-nightly/pulpcore/el7/x86_64/

wbclark:
> From discussions this https://github.com/theforeman/puppet-pulpcore/blob/master/spec/spec_helper_acceptance.rb#L36 can get updated to http://koji.katello.org/releases/yum/katello-nightly/pulpcore/el7/x86_64/

This will be resolved shortly by https://github.com/theforeman/puppet-pulpcore/pull/47

PR #53, opened 2019-12-18
Update pulpcore worker systemd naming

ianballou:
Related Pulp 3 PRs that are merged:

1: https://github.com/pulp/ansible-pulp/pull/197
2: https://github.com/pulp/pulpcore/pull/427
3: https://github.com/pulp/pulplift/pull/59

### DIFF
--- a/templates/pulpcore-worker@.service.erb
+++ b/templates/pulpcore-worker@.service.erb
@@ -14,7 +14,6 @@ WorkingDirectory=/var/run/pulpcore-worker-%i/
 RuntimeDirectory=pulpcore-worker-%i
 ExecStart=/usr/bin/rq worker \
           -w pulpcore.tasking.worker.PulpWorker \
-          -n reserved-resource-worker-%i@%%h \
           --pid=/var/run/pulpcore-worker-%i/reserved-resource-worker-%i.pid \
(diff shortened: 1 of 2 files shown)


templates/pulpcore-worker@.service.erb
```diff
@@ -14,7 +14,6 @@ WorkingDirectory=/var/run/pulpcore-worker-%i/
 RuntimeDirectory=pulpcore-worker-%i
 ExecStart=/usr/bin/rq worker \
           -w pulpcore.tasking.worker.PulpWorker \
-          -n reserved-resource-worker-%i@%%h \
```
wbclark: Did you mean to remove the name specification entirely here, rather than just removing the hostname from the worker name? Shouldn't we keep names like `reserved-resource-worker-1` and `reserved-resource-worker-2` ?

ianballou: @wbclark the name was removed as part of the changes. I'm mirroring this template: https://github.com/pulp/ansible-pulp/blob/master/roles/pulp-workers/templates/pulpcore-worker@.service.j2

wbclark: Great, thanks for the explanation, and the contribution. Will ack and merge.

PR #55, opened 2020-01-10
fixes #28655 - support fetching files via /pulp/isos with pulp3

wbclark:
should work with https://github.com/theforeman/puppet-foreman_proxy_content/pull/228 to ReverseProxy /pulp/isos to /pulp/content


wbclark:
Closing this to focus on making the change within puppet-FPC, since the /pulp/isos path is unique to Katello.

PR #56, opened 2020-01-14
Fixes #28654 - support client cert auth with pulp3

wbclark:
Fixes #28654

This patch updates settings.py to include REMOTE_USER_ENVIRONMENT_NAME = 'HTTP_REMOTE_USER'. This, along with a change to the vhost in puppet-foreman_proxy_content, will be necessary to support client cert auth with pulp3.

### DIFF
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -17,3 +17,4 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
+REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'


templates/settings.py.erb
```diff
@@ -17,3 +17,4 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
+REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'
```
ehelms: This has `ENVIRON_NAME`, and your comment mentions `ENVIRONMENT_NAME`. Looks like `ENVIRON_NAME` is correct from the issue.

ekohl:
> Fixes #28654

For future reference, there's no need to repeat it if it's already in the subject line. When there's multiple PRs needed to close an issue, generally you want to use Refs on all preparations and Fixes on the last one. Otherwise it's closed automatically after the first PR. In this particular case I'd say *this* PR is the Fixes and the other one is the Refs. My reasoning is that this closes the loop.

It would be even better if this is a parameter the calling module sets the right parameter to configure it. The reason for that is that now the vhost we deploy by default is insecure since every HTTP client can pass the REMOTE_USER header

Issue #33, opened 2019-12-11
Make PULP2 mongodb settings configurable


ehelms:

ekohl:
Fixed in https://github.com/theforeman/puppet-pulpcore/pull/57.

@wbclark you can use `Fixes GH-33` in the body of the PR (not commit) to automatically close the issue on merge.

PR #57, opened 2020-01-15
Fixes #28720 - connect to mongo for content migrations

wbclark:
https://github.com/theforeman/puppet-pulpcore/issues/33 could also be addressed by expanding on this but would introduce additionally parameters that don't yet seem necessary.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -45,6 +45,36 @@
 # @param redis_db
 #   Redis DB number to use. By default, Redis supports a DB number of 0 through 15.
 #
+# @param mongo_db_name
+#   MongoDB database name for Pulp2. Used for migrating Pulp2 content
+#
+# @param mongo_db_seeds
+#   MongoDB seeds. Used for migrating Pulp2 content
+#
+# @param mongo_db_username
+#   MongoDB username for Pulp2
+#
+# @param mongo_db_password
+#   Password of MongoDB user
+#
+# @param mongo_db_replica_set
+#   MongoDB Replica Set
+#
+# @param mongo_db_ssl
+#   Use SSL for MongoDB connection
+#
+# @param mongo_db_ssl_keyfile
+#   SSL Keyfile for MongoDB
+#
+# @param mongo_db_ssl_certfile
+#   SSL Certfile for MongoDB
+#
+# @param mongo_db_verify_ssl
+#   Whether SSL verification is required for MongoDB
+#
+# @param mongo_db_ca_path
+#   CA bundle path to use for MongoDB SSL connection
+#
 # @param servername
 #   Server name of the VirtualHost in the webserver
 #
@@ -71,6 +101,16 @@ class pulpcore (
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['fqdn'],
   Optional[String] $remote_user_environ_name = undef,
+  String $mongo_db_name = 'pulp_database',
+  String $mongo_db_seeds = 'localhost:27017',
+  Optional[String] $mongo_db_username = undef,
+  Optional[String] $mongo_db_password = undef,
+  Optional[String] $mongo_replica_set = undef,
+  String $mongo_db_ssl = 'False',
+  Optional[Stdlib::Absolutepath] $mongo_db_ssl_keyfile = undef,
+  Optional[Stdlib::Absolutepath] $mongo_db_ssl_certfile = undef,
+  String $mongo_db_verify_ssl = 'True',
+  Stdlib::Absolutepath $mongo_db_ca_path = '/etc/pki/tls/certs/ca-bundle.crt',
 ) {
 
   $settings_file = "${config_dir}/settings.py"
--- a/spec/classes/plugin_migration_spec.rb
+++ b/spec/classes/plugin_migration_spec.rb
@@ -1,0 +1,40 @@
+require 'spec_helper'
+
+migration_config = <<-EOL
+PULP2_MONGODB = {
+    "name": "pulp_database",
+    "seeds": "localhost:27017",
+    "username": "",
+    "password": "",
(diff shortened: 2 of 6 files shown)


wbclark:
> IMHO these should be parameters since mongodb can be external. Not 100% sure we support that in the katello scenario, but at that level we have all the correct information. Also if you don't pass them in by default, you don't give the impression that it's configured.
> 
> The way I would design this is to create `pulpcore::plugin::migration`. That can drop in the required configuration and ensure the migration package is installed. If you want a migration to be set up, you include the plugin class with the right parameters and it's all set up correctly.

https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin.pp might need to be reworked a bit because it currently always tries to install a package by the name of `python3-pulp-${title}` but I think that is doable.

ehelms:
We do indeed support external MongoDB, so we would need this to support:

 1) connecting to an external MongoDB
 2) connecting to MongoDB with some authentication

manifests/config.pp
```diff
@@ -16,6 +16,7 @@
 
   concat::fragment { 'base':
     target  => 'pulpcore settings',
+#    content => template('pulpcore/settings.py.erb', 'pulpcore/settings-mongodb.py.erb'),
```
ekohl: Not a fan of commented code :)

templates/settings-mongodb.py.erb
```diff
@@ -0,0 +1 @@
+PULP2_MONGODB = {"name": "<%= scope['pulpcore::mongo_db_name'] %>", "seeds": "<%= scope['pulpcore::mongo_db_seeds'] %>", "ssl": <%= scope['pulpcore::mongo_db_ssl'] %>}
```
ekohl: I'd suggest so format this in multiple lines.

manifests/init.pp
```diff
@@ -66,6 +66,9 @@
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
+  String $mongo_db_name = 'pulp_database',
```
ekohl: These should be documented as `@param`

wbclark: Will do, thanks

manifests/config.pp
wbclark: Just a temporary test. I can't run them locally due to some environmental issues that I'm putting off resolving until after this is complete.

templates/settings-mongodb.py.erb
wbclark: Will do. I'm planning as well to put this back in the main `settings.py.erb` template with a conditional.

ekohl:
https://github.com/theforeman/foreman-packaging/pull/4660 should provide the plugin package.

wbclark:
Will look at moving https://github.com/theforeman/puppet-pulpcore/pull/57/files#diff-9126f4197436df08d4e311a62a5daf47R22-R26 into manifests/plugin/migration.pp

spec/classes/plugin_file_spec.rb
```diff
@@ -15,6 +15,7 @@
           is_expected.to compile.with_all_deps
           is_expected.to contain_pulpcore__plugin('file')
             .that_subscribes_to('Class[Pulpcore::Install]')
+          is_expected.to contain_pulpcore__plugin('file')
```
ekohl: This shouldn't be needed

spec/classes/plugin_container_spec.rb
```diff
@@ -15,6 +15,7 @@
           is_expected.to compile.with_all_deps
           is_expected.to contain_pulpcore__plugin('container')
             .that_subscribes_to('Class[Pulpcore::Install]')
+          is_expected.to contain_pulpcore__plugin('container')
```
ekohl: This shouldn't be needed

spec/classes/plugin_migration_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      it { is_expected.to contain_pulpcore__plugin('migration') }
+
+      context 'with pulpcore' do
+        let(:pre_condition) { 'include pulpcore' }
+
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_pulpcore__plugin('migration')
+            .that_subscribes_to('Class[Pulpcore::Install]')
+          is_expected.to contain_pulpcore__plugin('migration')
```
ekohl: This shouldn't be needed. You can chain them

ekohl:
>   Notice: /Stage[main]/Pulpcore::Database/Pulpcore::Admin[migrate --noinput]/Exec[python3-django-admin migrate --noinput]/returns: pkg_resources.DistributionNotFound: The 'importlib_metadata' distribution was not found and is required by jsonschema

@evgeni this suggests a packaging issue

evgeni:
I agree, will look into it.

spec/classes/plugin_migration_spec.rb
wbclark: @ekohl I noticed that some were not getting executed when I did this. However removing the chaining did not resolve it. Is something excluding redundant tests?

evgeni:
@ekohl packaging fix in https://github.com/theforeman/foreman-packaging/pull/4677

spec/classes/plugin_migration_spec.rb
wbclark: @ekohl after trying a few different approaches, I believe that the chaining works however the full output is suppressed. Therefore I have reverted to the chained approach.

wbclark:
Will update this further with additional options from https://github.com/pulp/pulp-2to3-migration#configuration

PR #61, opened 2020-01-30
Fixes #28904 - make REMOTE_USER_ENVIRON_NAME configurable

### LINKED ISSUE #58: configurable REMOTE_USER_ENVIRON_NAME
> Fixes #28654

It would be even better if this is a parameter the calling module sets the right parameter to configure it. The reason for that is that now the vhost we deploy by default is insecure since every HTTP client can pass the REMOTE_USER header

_Originally posted by @ekohl in https://github.com/theforeman/puppet-pulpcore/pull/56#issuecomment-574419934_

synkd:
Fixes #28904

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -48,6 +48,9 @@
 # @param servername
 #   Server name of the VirtualHost in the webserver
 #
+# @param remote_user_environ_name
+#   Django remote user environment variable
+#
 # @example
 #   include pulpcore
 class pulpcore (
@@ -66,7 +69,8 @@ class pulpcore (
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
-  Stdlib::Fqdn $servername = $facts['fqdn']
+  Stdlib::Fqdn $servername = $facts['fqdn'],
+  Optional[String] $remote_user_environ_name = undef,
 ) {
 
   $settings_file = "${config_dir}/settings.py"
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -17,4 +17,6 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
-REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'
+<% if scope['pulpcore::remote_user_environ_name'] -%>
+REMOTE_USER_ENVIRON_NAME = '<%= scope['pulpcore::remote_user_environ_name'] %>'
+<% end -%>
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -7,6 +7,7 @@ describe 'pulpcore' do
 
       it { is_expected.to compile.with_all_deps }
       it { is_expected.to contain_package('python3-pulpcore') }
+      it { is_expected.to contain_concat__fragment('base').without_content(/REMOTE_USER_ENVIRON_NAME/) }
 
       context 'with a plugin' do
         let(:pre_condition) { "pulpcore::plugin { 'myplugin': }" }


manifests/init.pp
```diff
@@ -66,7 +69,8 @@
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
-  Stdlib::Fqdn $servername = $facts['fqdn']
+  Stdlib::Fqdn $servername = $facts['fqdn'],
+  String $django_remote_user = 'HTTP_REMOTE_USER'
```
ekohl: Could you ensure a trailing comma?
```suggestion
  String $django_remote_user = 'HTTP_REMOTE_USER',
```

I'm also wondering if we should make this optional:
```suggestion
  Optional[String] $django_remote_user = undef,
```

That way we are secure by default, unless the vhost has been set up to properly set it in a secure fashion.

synkd: Hey @ekohl ,

Thanks for the feedback. I added the trailing comma. Regarding the second suggestion, if this parameter should undefined by default, where do we set it so that it's present where Katello needs it?

ekohl: https://github.com/theforeman/puppet-foreman_proxy_content/blob/cf4cc2f6b402c447173e108d807ef28c0e2d75f9/manifests/init.pp#L295-L309
The first part is where this class is called and the `foreman::config::apache::fragment` actually sets the header.

wbclark: If the parameter is optional and undefined by default here, then I think we can assign some value to it when we declare the class here: https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L295-L298

We would also then make the line of the template `settings.py.erb` optional by placing it inside an erb `if` block.

@ekohl is this the type of approach you have in mind?

ekohl: @wbclark exactly

templates/settings.py.erb
```diff
@@ -17,4 +17,6 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
-REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'
+<% if $pulpcore::django_remote_user %>
```
ekohl: I think this should be 
```suggestion
<% if scope['pulpcore::django_remote_user'] -%>
```

templates/settings.py.erb
```diff
@@ -17,4 +17,6 @@ REDIS_HOST = "localhost"
...
+REMOTE_USER_ENVIRON_NAME = '<%= scope['pulpcore::django_remote_user'] %>'
+<% end %>
```
ekohl: If you strip the whitespace, the file will be more readable.
```suggestion
<% end -%>
```

templates/settings.py.erb
```diff
@@ -17,4 +17,6 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
-REMOTE_USER_ENVIRON_NAME = 'HTTP_REMOTE_USER'
+<% if defined?(scope['pulpcore::django_remote_user']) -%>
```
wbclark: Latest issue:

```
Failures:

  1) pulpcore on centos-7-x86_64 is expected to contain Concat::Fragment[base] with content !~ /REMOTE_USER_ENVIRON_NAME/

     Failure/Error: it { is_expected.to contain_concat__fragment('base').without_content(/REMOTE_USER_ENVIRON_NAME/) }

     

       expected that the catalogue would contain Concat::Fragment[base] with content not set to /REMOTE_USER_ENVIRON_NAME/ but it is set to "CONTENT_HOST = \"foo.example.com\"\nCONTENT_ORIGIN = \"http://foo.example.com\"\nPULP2_MONGODB = {\"name\": \"pulp_database\", \"seeds\": \"localhost:27017\"}\nSECRET_KEY = \"qoXnTdTofHouJcuARF4XZruhoEtqzhwk\"\nDATABASES = {\n    'default': {\n        'ENGINE': 'django.db.backends.postgresql',\n        'NAME': 'pulpcore',\n        'USER': 'pulp',\n        'PASSWORD': 'votFojScBLa3jbR33byoGMbMGQPabg9M',\n        'HOST': '127.0.0.1',\n        'PORT': '5432',\n    }\n}\nMEDIA_ROOT = \"/var/lib/pulp/docroot\"\nREDIS_HOST = \"localhost\"\nREDIS_PORT = \"6379\"\nREDIS_DB = 8\nWORKING_DIRECTORY = \"/var/lib/pulp/tmp\"\nREMOTE_USER_ENVIRON_NAME = ''\n"
```

This suggests that the template rendered to include `REMOTE_USER_ENVIRONMENT_NAME = ''` when the whole line should have been omitted. This implies that something is wrong with our approach of using `if defined?(scope['pulpcore::django_remote_user'])`

wbclark: I think in the case where `$pulpcore::django_remote_user = undef` then `scope['pulpcore::django_remote_user']` is getting evaluated as `''` so we get `if defined?('')` which evalues as `true` and renders the block.

I think we need something like `if defined?(@pulpcore::django_remote_user)`

wbclark: Per https://docs.huihoo.com/puppet/guides/templating.html we have a few options.

`if scope['pulpcore::django_remote_user'] != ''` seems straightforward, as I can't see an use case where we ever want to write the line in the config without actually defining the setting.

Another option I think is adding `$django_remote_user = $pulpcore::django_remote_user` inside `manifests/config.pp` so that the template could do `if defined?(@django_remote_user)` without going out of scope.

Will get with https://github.com/synkd in the morning and try this.

wbclark: Second approach wouldn't work: https://puppet.com/docs/puppet/5.5/lang_template_erb.html

Puppet's `undef` translates to `nil:nilClass` in Ruby. This explains an undefined method defined? for class nil:nilClass that we saw previously.

So I think `if scope['pulpcore::django_remote_user'] != ''` is the correct approach.

ekohl: The variable is always defined so `defined?` always evaluates to true. I wonder if we should switch the template to [epp](https://puppet.com/docs/puppet/latest/lang_template_epp.html) and use Puppet logic. Then we can write a check like `<% if $remote_user_environ_name =~ String[1] -%>`

ekohl: Btw, did you check whether just `if scope['pulpcore::django_remote_user']` works?

manifests/init.pp
```diff
@@ -66,7 +69,8 @@
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
-  Stdlib::Fqdn $servername = $facts['fqdn']
+  Stdlib::Fqdn $servername = $facts['fqdn'],
+  Optional[String] $django_remote_user = undef,
```
ekohl: Some bikeshedding: does it make sense to name this closer to the actual setting? When you're familiar with Pulp then you may know `REMOTE_USER_ENVIRON_NAME` but `django_remote_user` doesn't explain itself.

wbclark: We talked it over and we agree with you.

templates/settings.py.erb
wbclark: I thought we had tried that yesterday but apparently not -- this time it works.

PR #63, opened 2020-02-06
Refs #28720 - move migration params into migration plugin

wbclark:


ehelms:
A test for this would be nice to have.

wbclark:
> A test for this would be nice to have.

Thanks for the comment, @ehelms 

The previous commit added a spec test for the migration plugin, https://github.com/theforeman/puppet-pulpcore/blob/master/spec/classes/plugin_migration_spec.rb , which tests the default values of the params.

Is it better to test using non-default instead?

templates/migration-settings.py.erb
```diff
@@ -1,12 +1,12 @@
...
-    "seeds": "<%= scope['pulpcore::mongo_db_seeds'] %>",
-    "username": "<%= scope['pulpcore::mongo_db_username'] %>",
-    "password": "<%= scope['pulpcore::mongo_db_password'] %>",
-    "replica_set": "<%= scope['pulpcore::mongo_db_replica_set'] %>",
-    "ssl": "<%= scope['pulpcore::mongo_db_ssl'] %>",
-    "ssl_keyfile": "<%= scope['pulpcore::mongo_db_ssl_keyfile'] %>",
-    "ssl_certfile": "<%= scope['pulpcore::mongo_db_ssl_certfile'] %>",
-    "verify_ssl": "<%= scope['pulpcore::mongo_db_verify_ssl'] %>",
-    "ca_path": "<%= scope['pulpcore::mongo_db_ca_path'] %>",
+    "name": "<%= scope['pulpcore::plugin::migration::mongo_db_name'] %>",
```
ekohl: For future reference, these variables are now in scope so you can use `@mongodb_db_name`

manifests/plugin/migration.pp
```diff
@@ -1,5 +1,47 @@
...
+# @param mongo_db_ca_path
+#   CA bundle path to use for MongoDB SSL connection
+#
+class pulpcore::plugin::migration (
+  String $mongo_db_name = 'pulp_database',
+  String $mongo_db_seeds = 'localhost:27017',
+  Optional[String] $mongo_db_username = undef,
+  Optional[String] $mongo_db_password = undef,
+  Optional[String] $mongo_replica_set = undef,
+  String $mongo_db_ssl = 'False',
```
ekohl: We should be using real booleans and only translate them in the templates. For example here `Boolean $mongo_db_ssl = false`. Then in the template `<%= @mongo_db_ssl ? 'True' : 'False' %>`. The same goes for the ssl verify parameter.

manifests/plugin/migration.pp
```diff
@@ -1,5 +1,47 @@
...
+#
+# @param mongo_db_ca_path
+#   CA bundle path to use for MongoDB SSL connection
+#
+class pulpcore::plugin::migration (
+  String $mongo_db_name = 'pulp_database',
+  String $mongo_db_seeds = 'localhost:27017',
+  Optional[String] $mongo_db_username = undef,
+  Optional[String] $mongo_db_password = undef,
+  Optional[String] $mongo_replica_set = undef,
```
ekohl: This parameter doesn't match the documentation which is for `mongo_db_replica_set`.

PR #64, opened 2020-02-07
Fixes #28901 - support external postgresql

wbclark:
Fixes GH-4

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -1,17 +1,19 @@
 # Set up the PostgreSQL and Redis databases
 # @api private
 class pulpcore::database {
-  include postgresql::client
-  include postgresql::server
-  postgresql::server::db { $pulpcore::postgresql_db_name:
-    user     => $pulpcore::user,
-    password => postgresql_password($pulpcore::user, $pulpcore::postgresql_db_password),
+  if $pulpcore::postgresql_manage_db {
+    include postgresql::client
+    include postgresql::server
+    postgresql::server::db { $pulpcore::postgresql_db_name:
+      user     => $pulpcore::user,
+      password => postgresql_password($pulpcore::user, $pulpcore::postgresql_db_password),
+      before   => Pulpcore::Admin['migrate --noinput'],
+    }
   }
 
   pulpcore::admin { 'migrate --noinput':
     unless      => 'python3-django-admin migrate --plan | grep "No planned migration operations"',
     refreshonly => false,
-    require     => Postgresql::Server::Db[$pulpcore::postgresql_db_name],
   }
 
   include redis
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -39,6 +39,15 @@
 # @param postgresql_db_password
 #   Password of Pulp database
 #
+# @param postgresql_db_host
+#   Host to connect to Pulp database
+#
+# @param postgresql_db_port
+#   Port to connect to Pulp database
+#
+# @param postgresql_manage_db
+#   Whether or not to manage the PostgreSQL installation. If false, a database at the specified host and port is expected to exist and the user should have sufficient permissions.
+#
 # @param django_secret_key
 #   SECRET_KEY for Django
 #
@@ -67,6 +76,9 @@ class pulpcore (
   Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
+  Stdlib::Host $postgresql_db_host = 'localhost',
+  Stdlib::Port $postgresql_db_port = 5432,
+  Boolean $postgresql_manage_db = true,
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['fqdn'],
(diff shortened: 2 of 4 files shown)


manifests/database.pp
```diff
@@ -1,17 +1,18 @@
...
+    postgresql::server::db { $pulpcore::postgresql_db_name:
+      user     => $pulpcore::user,
+      password => postgresql_password($pulpcore::user, $pulpcore::postgresql_db_password),
+    }
   }
 
   pulpcore::admin { 'migrate --noinput':
     unless      => 'python3-django-admin migrate --plan | grep "No planned migration operations"',
     refreshonly => false,
-    require     => Postgresql::Server::Db[$pulpcore::postgresql_db_name],
```
wbclark: @ehelms Is it ok to remove the `require`? I believe we should be able to run migrations still when configured to use an external DB.

manifests/database.pp
```diff
@@ -1,17 +1,19 @@
...
+    postgresql::server::db { $pulpcore::postgresql_db_name:
+      user     => $pulpcore::user,
+      password => postgresql_password($pulpcore::user, $pulpcore::postgresql_db_password),
+    }
 
-  pulpcore::admin { 'migrate --noinput':
-    unless      => 'python3-django-admin migrate --plan | grep "No planned migration operations"',
-    refreshonly => false,
-    require     => Postgresql::Server::Db[$pulpcore::postgresql_db_name],
+    pulpcore::admin { 'migrate --noinput':
```
ehelms: We'll still want to this to happen even if the database is external since it's handling migration of the data.

wbclark: @ehelms thanks. I thought so, but I ran into problems when I tried it.

I realized now, the issue was that `require     => Postgresql::Server::Db[$pulpcore::postgresql_db_name],` doesn't work when we aren't managing the DB. Flipping this to a `before` on the other resource instead seems to fix it.

spec/classes/pulpcore_spec.rb
```diff
@@ -19,6 +19,22 @@
             .that_notifies(['Class[Pulpcore::Database]', 'Class[Pulpcore::Service]'])
         end
       end
+
+      context 'with external postgresql' do
+        let :params do
+          {
+            postgresql_db_host: '222.222.222.222',
```
ekohl: https://community.theforeman.org/t/using-ips-and-domains-in-tests-documentation/16600

manifests/init.pp
```diff
@@ -67,6 +76,9 @@
   Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
+  Stdlib::Host $postgresql_db_host = '127.0.0.1',
```
ekohl: Probably better to use `localhost` since we're slowly leaving the IPv4 world behind.
```suggestion
  Stdlib::Host $postgresql_db_host = 'localhost',
```

spec/classes/pulpcore_spec.rb
wbclark: Bookmarked, thanks

spec/classes/pulpcore_spec.rb
```diff
@@ -19,6 +19,22 @@
...
+        let :params do
+          {
+            postgresql_db_host: '192.0.2.222',
+            postgresql_manage_db: false,
+          }
+        end
+
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.not_to contain_class('postgresql::server::db')
```
ekohl: The DB is a define so this will always be true. I'd recommend adding a check i the default case as `is_expected.to` and here as `not_to`. Probably like this
```suggestion
          is_expected.not_to contain_class('postgresql::server')
          is_expected.not_to contain_postgresql__server__db('pulpcore')
```

wbclark: @ekohl super helpful feedback. Thanks as always.

spec/classes/pulpcore_spec.rb
```diff
@@ -14,11 +14,30 @@
 
         it do
           is_expected.to compile.with_all_deps
+          is_expected.to contain_class('postgresql::server')
```
ekohl: You can also move this to the base context (line 10) to keep this test just about adding a plugin.

manifests/init.pp
```diff
@@ -39,6 +39,15 @@
...
 #   Password of Pulp database
 #
+# @param postgresl_db_host
+#   Host to connect to Pulp database
+#
+# @param postgresl_db_port
+#   Port to connect to Pulp database
+#
+# @param postgresql_manage_db
+#   Disable for external Pulp database. Host and Port must be specified.
```
ekohl: Host and port both have a default so it's redundant.
```suggestion
#   Whether or not to manage the PostgreSQL installation. If false, a database at the specified host and port is expected to exist and the user should have sufficient permissions.
```

manifests/init.pp
```diff
@@ -39,6 +39,15 @@
 # @param postgresql_db_password
 #   Password of Pulp database
 #
+# @param postgresl_db_host
```
ekohl: The lint plugin should have caught this
```suggestion
# @param postgresql_db_host
```

manifests/init.pp
```diff
@@ -39,6 +39,15 @@
...
+#   Host to connect to Pulp database
+#
+# @param postgresl_db_port
```
ekohl:
```suggestion
# @param postgresql_db_port
```

PR #65, opened 2020-02-10
Verify everything to have documented parameters

ekohl:
Modulesync will overwrite this, but I'll write a patch for that when we do another round.


ekohl:
Will fail on https://github.com/theforeman/puppet-pulpcore/pull/63/files#r377216556

PR #66, opened 2020-02-10
Refs #28720 - Fixup migration variables and template

wbclark:
Based on feedback here: https://github.com/theforeman/puppet-pulpcore/pull/63/files

### DIFF
--- a/spec/classes/plugin_migration_spec.rb
+++ b/spec/classes/plugin_migration_spec.rb
@@ -4,13 +4,8 @@ migration_config = <<-EOL
 PULP2_MONGODB = {
     "name": "pulp_database",
     "seeds": "localhost:27017",
-    "username": "",
-    "password": "",
-    "replica_set": "",
-    "ssl": "False",
-    "ssl_keyfile": "",
-    "ssl_certfile": "",
-    "verify_ssl": "True",
+    "ssl": False,
+    "verify_ssl": True,
     "ca_path": "/etc/pki/tls/certs/ca-bundle.crt",
 }
 EOL
--- a/templates/migration-settings.py.erb
+++ b/templates/migration-settings.py.erb
@@ -1,12 +1,22 @@
 PULP2_MONGODB = {
-    "name": "<%= scope['pulpcore::plugin::migration::mongo_db_name'] %>",
-    "seeds": "<%= scope['pulpcore::plugin::migration::mongo_db_seeds'] %>",
-    "username": "<%= scope['pulpcore::plugin::migration::mongo_db_username'] %>",
-    "password": "<%= scope['pulpcore::plugin::migration::mongo_db_password'] %>",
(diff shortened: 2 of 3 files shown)


templates/migration-settings.py.erb
```diff
@@ -1,12 +1,22 @@
...
-    "password": "<%= scope['pulpcore::plugin::migration::mongo_db_password'] %>",
-    "replica_set": "<%= scope['pulpcore::plugin::migration::mongo_db_replica_set'] %>",
-    "ssl": "<%= scope['pulpcore::plugin::migration::mongo_db_ssl'] %>",
-    "ssl_keyfile": "<%= scope['pulpcore::plugin::migration::mongo_db_ssl_keyfile'] %>",
-    "ssl_certfile": "<%= scope['pulpcore::plugin::migration::mongo_db_ssl_certfile'] %>",
-    "verify_ssl": "<%= scope['pulpcore::plugin::migration::mongo_db_verify_ssl'] %>",
-    "ca_path": "<%= scope['pulpcore::plugin::migration::mongo_db_ca_path'] %>",
+    "name": "<%= @mongo_db_name %>",
+    "seeds": "<%= @mongo_db_seeds %>",
+<% if not @mongo_db_username.nil? -%>
```
ekohl: Perhaps easier to read
```suggestion
<% unless @mongo_db_username.nil? -%>
```

templates/migration-settings.py.erb
```diff
@@ -1,12 +1,22 @@
...
+    "username": "<%= @mongo_db_username %>",
+<% end -%>
+<% if not @mongo_db_password.nil? -%>
+    "password": "<%= @mongo_db_password %>",
+<% end -%>
+<% if not @mongo_db_replica_set.nil? -%>
+    "replica_set": "<%= @mongo_db_replica_set %>",
+<% end -%>
+    "ssl": "<%= @mongo_db_ssl ? 'True' : 'False' %>",
```
ekohl: I think this should actually be a Python boolean
```suggestion
    "ssl": <%= @mongo_db_ssl ? 'True' : 'False' %>,
```

templates/migration-settings.py.erb
wbclark: done.

templates/migration-settings.py.erb
wbclark: and done.

spec/classes/plugin_migration_spec.rb
```diff
@@ -4,12 +4,7 @@
 PULP2_MONGODB = {
     "name": "pulp_database",
     "seeds": "localhost:27017",
-    "username": "",
-    "password": "",
-    "replica_set": "",
     "ssl": "False",
```
ekohl: I would expect this not to be quoted

wbclark: good point

PR #67, opened 2020-02-11
Refs #28901 - add setting postgresql_db_user

wbclark:


wbclark:
@ekohl this fails in `spec/acceptance/basic_spec.rb` when I change the `postgresql_database_user` to a value other than `pulp`. For example with the user 'pulpcore' we get a password authentication error for that user during `django-admin migrate --noinput`. Any idea why that happens?

ekohl:
On the same host it may be using ident auth rather than password auth which would fail. Not entirely sure.

evgeni:
> @ekohl this fails in `spec/acceptance/basic_spec.rb` when I change the `postgresql_database_user` to a value other than 'pulp'. For example with the user 'pulpcore' we get a password authentication error for that user during `django-admin migrate --noinput`. Any idea why that happens?

You did not pass the username to the password hash function, so it was always computing the hash like the user is `pulp` ;)

https://github.com/theforeman/puppet-pulpcore/pull/373

PR #69, opened 2020-02-13
Fixes #28996 - Set PULP_STATIC_ROOT

wbclark:


manifests/admin.pp
```diff
@@ -31,6 +31,7 @@
     environment => [
       'DJANGO_SETTINGS_MODULE=pulpcore.app.settings',
       "PULP_SETTINGS=${pulp_settings}",
+      'PULP_STATIC_ROOT=/var/lib/pulp/assets/',
```
ekohl: Could you make this a parameter on the `pulpcore` class so we're certain they are in sync?

wbclark:
I've introduced a new parameter `pulp_static_root` and added tests for custom static directories.

spec/defines/admin_spec.rb
```diff
@@ -5,6 +5,11 @@
     context "on #{os}" do
       let(:facts) { os_facts }
       let(:title) { 'help' }
+      let(:pre_condition) do
```
ekohl: The alternative to this was to add the parameter to the `let(:params)` block. That was the intended way of this file and it would be good to keep that. There's a block `context 'with inheritance' do` that is testing it with inheritance.

PR #70, opened 2020-02-17
fixup admin spec

wbclark:


spec/defines/admin_spec.rb
```diff
@@ -5,14 +5,14 @@
...
-        PUPPET
-      end
 
-      context 'with a fixed pulp_settings' do
-        let(:params) { { pulp_settings: '/etc/pulpcore/settings.py' } }
+      context 'with a fixed pulp_settings and static_root' do
+        let(:params) do
+          super().merge(
+            pulp_settings: '/etc/pulpcore/settings.py',
+            static_root: '/var/lib/pulp/assets'
```
ekohl: Generally I'm fan of trailing comma's. The benefit is that if you add another line, you don't need to modify this line. You can also sort them. Same goes for the line you added near the end.

spec/defines/admin_spec.rb
```diff
@@ -5,14 +5,14 @@
...
-        <<-PUPPET
-          include pulpcore
-        PUPPET
-      end
 
-      context 'with a fixed pulp_settings' do
-        let(:params) { { pulp_settings: '/etc/pulpcore/settings.py' } }
+      context 'with a fixed pulp_settings and static_root' do
+        let(:params) do
+          super().merge(
```
ekohl: There is no super so just define it as a regular hash

wbclark: :+1:, thanks

PR #71, opened 2020-02-19
Refs #28901 - Support SSL connection for external postgresql database

wbclark:


manifests/init.pp
```diff
@@ -54,6 +54,21 @@
 # @param postgresql_manage_db
 #   Whether or not to manage the PostgreSQL installation. If false, a database at the specified host and port is expected to exist and the user should have sufficient permissions.
 #
+# @param postgresql_db_ssl
+#   Whether to configure SSL connection for postgres database. The configuration is only applied if the postgres database is unmanaged.
```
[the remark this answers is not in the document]
wbclark: @ekohl done.

wbclark:
Thanks for suggesting the use of the helper function, @ekohl . That works very nicely.

templates/settings.py.erb
```diff
@@ -9,7 +9,15 @@ DATABASES = {
...
         'HOST': '<%= scope['pulpcore::postgresql_db_host'] %>',
         'PORT': '<%= scope['pulpcore::postgresql_db_port'] %>',
-    }
+<% if scope['pulpcore::postgresql_db_ssl'] && !scope['pulpcore::postgresql_db_manage'] -%>
+	'OPTIONS': {
+            'sslmode': '<%= scope['pulpcore::postgresql_db_ssl_require'] ? 'require' : 'disable' %>',
+            'sslcert': '<%= scope['pulpcore::postgresql_db_ssl_cert'] %>',
+            'sslkey': '<%= scope['pulpcore::postgresql_db_ssl_key'] %>',
+            'sslrootcert': '<%= scope['pulpcore::postgresql_db_ssl_root_ca'] %>',
+        },
```
ehelms: Nitpick, pull this back one set of tabs to have it align when its actually rendered

ehelms: Err sorry, I think actually push it forward one set the closer I look.

wbclark: I think it's correct as is (except for the fact that the spaces in front of `OPTIONS` keep converting to a tab somehow, which I have corrected again). To be sure, I ran a puppet apply with the db unmanaged and ssl enabled on a test system and found the correct level of indentation in `/etc/pulp/settings.py`

wbclark:
> In spec/classes/pulpcore_spec.rb around line 12 you could add a test that `OPTIONS` is not present, or the sslmode. Alternatively just verify the exact contents there. Just to make sure the ssl options are not present by default.

Thanks @ekohl . I've added a test to validate these settings are not included by default.

PR #72, opened 2020-02-24
Fixes #29069 - Apply selinux labels to pulpcore ports

wbclark:


manifests/install.pp
```diff
@@ -18,4 +18,18 @@
     ensure => present,
   }
 
+  selinux::port { 'pulpcore-api-port':
```
ekohl: I'd consider this config because it needs the install for the package to be installed. We may also need to ensure `pulpcore-selinux` is installed (in `install.pp`)

wbclark: @ekohl I've moved those declarations into `config.pp`

About the package, I found this: https://github.com/theforeman/foreman-packaging/pull/4648/files#r382513097

From that discussion it seems that we really can't rely on `python3-pulpcore` to install `pulpcore-selinux` so I've declared the package as well.

manifests/install.pp
```diff
@@ -1,12 +1,16 @@
-# @summary Install pulpcore packages
-# Currently this is done with pip3. This should have the option to also install via RPMs.
+# @summary Install pulpcore packages, configure user and group, label ports
```
ekohl:
```suggestion
# @summary Install pulpcore packages, configure user and group
```

manifests/install.pp
```diff
@@ -1,12 +1,16 @@
...
+#
 # @api private
 class pulpcore::install {
 
   package { 'python3-pulpcore':
     ensure => present,
   }
 
+  package { 'pulpcore-selinux':
```
ekohl: Our CI env in Travis doesn't have SELinux enabled because it's a container. You may run into this and in that case this (and the ports) probably need to be wrapped in `if $facts['os']['selinux']['enabled'] {}`

ekohl: To be clear: I prefer to use the types unconditionally. https://github.com/voxpupuli/puppet-selinux/commit/ff459f3747e496670a10485cfd2c22d3e96e97d9 is in puppet-selinux 3.1.0 so it should work without. I'd also expect the RPM to be installable.

wbclark: Great. I'll unrwap the ports and but leave the pulpcore-selinux package wrapped in that case.

PR #73, opened 2020-02-28
Refs #28901 - Fix settings.py.erb

rafaelguerra01:


templates/settings.py.erb
```diff
@@ -9,7 +9,7 @@ DATABASES = {
         'PASSWORD': '<%= scope['pulpcore::postgresql_db_password'] %>',
         'HOST': '<%= scope['pulpcore::postgresql_db_host'] %>',
         'PORT': '<%= scope['pulpcore::postgresql_db_port'] %>',
-<% if scope['pulpcore::postgresql_db_ssl'] && !scope['pulpcore::postgresql_db_manage'] -%>
+<% if scope['pulpcore::postgresql_db_ssl'] && !scope['pulpcore::postgresql_manage_db'] -%>
```
wbclark: Looks great, @rafaelguerra01 ! Thanks for fixing my bug

ekohl: This indicates a lack of test coverage. It would be good to add a test for this.

PR #75, opened 2020-03-04
Prepare acceptance tests for EL8

ekohl:
Also updates the SCL from pg 10 to 12.


ehelms:
Can this be tied back to the Redmine issue that @wbclark has tracking EL8 support across modules?

ekohl:
Updated

PR #76, opened 2020-03-06
enable httpd_can_network_connect selinux boolean

wbclark:


ekohl:
Looks like the `selinux_core` fixture needs to be added, similar to `augeas_core` in `.fixtures.yml`.

PR #78, opened 2020-03-17
restorecon after creating cache dir

wbclark:
Previously, the catalogue is compiled before package `pulpcore-selinux` is installed, so the cache dir gets created with the wrong context initially. This breaks idempotence, as when the module is run a second time, it does so with the policy which is queried after `pulpcore-selinux` RPM is installed.

This change fixrd that issue by ensuring a `restorecon` is run after the directory gets created, so that it correctly gets the `pulpcore_var_lib_t` label defined in the policy by the `pulpcore-selinux` RPM even if that package wasn't installed at the time the catalogue is compiled.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -25,6 +25,10 @@ class pulpcore::config {
     owner  => $pulpcore::user,
     group  => $pulpcore::group,
     mode   => '0775',
+  } ~> exec { "restorecon -RvF ${pulpcore::cache_dir}":
+    path        => '/usr/bin:/usr/sbin:/bin:/sbin',
+    command     => "restorecon -RvF ${pulpcore::cache_dir}",
+    refreshonly => true,
   }
 
   pulpcore::admin { 'collectstatic --noinput':
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -16,6 +16,8 @@ describe 'pulpcore' do
           is_expected.to contain_postgresql__server__db('pulpcore')
           is_expected.to contain_apache__vhost('pulp')
           is_expected.to contain_selinux__boolean('httpd_can_network_connect')
+          is_expected.to contain_file('/var/lib/pulp/tmp')
+          is_expected.to contain_exec("restorecon -RvF /var/lib/pulp/tmp")
         end
       end
 


manifests/config.pp
```diff
@@ -25,6 +25,10 @@
     owner  => $pulpcore::user,
     group  => $pulpcore::group,
     mode   => '0775',
+  } ~> exec { "restorecon -RvF ${pulpcore::cache_dir}":
+    path        => '/usr/bin:/usr/sbin:/bin:/sbin',
+    command     => "restorecon -RvF ${pulpcore::cache_dir}",
+    refreshonly => true,
```
ekohl: Is this really needed? Can't we just set the type on the directory statically? It's only Puppet created files that suffer from it. Didn't check the exact type, but something like:
```suggestion
    seltype => 'pulp_var_t',
```

wbclark: I preferred my approach because this way the module doesn't make any assumptions about what policy is applied by the `pulpcore-selinux` RPM (i.e. it doesn't have to know that the proper label for this dir is `pulpcore_var_lib_t`) it simply ensures that the puppet created dirs get the correct label according to whatever is in the policy.

ehelms: Any idea why its only the cache_dir that suffers from this? And not any of the other directories.

ekohl: I think we don't use the directories that are in the policy.

I'm not a fan of `exec`. It's almost always a bad thing because idempotence is very hard. In this case, it can be very slow if you have a lot of cache files because this is recursive. That feels like a ticking time bomb.

Note that an alternative fix for this is to create the directory in the RPM packaging or install the selinux package after creating these directories. The selinux package has a `fixfiles` in `%post` so that would work. For simplicity I'd prefer the former, but the pulpcore package is particularly minimal and doesn't manage any files for actual installation so I'd be fine with the latter as well.

wbclark: Good question @ehelms . Here is what I found after applying the test manifest once:

```
[vagrant@centos8-64-1 ~]$ ll -Z /etc/pulp
total 4
-rw-r-----. 1 root pulp system_u:object_r:etc_t:s0 504 Mar 18 02:34 settings.py
[vagrant@centos8-64-1 ~]$ ll -Z /var/lib/pulp
total 0
drwxr-xr-x. 5 root root unconfined_u:object_r:var_lib_t:s0      57 Mar 18 02:34 assets
drwxrwxr-x. 2 pulp pulp system_u:object_r:var_lib_t:s0           6 Mar 18 02:34 docroot
drwxrwxr-x. 5 pulp pulp system_u:object_r:pulpcore_var_lib_t:s0 82 Mar 18 02:35 tmp
[vagrant@centos8-64-1 ~]$
```

checking the content of the files installed by `pulpcore-selinux` RPM:

```
[root@centos8-64-1 ~]# for file in $(rpm -ql pulpcore-selinux); do grep -v pulpcore $file | grep 'var_lib_t'; done
[root@centos8-64-1 ~]# for file in $(rpm -ql pulpcore-selinux); do grep 'var_lib_t' $file; done
Binary file /usr/share/selinux/mls/pulpcore.pp matches
Binary file /usr/share/selinux/strict/pulpcore.pp matches
Binary file /usr/share/selinux/targeted/pulpcore.pp matches
[root@centos8-64-1 ~]#
```

So it seems to me that the cache_dir is the only dir we are creating where the policy defined in `pulpcore-selinux` differs from the OS default.

ekohl: https://github.com/pulp/pulpcore-selinux/blob/master/pulpcore.fc are the file contexts. It's short enough to paste here:
```
/usr/lib/pulp/bin/gunicorn		--	gen_context(system_u:object_r:pulpcore_exec_t,s0)
/usr/lib/pulp/bin/rq		--	gen_context(system_u:object_r:pulpcore_exec_t,s0)

/usr/local/lib/pulp/bin/gunicorn		--	gen_context(system_u:object_r:pulpcore_exec_t,s0)
/usr/local/lib/pulp/bin/rq		--	gen_context(system_u:object_r:pulpcore_exec_t,s0)

/var/lib/pulp/artifact(/.*)? 	gen_context(system_u:object_r:pulpcore_var_lib_t,s0)
/var/lib/pulp/assets(/.*)? 	gen_context(system_u:object_r:pulpcore_var_lib_t,s0)
/var/lib/pulp/tmp(/.*)? 	gen_context(system_u:object_r:pulpcore_var_lib_t,s0)
/var/lib/pulp/upload(/.*)? 	gen_context(system_u:object_r:pulpcore_var_lib_t,s0)

/var/run/pulpcore*(/.*)? 	gen_context(system_u:object_r:pulpcore_var_run_t,s0)
/var/run/pulpcore-api(/.*)? 	gen_context(system_u:object_r:pulpcore_var_run_t,s0)
/var/run/pulpcore-content(/.*)? 	gen_context(system_u:object_r:pulpcore_var_run_t,s0)
/var/run/pulpcore-resource-manager(/.*)? 	gen_context(system_u:object_r:pulpcore_var_run_t,s0)
/var/run/pulpcore-resource-worker[0-9]*(/.*)? 	gen_context(system_u:object_r:pulpcore_var_run_t,s0)
```

Looking at this, the `/usr/local` paths are weird because we never install those. I'm also wondering about gunicorn and rq because I would expect those bins to be in `/usr/bin` and not `/usr/lib/pulp/bin`.

Also makes me wonder if our paths are correct since we don't use `artifact`, `assets` and `upload`.

Lastly, `/var/run` is usually a symlink to `/run` and I'd verify if that really gets applied.

In summary: this raises more questions than it answers. I had wanted to cherry pick SELinux into 2.0.0-rc3 but now I'm starting to doubt that.

ehelms: You are seeing some aspects of the policy designed to fit the Pulp PyPi installs into a virtualenv where the paths are going to differ for some parts (like rq) than typical system paths. However, everything under `/var/lib/pulp` should be the same.

I'd be good with creating the directories before installing the selinux package.

wbclark:
Closing in favor of #88

PR #79, opened 2020-03-18
Fixes #29371 - Update settings.py for pulpcore3.2

sjha4:
We are working on getting pulpcore3.2 built and released. This is required to support the changes required to pulp/settings 

@wbclark : This is the the only place I identified a need for changes to support pulpcore 3.2 with this module. Is there anything else that you can think of that needs to be done?

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -81,6 +81,9 @@
 # @param remote_user_environ_name
 #   Django remote user environment variable
 #
+# @param allowed_import_path
+#   Allowed paths that pulp can sync from using file:// protocol
+#
 # @example
 #   include pulpcore
 class pulpcore (
@@ -110,6 +113,7 @@ class pulpcore (
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['fqdn'],
+  Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Optional[String] $remote_user_environ_name = undef,
 ) {
 
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -120,6 +120,30 @@ describe 'pulpcore' do
         end
       end
 
+      context 'with allowed import paths' do
+        let :params do
+          {
+            allowed_import_path: ['/test/path', '/test/path2'],
+          }
+        end
+
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{ALLOWED_IMPORT_PATHS = \["/test/path", "/test/path2"\]})
+
+        end
+      end
+
+      context 'with empty allowed import paths' do
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{ALLOWED_IMPORT_PATHS = \["/var/lib/pulp/sync_imports"\]})
+
+        end
+      end
+
       context 'with custom static dirs' do
         let :params do
           {
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -27,3 +27,12 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
 <% if scope['pulpcore::remote_user_environ_name'] -%>
 REMOTE_USER_ENVIRON_NAME = '<%= scope['pulpcore::remote_user_environ_name'] %>'
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
+ALLOWED_IMPORT_PATHS = <%= scope['pulpcore::allowed_import_path'] %>
--- a/README.md
+++ b/README.md
@@ -1,2 +1,11 @@
 # puppet-pulpcore
 Puppet module for setting up Pulp 3 as part of Katello installation
+
+# Pulpcore 3.2
+
+We are adding a few new settings to make the installer compatible with pulpcore 3.2. These settings are not available in releases prior to 3.2 and should not be used in earlier versions.                                            
+
+**ALLOWED_IMPORT_PATHS** : This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests. For more information on this, see [https://docs.pulpproject.org/settings.html#allowed-import-paths](https://docs.pulpproject.org/settings.html#allowed-import-paths).
+
+**AUTHENTICATION_BACKENDS , REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES** : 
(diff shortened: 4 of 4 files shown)


templates/settings.py.erb
```diff
@@ -27,3 +27,11 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
...
 REMOTE_USER_ENVIRON_NAME = '<%= scope['pulpcore::remote_user_environ_name'] %>'
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+ALLOWED_IMPORT_PATHS = ['/var/www/test_repos']
```
ehelms: What is this line handling?

jlsherrill: Pulp 3 added a restriction on what paths you can sync from using the 'file://' 'protocol'.    So this is needed to allow our tests to run, however @sjha4 i think we probably want to set this to a more generic path for users and probably one that would allow python to optimize the copy (which later versions of python do).  I would recommend /var/lib/pulp/sync_imports/  or something like that.  And we'll need to update our tests to use that

sjha4: Yes..this is to handle our tests only for now..Let me try modifying tests with this /var/lib/pulp/sync_imports/ and check if I run into any issues.. :+1:

ehelms: If only for testing, a comment in this case would be useful as that is information I'm sure will be lost on us down the road.

sjha4: Added the comment and updated katello tests to use path '/var/lib/pulp/sync_imports/'

jlsherrill: @ehelms its not *only* for tests, pulp will ONLY sync from file:// repos in the directories we specify, this is a change from pulp2.  Our tests that do this need to use this directory as well.

jlsherrill:
ACK from me, but i think this needs to wait to be merged until https://github.com/Katello/katello/pull/8617 is ready

templates/settings.py.erb
```diff
@@ -27,3 +27,12 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
...
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests%>
+ALLOWED_IMPORT_PATHS = ['/var/lib/pulp/sync_imports/']
```
ekohl: It's a good practice to always use variables for paths. Even if you don't want to change it, this allows reusing this value if Katello needs it. It also allows setting it to an empty array in production for additional security.

templates/settings.py.erb
```diff
@@ -27,3 +27,12 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
...
 REMOTE_USER_ENVIRON_NAME = '<%= scope['pulpcore::remote_user_environ_name'] %>'
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests%>
```
ekohl: IMHO it's fine to place this as a Python comment rather than an ERB comment. If you keep it as an ERB comment, please strip the trailing whitespace (`-%>` instead of `%>`).

manifests/init.pp
```diff
@@ -110,6 +113,7 @@
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['fqdn'],
+  Stdlib::Absolutepath $allowed_import_path = '/var/lib/pulp/sync_imports',
   Optional[String] $remote_user_environ_name = undef,
```
sjha4: @ekohl: I was trying to look it up but how to do I define this to be an array of absolute paths?

sjha4: Looks like I can do this https://github.com/theforeman/puppet-pulpcore/blob/c27d67a6c9705b99c942729dd539698e734f6bd3/manifests/admin.pp#L28

ekohl: Exactly. I love it when people can answer their own questions :)

templates/settings.py.erb
```diff
@@ -27,3 +27,12 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
...
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
+ALLOWED_IMPORT_PATHS = "<%= scope['pulpcore::allowed_import_path'] %>"
```
ekohl: I think this needs to be a loop to construct an array.

sjha4:
@ekohl : Added some tests and looping over param to get the correct array.

spec/classes/pulpcore_spec.rb
```diff
@@ -120,6 +120,30 @@
...
+      context 'with allowed import paths' do
+        let :params do
+          {
+            allowed_import_path: ['/test/path', '/test/path2'],
+          }
+        end
+
+        it do
+          is_expected.to compile.with_all_deps
+	  is_expected.to contain_concat__fragment('base')
```
ekohl: Nit: looks like there's a tab here

templates/settings.py.erb
```diff
@@ -27,3 +27,17 @@ WORKING_DIRECTORY = "<%= scope['pulpcore::cache_dir'] %>"
...
 <% end -%>
+AUTHENTICATION_BACKENDS = ['pulpcore.app.authentication.PulpNoCreateRemoteUserBackend']
+
+REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
+    'rest_framework.authentication.SessionAuthentication',
+    'pulpcore.app.authentication.PulpRemoteUserAuthentication'
+)
+
+<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
+<% allowed_paths = [] -%>
```
ekohl: Why is this needed? If anything you can use `ALLOWED_IMPORT_PATHS = <%= Array.new(scope['pulpcore::allowed_import_path']) %>` but I think it should not be needed since it's already an array.

sjha4: That is correct..Updated..This works `scope['pulpcore::allowed_import_path']` .. :+1:

jlsherrill:
These two settings:

AUTHENTICATION_BACKENDS
REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES 

are needed for katello's cert-auth to work, and may cause previous versions of pulp to not function properly, although there were so many bugs around it within pulp, it may just ignore them alltogether.  I can give it a test real quick

jlsherrill:
Testing revealed that it does in fact break pulpcore 3.0 because 'PulpNoCreateRemoteUserBackend' is not defined

ekohl:
I don't mind breaking it, but I want to clearly know that we did so we don't accidentally cherry pick it. It'd also be good to have this in the README.

sjha4:
@jlsherrill , @ekohl : Updated Readme to add some information on the new settings detailing these are meant to be used in 3.2

PR #80, opened 2020-03-18
install rh-redis5-redis in el7 acceptance spec

wbclark:
This should resolve (rather, hide) the other idempotence issue in the el7 acceptance tests, by ensuring the redis packages are installed already at the time our catalogue is compiled.


evgeni:
The travis errors look unrelated?

ekohl:
I think I saw master fail in the same way, going to have a look.

ekohl:
https://github.com/voxpupuli/puppet-selinux/issues/318 is causing issues, will create a PR to pin to Facter < 4.

PR #85, opened 2020-04-02
Create admin user

sjha4:
We need to create the default admin user since pulpcore.app.authentication.PulpNoCreateRemoteUserBackend authentication backend does not create it automatically for us.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -16,6 +16,12 @@ class pulpcore::database {
     refreshonly => false,
   }
 
+  pulpcore::admin { 'reset-admin-password --random':
+    unless      => 'python3-django-admin dumpdata auth.User | grep "auth.user"',
+    refreshonly => false,
+    require     => Pulpcore::Admin['migrate --noinput'],
+  }
+
   include redis
 
 }
--- a/spec/acceptance/basic_spec.rb
+++ b/spec/acceptance/basic_spec.rb
@@ -54,4 +54,9 @@ describe 'basic installation' do
     its(:exit_status) { is_expected.to eq 0 }
   end
 
+ describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py python3-django-admin dumpdata auth.User") do
+   its(:stdout) { is_expected.to match(/auth\.user/) }
+   its(:exit_status) { is_expected.to eq 0 }
+ end
+
 end
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -16,6 +16,7 @@ describe 'pulpcore' do
           is_expected.to contain_postgresql__server__db('pulpcore')
           is_expected.to contain_apache__vhost('pulp')
           is_expected.to contain_selinux__boolean('httpd_can_network_connect')
+          is_expected.to contain_pulpcore__admin('reset-admin-password --random')
         end
       end
 


manifests/database.pp
```diff
@@ -16,6 +16,8 @@
     refreshonly => false,
   }
 
+  pulpcore::admin { 'reset-admin-password --password' }
```
ekohl, line 19: This doesn't look very idempotent. There's also no guarantee this actually runs after the migration now.

manifests/database.pp
```diff
@@ -16,6 +16,12 @@
     refreshonly => false,
   }
 
+  pulpcore::admin { 'reset-admin-password --random':
+    unless      => 'python3-django-admin dumpdata | grep "auth.user"',
+    refreshonly => false,
+    require     => Pulpcore::Admin['migrate --noinput'],
+  }
+
```
sjha4, line 24: @ekohl: Testing this now..Does this look ok?

sjha4:
@wbclark : Can you try this:
sudo -u pulp DJANGO_SETTINGS_MODULE=pulpcore.app.settings python3-django-admin dumpdata | grep "auth.user"

I am looking at the tests currently in the repo..Where would the tests for this logically go?

wbclark:
@sjha4 same output when I try sudo as pulp user.

I checked `settings.py`:

```shell
[vagrant@centos8-64-1 ~]$ sudo fgrep CONTENT_ORIGIN /etc/pulp/settings.py 
CONTENT_ORIGIN = "http://centos8-64-1"
```
From the documentation[1] added in the same commit that introduced `CONTENT_ORIGIN` setting into Pulpcore, it looks like we may need the port defined as well.

[1] https://github.com/pulp/pulpcore/commit/e027cd3fca1f89a037637b2b08492e2c5624a679#diff-572a386e0821364b3217fc84b059de24R212-R220

But I tried adding the port here on a separate branch and still faced the same issue. I also tried other `python3-django-admin` commands and saw the same stack trace, so I don't think this PR caused it but I'm not confident that `unless => 'python3-django-admin dumpdata | grep "auth.user"'` will work properly until we fix it.

spec/defines/admin_spec.rb
```diff
@@ -50,6 +50,29 @@
               .with_unless('/usr/bin/false')
           end
         end
+
+        context 'default parameters' do
```
wbclark, line 54: `spec/defines/admin_spec.rb` is used to test that the `pulpcore::admin` class compiles properly under various params.

In all the examples here, we're testing `pulpcore::admin { 'help' }` as you can see when we defined the `title` for the class here: https://github.com/theforeman/puppet-pulpcore/blob/master/spec/defines/admin_spec.rb#L7 , basically using `help` as a dummy admin command we might to run, and then testing that it behaves properly with various combinations of `unless`, `refreshonly`, etc.

So probably this isn't the right test to use for our purposes. I'll suggest alternatives in another comment.

wbclark: In `spec/acceptance/basic_spec.rb` I think we could test that the admin user got created after running puppet-apply, using the same logic to test as in your `unless` condition, adding after L55:

```
describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings python3-django-admin dumpdata | grep "auth.user"") do
```

We have to fix the issue `NameError: name 'CONTENT_ORIGIN' is not defined` that we saw previously before this will work.

ekohl: As @wbclark said, this probably belongs in `spec/classes/pulpcore_spec.rb`

sjha4:
> same output when I try sudo as pulp user.

sudo -u pulp DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py python3-django-admin dumpdata | grep "auth.user"

wbclark:
> > same output when I try sudo as pulp user.
> 
> sudo -u pulp DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py python3-django-admin dumpdata | grep "auth.user"

Great, that worked for me! Thanks @sjha4 

One more request, could you add a unit test in `spec/classes/pulpcore_spec.rb`, adding after L18:

```
is_expected.to contain_pulpcore__admin('reset-admin-password --random')
```

I think at that point we are good to ACK and merge this :)

manifests/database.pp
```diff
@@ -16,6 +16,12 @@
     refreshonly => false,
   }
 
+  pulpcore::admin { 'reset-admin-password --random':
+    unless      => 'python3-django-admin dumpdata | grep "auth.user"',
```
ekohl, line 20: This dumps the entire database which can be very expensive. You can dump data just for a single model (https://docs.djangoproject.com/en/3.0/ref/django-admin/#dumpdata) so it think this would work:
```suggestion
    unless      => 'python3-django-admin dumpdata auth.User | grep "auth.user"',
```

sjha4: This works. Updated.. :+1:

spec/acceptance/basic_spec.rb
```diff
@@ -54,4 +54,8 @@ class { 'redis::globals':
     its(:exit_status) { is_expected.to eq 0 }
   end
 
+ describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py python3-django-admin dumpdata auth.User | grep 'auth.user'") do
```
ekohl, line 57: You can avoid using grep here:

```suggestion
 describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py python3-django-admin dumpdata auth.User") do
   its(:stdout) { is_expected.to match(/auth\.user/) }
```

This is useful when it fails. Grep doesn't show non-matches and it'll just be a silent failure. Using the matcher it'll show a usable error.

wbclark: I agree this approach is better... thanks for pointing it out.

sjha4: Updated.. :+1:

wbclark: Thanks @sjha4 and @ekohl ! Will merge once tests are complete.

PR #87, opened 2020-04-03
Refs #29371 - use django syntax to set DRF authenticators

evgeni:
After longish reading of https://docs.pulpproject.org/installation/authentication.html, I think that the `REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES` is supposed to be used as an environment variable, while `settings.py` is a Django settings file, so it needs proper Python dicts :)

After changing this, I could create file repos again.

@sjha4 can you have a look please?


evgeni:
according to @jlsherrill I am stupid and this makes it not use any auth at all (which is weird and true as I tested it).

PR #77, opened 2020-03-10
Fixes #29190 - Support EL8

wbclark:
After marking el8 as a supported OS in metadata.json, I can run spec tests using el8 facts:

```
[wclark@perseids puppet-pulpcore]$ SPEC_FACTS_OS=centos-8-x86_64 bundle exec rspec spec/classes/plugin_migration_spec.rb 
...

Finished in 4.83 seconds (files took 5.29 seconds to load)
3 examples, 0 failures

[wclark@perseids puppet-pulpcore]$ SPEC_FACTS_OS=centos-8-x86_64 bundle exec rspec spec/classes/plugin_file_spec.rb 
...

Finished in 4.03 seconds (files took 5.54 seconds to load)
3 examples, 0 failures

[wclark@perseids puppet-pulpcore]$ SPEC_FACTS_OS=centos-8-x86_64 bundle exec rspec spec/classes/plugin_container_spec.rb 
...

Finished in 3.8 seconds (files took 5.1 seconds to load)
3 examples, 0 failures

[wclark@perseids puppet-pulpcore]$ SPEC_FACTS_OS=centos-8-x86_64 bundle exec rspec spec/classes/pulpcore_spec.rb 
.......

Finished in 19.54 seconds (files took 5.41 seconds to load)
7 examples, 0 failures

[wclark@perseids puppet-pulpcore]$ 
```

PR #90, opened 2020-04-11
Modulesync: use voxpupuli helper gems

ekohl:


spec/spec_helper_acceptance.rb
```diff
@@ -1,53 +1,20 @@
...
-        end
-
-        baseurl = if ENV['PULPCORE_REPO_RELEASE']
-                    "https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el#{major}/x86_64/"
-                  else
-                    "http://koji.katello.org/releases/yum/katello-nightly/pulpcore/el#{major}/x86_64/"
-                  end
+ENV['BEAKER_setfile'] ||= 'centos7-64{hostname=centos7-64.example.com}'
 
-        on host, puppet_resource('yumrepo', 'pulpcore', "baseurl=#{baseurl}", 'gpgcheck=0')
```
wbclark, line 45: L32-45 here shouldn't be replaceable by the gem. We need `rh-redis5-redis` and the pulpcore repository, and those needs are specific to this module.

wbclark: Also postgresql installation is failing without `centos-release-scl-rh`

ekohl: Yes, but this is the idea of modulesync: the file is managed in a central way. We have diverted from it and that causes issues now. We need to fix this in a central way or not manage the file.

wbclark: Perhaps the intention is to address this using: https://github.com/theforeman/foreman-installer-modulesync/pull/116#issuecomment-610095824 ?

ekohl: Addressed via `setup_acceptance_node.pp` now

spec/setup_acceptance_node.pp
```diff
@@ -0,0 +1,37 @@
...
+      ensure => installed,
+    }
+
+    package { 'rh-redis5-redis':
+      ensure  => installed,
+      require => Package['centos-release-scl-rh'],
+    }
+  }
+  '8': {
+    package { 'glibc-langpack-en':
```
wbclark, line 15: I like the approach of splitting off all module specific acceptance config into a new file.

I want to say, could we add a comment that installing `glibc-langpack-en` will no longer be necessary once we have: https://github.com/puppetlabs/beaker-hostgenerator/pull/183

But I could tackle that in https://github.com/theforeman/puppet-pulpcore/pull/77 after rebasing, since I assume I will need to make some further changes in this file to get EL8 working regardless.

ekohl: the language pack is only needed if we need PostgreSQL, so I opted to keep it in a local override.

PR #91, opened 2020-04-13
Update pulpcore::admin to use the command parameter if it is passed

gcoxmoz:
In pulpcore::admin, there is a `$command` parameter that defaults to `$title`.  However, `$command` is never used, and `${title}` is used inside an exec where the $command ought to be.

### DIFF
--- a/spec/defines/admin_spec.rb
+++ b/spec/defines/admin_spec.rb
@@ -40,7 +40,7 @@ describe 'pulpcore::admin' do
 
           it do
             is_expected.to compile.with_all_deps
-            is_expected.to contain_exec('python3-django-admin help')
+            is_expected.to contain_exec('python3-django-admin migrate --noinput')
               .with_environment([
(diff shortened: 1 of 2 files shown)


ekohl:
It looks like the tests need to be updated.

gcoxmoz:
erp.  unfortunately I've never figured out the test language in order to write them.

spec/defines/admin_spec.rb
```diff
@@ -40,7 +40,7 @@
 
           it do
             is_expected.to compile.with_all_deps
-            is_expected.to contain_exec('python3-django-admin help')
+            is_expected.to contain_exec('python3-django-admin migrate --noinput')
```
wbclark, line 43: Looks good. In fact, when I scroll up a few lines I can see that this is what it should have been all along.

wbclark:
Thanks for the contribution @gcoxmoz !

PR #93, opened 2020-05-06
set module_hotfix=1 for pulpcore repo on el8 acceptance spec

wbclark:


wbclark:
This works but I need to rebase it. I thought I pulled latest master before building this, apparently not.

PR #99, opened 2020-06-09
Fixes #30059 - Add certguard plugin

sjha4:


manifests/plugin/certguard.pp
```diff
@@ -0,0 +1,4 @@
+class pulpcore::plugin::certguard {
```
wbclark, line 1:
```suggestion
# @summary Pulp Certguard plugin
class pulpcore::plugin::certguard {
```

This is helpful to have because we generate docs automatically from these strings.

wbclark:
@sjha4 could you also please add a unit test like below at spec/classes/plugin_certguard_spec.rb:

```
require 'spec_helper'

describe 'pulpcore::plugin::certguard' do
  on_supported_os.each do |os, os_facts|
    context "on #{os}" do
      let(:facts) { os_facts }

      it { is_expected.to contain_pulpcore__plugin('certguard') }

      context 'with pulpcore' do
        let(:pre_condition) { 'include pulpcore' }

        it do
          is_expected.to contain_pulpcore__plugin('certguard')
            .that_subscribes_to('Class[Pulpcore::Install]')
            .that_notifies(['Class[Pulpcore::Database]', 'Class[Pulpcore::Service]'])
        end
      end
    end
  end
end
```

It's maybe a little bit overkill but this skeleton is nice to have to build upon when any future changes are made

PR #100, opened 2020-06-11
Refs #30057 - Configure Pulpcore Worker Count

wbclark:


manifests/init.pp
```diff
@@ -115,6 +118,7 @@
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Optional[String] $remote_user_environ_name = undef,
+  Integer[1, 24] $worker_count = min(8, $facts['processors']['count']),
```
ehelms, line 121: Is a worker per CPU recommended by Pulp team?

wbclark: @jlsherrill thoughts on this, or do you know who from the Pulp team would be able to answer? The default behavior I have set here is the same as what we did with Pulp2 but perhaps there are different considerations for Pulpcore.

ehelms: We were discussing this just the other day actually. We need to seek guidance from the Pulp community. @wbclark would you like to send an email to pulp-list@redhat.com to ask about Pulp 3 worker tuning?

manifests/init.pp
```diff
@@ -84,6 +84,9 @@
 # @param allowed_import_path
 #   Allowed paths that pulp can sync from using file:// protocol
 #
+# @param worker_count
+#   Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller. Enabling more than 8 workers, even with additional CPU cores available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
```
ehelms, line 88: Mind wrapping the help text 1-2 lines so the file is not rendered really wide?

manifests/init.pp
ehelms: Should the second number be 8? `Integer[1, 8]`

manifests/service.pp
```diff
@@ -24,10 +24,20 @@
...
+  Integer[1, $pulpcore::worker_count].each |$n| {
+    service { "pulpcore-worker@${n}":
+      ensure  => running,
+      enable  => true,
+      require => [Systemd::Unit_file['pulpcore-worker@.service'], Class['systemd::systemctl::daemon_reload']],
+    }
   }
 
+  # Prevent a bug where decreasing worker_count leaves excess workers defined and running
+  Integer[$pulpcore::worker_count + 1, 25].each |$n| {
```
ehelms, line 36: Should this be `8` instead of `25` given the minimum?

wbclark:
@ehelms I checked with several of the largest enterprise Katello users and found that some do indeed use more than 8 workers for Pulp 2. The largest count I'm aware of in production is one organization that uses 16 workers in their deployment. That is an environment with extensive hardware resources and I believe well over 65,000 Content Hosts.

For such use cases, I felt it was necessary to set an upper limit which is beyond what anyone could reasonably use.

So my intended behavior at the time that I wrote this is that default parameters will enable workers up to the number of logical CPUs up to the soft limit of 8, but users who understand the performance trade-offs should still have the option to increase that value up to a hard limit.

The hard limit wouldn't be strictly necessary at all, except that it provides a convenient way to disable and stop excess workers when the count is decreased by the user. With that said, perhaps there is a better pattern in Puppet for that but I'm not aware of it.

ehelms:
Sounds good, once tests are passing I'm good with this change.

manifests/init.pp
wbclark: Done, thanks

manifests/init.pp
```diff
@@ -115,6 +120,7 @@
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Optional[String] $remote_user_environ_name = undef,
+  Integer[1, 24] $worker_count = min(8, $facts['processors']['count']), # lint:ignore:parameter_documentation
```
ekohl, line 123: Why do you ignore this? It is documented now.

wbclark: I was getting a lint failure that `$facts['processors']['count']` is not documented. (Interesting to me that we didn't see the same issue with `$facts['networking']['fqdn']` three lines above)

wbclark:
Thanks @ekohl for the suggestions. I've pushed some changes and I'll explain these below:

1. I removed the #lint_ignore but I'm expecting it to complain due to a bug with the linter (we'll probably add that back for now while investigating a fix for the bug)

2. I'm taking the suggested approach of using a custom fact, which allows me to lift the arbitrary limit of 24 workers. While I think it's unlikely anyone would ever need more than 24 workers, I think the code is a bit cleaner this way.

3. TODO: I want to implement a test case of decreasing the worker count to have confidence about the above implementation

manifests/init.pp
ekohl: https://github.com/voxpupuli/puppet-lint-param-docs/pull/12 should address that.

lib/facter/pulpcore_workers.rb
```diff
@@ -0,0 +1,11 @@
+Facter.add(:pulpcore_workers) do
+  setcode do
+    directory = '/etc/systemd/system/multi-user.target.wants'
+    extension = 'pulpcore-worker@*.service'
+    return unless File.exist?(directory)
+
+    Dir[File.join(directory, "*#{extension}")].map do |config|
```
ekohl, line 7: Why is the `*` needed? `extension` already contains a glob (and is thus not really well named).
```suggestion
    Dir[File.join(directory, extension)].map do |config|
```

lib/facter/pulpcore_workers.rb
```diff
@@ -0,0 +1,11 @@
...
+      File.basename(config, extension)
```
ekohl, line 8: I don't think this works since it's not an extension but a glob pattern:
```console
$ irb
irb(main):001:0> File.basename('pulpcore-worker@2.service', 'pulpcore-worker@*.service')
=> "pulpcore-worker@2.service"
irb(main):002:0> File.basename('pulpcore-worker@2.service', '.service')
=> "pulpcore-worker@2"
```

manifests/init.pp
```diff
@@ -115,6 +120,7 @@
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Optional[String] $remote_user_environ_name = undef,
+  Integer[1] $worker_count = min(8, $facts['processors']['count']),
```
ekohl, line 123: Just thinking out loud: for the future it may actually make sense that you want to deploy a server that's only a webserver and do the worker processing elsewhere. So is `0` a better lower bound?

spec/classes/pulpcore_spec.rb
```diff
@@ -20,6 +20,24 @@
...
+            worker_count: 5
+          }
+        end
+
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_service('pulpcore-worker@5')
+            .with_ensure('running')
+            .with_enable('true')
+          is_expected.to contain_service('pulpcore-worker@6')
```
ekohl, line 35: For this to be true, you need to set the fact:
```ruby
let(:facts) { super().merge(pulpcore_workers: ['pulpcore-worker@6.service']) }
```

I'd also make this a separate it block since it's something different you want to ensure from the previous one. The test with worker 5 ensures worker_count sets the right number of running workers. Worker 6 tests that the purging happens.

If you want to test more edge cases, you'd test with:
* worker count 5
  * fact missing
  * fact present, but empty
  * fact with 6 workers
  * fact with 6 workers

wbclark: @ekohl how could the fact be missing? my understanding is that the fact should always be present, and only if the current worker count is zero should it be empty.

ekohl: The choice I made in puppet-foreman was not to generate a fact if the directory didn't exist. Similar, you could check for the existence of the `pulp-worker@.service` file. One benefit of not presenting it is that you don't generate the fact for every host in your Puppet environment. That helps with fact storage, though I don't know how much it really matters.

wbclark: Thanks for the explanation @ekohl . In that case I think it does matter, c.f. https://access.redhat.com/solutions/3764171

I'll check for `pulpcore-worker@service` in the fact definition

spec/classes/pulpcore_spec.rb
```diff
@@ -20,6 +20,58 @@
...
 
+      context 'with worker_count set to 5' do
+        let(:params) { { worker_count: 5 } }
+
+        context 'with 6 workers from fact' do
+          let(:facts) { super().merge(pulpcore_workers: (1..6).map { |n| "pulpcore-worker@#{n}.service" } ) }
+
+          it 'is expected to reduce the workers to 5' do
+            is_expected.to compile.with_all_deps
+            is_expected.to contain_service('pulpcore-worker@5')
```
wbclark, line 31: @ekohl could you recommend syntax to actually loop over workers 1 through 5 here rather than just checking the fifth worker?

ekohl: In Ruby you can do it:
```ruby
it 'is expected to reduce the workers to 5' do
  (1..5).each do |n|
    is_expected.to contain_service("pulpcore-worker@#{n}")
  end
end
```

wbclark: Thanks, I wasn't sure about nesting it within `it...do`

ekohl: https://www.rubyguides.com/2016/02/ruby-procs-and-lambdas/ might be interesting to read. The signature is roughly `it(message = nil, &block)`. Once you know that, it feels more familiar.

PR #102, opened 2020-06-25
use consistent style

wbclark:


spec/classes/plugin_certguard_spec.rb
```diff
@@ -6,11 +6,12 @@
       let(:facts) { os_facts }
 
       it { is_expected.to compile.with_all_deps }
-      case os
-      when /\-7-/
-         it { is_expected.to contain_package('python3-subscription-manager-rhsm').with(ensure: 'present') }
+      it { is_expected.to contain_pulpcore__plugin('certguard') }
+      if facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] == '7'
```
ekohl, line 10: The contents of `let` blocks are only available inside `it` blocks so you need to use the `os_facts` variable.
```suggestion
      if os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'] == '7'
```

evgeni:
We can, but @jlsherrill wanted to do it in the installer -- you'd have to ask him why.

Also the package is needed on el8 too, just that there it's built by centos/rhel, not us.

wbclark:
> We can, but @jlsherrill wanted to do it in the installer -- you'd have to ask him why.
> 
> Also the package is needed on el8 too, just that there it's built by centos/rhel, not us.

The reason for this is that the certguard plugin itself does not require `python3-subscription-manager-rhsm`, only the particular way Katello uses certguard requires that package.

ekohl:
Could you elaborate? Is there a use case where the certguard plugin is useful without `python3-subscription-manager-rhsm` being installed?

wbclark:
> Could you elaborate? Is there a use case where the certguard plugin is useful without `python3-subscription-manager-rhsm` being installed?

This is my understanding, though I'm not familiar with the details.

evgeni:
I "think* we need it on el8 too, not 💯 sure

jlsherrill:
yes, the pulp certguard plugin has two different 'guards', only one of them actually requires the rhsm package.  Thus the python package doesn't actually require it, and it seemed sensible to follow that through and not rquire it in the packaging.  

It is needed by el7 and el8, but for el7 we have to build it ourselves as @evgeni said

jlsherrill:
@ekohl makes sense to me!

ekohl:
https://github.com/theforeman/puppet-pulpcore/pull/103 fixes that. Let's merge that and rebase this.

ekohl:
Mind rebasing this now?

PR #105, opened 2020-06-26
Add full Apache management

ekohl:
The goal of this PR is to keep all Apache configuration within the module. There should be no Apache config fragments at all in puppet-foreman_proxy_content. There are parameters to attach fragments to an existing vhost. This makes it easy to switch between the co-located implementation (Foreman is the vhost) or standalone.

Currently this is very much a draft. Tests will fail and the proxy setup is not configured when attaching. It may also need parameters for HTTPS certificates.


manifests/apache.pp
```diff
@@ -1,34 +1,118 @@
...
+    false: {
+      $http_vhost_name = undef
     }
+    default: {
+      $http_vhost_name = $pulpcore::apache_http_vhost
 
-    if $facts['os']['selinux']['enabled'] {
-      selinux::boolean { 'httpd_can_network_connect': }
+      pulpcore::apache::fragment { 'pulpcore':
+        http_content => epp('pulpcore/apache-fragment.epp', {'proxy_pass' => [$proxy_pass_content]}),
```
ehelms, line 80: What is `epp` ?

manifests/plugin/file.pp
```diff
@@ -1,5 +1,17 @@
 # @summary Pulp File plugin
-class pulpcore::plugin::file {
+# @param redirect_legacy
+#   Whether to redirect the legacy (Pulp 2) URLs to the content server
+class pulpcore::plugin::file (
+  Boolean $redirect_legacy = false,
```
ehelms, line 5: I lean towards being more explicit here since legacy can be a generic term. Something like `use_pulp_2_content_route`

manifests/plugin/container.pp
```diff
@@ -1,6 +1,18 @@
 # @summary Pulp Container plugin
-class pulpcore::plugin::container {
+# @param location_prefix
+#   In the Apache configuration a location with this prefix is exposed. The
+#   version (currently v2) will be appended.
+class pulpcore::plugin::container(
+  String $location_prefix = '/pulpcore_registry',
```
ehelms, line 6: I'm not sure there is a reason to make this configurable as far as I can tell. This appears to then make the fragment deviate from the other fragments (https://github.com/theforeman/puppet-pulpcore/pull/105/files#diff-fb6d380303c53d97273a43f661d6d794) as it appears more generic and requires more developer overhead to understand it.

ehelms:
Something you may want to consider as part of this re-design is the ability to deploy a content app and its Apache configurations without deploying the API app.

manifests/apache.pp
ekohl: https://puppet.com/docs/puppet/latest/lang_template_epp.html

One benefit is that you explicitly pass in scope. This allows using the same fragment twice in the same manifest file. It also becomes clearer what variables are passed in. You can even type check parameters in the template itself rather than relying on it. Until now I've held off on changing all templates but here I want to do the same thing twice with different variables.

manifests/plugin/file.pp
ekohl: Good point and I like your suggestion.

manifests/plugin/container.pp
ekohl: Looking at that, I think I can actually generalize the template to redirect additional content URLs. I'm not sure if this needed to be configurable but in my experience having these present as variables is a good practice. I'll take a further look at this.

ekohl:
Yes, I agree. But then you also want to be able to deploy a content app without workers. By that point you're really blowing up the size of the PR.

ehelms:
This is already a large change so I don't mind it covering that use case :) If you want to plan to come back and layer that change on top of this design I'm OK with that too.

ekohl:
> Something you may want to consider as part of this re-design is the ability to deploy a content app and its Apache configurations without deploying the API app.

What's the goal of deploying one without the API? Shared filesystem with shared database to scale out? I suspect we may also need to support a shared Redis deployment then so I'd prefer to tackle it later. We can write acceptance tests with multiple servers so I'd hope to really set a new level of verification in this module. That does require some research on my side though.

ehelms:
Correct. With the Pulp 3 architecture, it's possible to deploy just a content app connected to the database of a main Pulp or mirror, as well as the storage (either shared filesystem or S3) for a lighter weight scale out. For that architecture, Redis is only needed by the main Pulp where the API app and workers are.

manifests/apache.pp
```diff
@@ -1,34 +1,118 @@
 # Configure an Apache vhost
 # @api private
-class pulpcore::apache {
+class pulpcore::apache (
+  String[1] $vhost_priority = '10',
```
ehelms, line 4: Is this something typically configured by a user?

ekohl: No, but it's a private class. There is 0 benefit to hardcoding it and only downsides if the user did modify it.

ekohl:
I've split off https://github.com/theforeman/puppet-pulpcore/pull/108 as a preparation PR which will reduce the size of this PR. That can already be merged.

I'm starting to feel fairly confident about this but no tests have been done on attaching to another vhost (which is currently the primary deployment). Hence it's still a draft.

ekohl:
Looks like EL8 doesn't generate a certificate by default where EL7 does so which is why they fail.

manifests/apache.pp
```diff
@@ -1,34 +1,122 @@
...
+
+  $proxy_pass_content = {
+    'path' => $content_path,
+    'url'  => $content_url,
+  }
+
+  $proxy_pass_api = {
+    'path' => $api_path,
+    'url'  => $api_url,
+  }
```
ehelms, line 55: Previously, these resulting ProxyPass declarations, when used with the Foreman vhost were included from a directory. That include was near the top of the vhost and therefore those ProxyPass rules got declared before the primary ones defined in the Foreman vhost (https://github.com/theforeman/puppet-foreman/blob/master/manifests/config/apache.pp#L110). These changes the ordering such that now the `/pulp` paths are unreachable since their ProxyPass declarations come after the these declarations.

Trouble is, can you force the proxy pass declarations on the vhost via fragment to appear in a particular order? We cannot drop `/pulp !` as since the Foreman's ProxyPass is a catchall on `/` we would not get `/pulp` routes connecting properly either.

manifests/apache.pp
```diff
@@ -1,34 +1,122 @@
...
+
+      include apache
+      include apache::mod::headers
+      apache::vhost { $https_vhost_name:
+        servername  => $pulpcore::servername,
+        port        => $https_port,
+        ssl         => true,
+        priority    => $vhost_priority,
+        docroot     => $pulpcore::webserver_static_dir,
+        directories => $base_directories + $api_directories,
```
ehelms, line 97: I think these also need to be passed to the `pulpcore::apache::fragment` use as well so that authentication is configured through remote user environ.

manifests/apache.pp
ekohl: As [discussed](https://github.com/theforeman/puppet-foreman_proxy_content/pull/271#issuecomment-654416991), it turns it this actually needs to be part of the public API.

manifests/apache.pp
ekohl: The concat module uses a priority on every fragment and then alphabetically sorts them to determine the order. Right now this doesn't use priority on fragments, but that would be the way to go.

puppet-foreman should not have to know about what other modules are adding to it. I'd like it if Foreman didn't have to specify `/pulp !` and we rely on inserting fragments before the actual ProxyPass statement.

ehelms: Does this mean there are changes needed here or on puppet-foreman to fix this?

ekohl: Probably not - we use the standard parameters on puppetlabs-apache. That order is probably [here](https://github.com/puppetlabs/puppetlabs-apache/blob/f8af65497f7198eac5e6eedd4bae529f561f2da4/manifests/vhost.pp#L2454-L2485). So using `order => 155` in our fragments would be a good thing to test with.

ekohl:
So a small update. I started with better integration tests:
* https://github.com/theforeman/puppet-pulpcore/pull/110 makes doing curl commands easier
* A client certificate is generated that can auth
* The API is verified as authenticated and non-authenticated to verify that works

Now that this appears to mostly work, I'm going to continue with testing deployment as part of the Foreman vhost.

manifests/apache.pp
```diff
@@ -1,34 +1,133 @@
...
-class pulpcore::apache {
+class pulpcore::apache (
+  Boolean $manage_selinux_boolean = true,
+  Stdlib::Port $http_port = 80,
+  Stdlib::Port $https_port = 443,
+  Hash[String, Any] $http_vhost_options = {},
+  Hash[String, Any] $https_vhost_options = {},
+  Enum['none', 'optional', 'require', 'optional_no_ca'] $ssl_verify_client = 'optional',
+) {
+  $vhost_priority = $pulpcore::apache_vhost_priority
```
ehelms, line 11: Should this be a param on this class? I am still not clear in what cases you want to tie it to the modules variables and when you want to make it available as a param on the class to pass in when it's a private API.

manifests/plugin.pp
```diff
@@ -20,4 +28,12 @@
       order   => '10',
     }
   }
+
+  if $http_content or $https_content {
+    # TODO: prio below main
```
ehelms, line 33: I wouldn't know what to do with this TODO. Could you expand on the details?

spec/acceptance/full_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      class { 'redis::globals':
+        scl => 'rh-redis5',
+      }
+    }
+
+    class { 'pulpcore':
+      remote_user_environ_name => 'HTTP_REMOTE_USER',
+      apache_https_cert        => '#{certdir}/ca-cert.pem',
+      apache_https_key         => '#{certdir}/ca-key.pem',
+      apache_https_ca          => '#{certdir}/ca-cert.pem',
```
ehelms, line 28: If this is the CA that signed the cert/key pair for Apache, shouldn't the option being passed in be `apache_https_chain` ?

manifests/plugin.pp
ekohl: I *think* this is where I need to solve the issue you saw in fpc where the content was added after the main reverse proxy so they ended up not being proxied.

spec/acceptance/full_spec.rb
ekohl: The CA can be used to identify client certs. Since we need client certs, we must set that option. I think chain is optional if it's the same as the CA but can be passed in if you want to allow client certs from one CA and serve a certificate from a different CA.

manifests/apache.pp
ekohl: I don't have a hard rule on when I make it a class parameter or inside the class. What's important to me is how coupled things are. If you use a lot from the other class and the private class ends up being unusable without the main class, it may not worth it. If there's a few, you can end up in a situation where you can run the rspec tests on just the private class without including the main class. That means it's smaller to compile and thus faster tests.

In this particular case I think I started with hardcoding it but later realized it should be a proper parameter.

spec/acceptance/full_spec.rb
ehelms: This probably does work as your tests are showing. If I go by the Apache docs:

```
 This can be used alternatively and/or additionally to SSLCACertificatePath.
```

This makes me think, for completeness of examples we should opt to specify both since it's quite confusing and the param documentation would imply that `apache_https_chain` should be set alongside cert and key. This also, as I understand it, ensures that the client will get the CA certificate to ensure trust on their end.

This works and its just for testing, but I do like when examples in testing show the best picture we can for how to use it and especially with SSL which can get confusing. So, I would prefer we specify both from my reading but wont block on it.

ekohl: I'm considering a real test of it: generate a cert to serve and a cert to auth. Verify that end to end.

Another thing to consider is that chain is deprecated. I think they want a full chain in the cert file but Apache on EL7 doesn't support that because it's too old.

However, my first goal is to even get the basic implementation for us to work. Right now I'm happy that I can test the client auth at all.

ekohl:
No changes in last push, just rebasing on a correct curl_command helper (https://github.com/theforeman/puppet-pulpcore/pull/110) since there were some bugs. Also https://github.com/theforeman/puppet-pulpcore/pull/114 was merged so that's no longer here either.

manifests/init.pp
```diff
@@ -91,19 +131,29 @@
...
-  Stdlib::Absolutepath $cache_dir = '/var/lib/pulp/tmp',
-  Stdlib::Absolutepath $config_dir = '/etc/pulp',
   String $user = 'pulp',
   String $group = 'pulp',
   Stdlib::Absolutepath $user_home = '/var/lib/pulp',
-  Boolean $manage_apache = true,
+  Stdlib::Absolutepath $config_dir = '/etc/pulp',
+  Stdlib::Absolutepath $cache_dir = '/var/lib/pulp/tmp',
+  Stdlib::Absolutepath $media_root = '/var/lib/pulp/media',
+  Stdlib::Absolutepath $static_root = '/var/lib/pulp/assets',
```
ekohl, line 143: SELinux prevents Apache from serving this directory since it has the content type `pulpcore_var_lib_t`. Going to change it `static` just to continue.

README.md
```diff
@@ -1,11 +1,38 @@
...
-**ALLOWED_IMPORT_PATHS** : This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests. For more information on this, see [https://docs.pulpproject.org/settings.html#allowed-import-paths](https://docs.pulpproject.org/settings.html#allowed-import-paths).
+The module requires at least Pulpcore 3.2.
 
-**AUTHENTICATION_BACKENDS , REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES** :
-The defaults that katello uses are defined in [templates/settings.py.erb](https://github.com/theforeman/puppet-pulpcore/blob/master/templates/settings.py.erb). For more information on these authentication settings, see [https://docs.pulpproject.org/installation/authentication.html](https://docs.pulpproject.org/installation/authentication.html)
+Currently only EL 7 & 8 are supported since this module depends on packages and within [foreman-packaging](https://github.com/theforeman/foreman-packaging) only RPM specs are present.
+
+## Installation layout
+
+Pulpcore doesn't madate a specific layout so this module creates and manages this. There are some constraints, mostly due to SELinux support.
```
ehelms, line 13: s/madate/mandate

ekohl:
Since there's a lot more issues with the deployment, I started https://pulp.plan.io/issues/7178 to discuss these with upstream. One thing to note is that currently our services run unconfined in SELinux.

ekohl:
This now includes https://github.com/theforeman/puppet-pulpcore/pull/110 & https://github.com/theforeman/puppet-pulpcore/pull/115 which should be reviewed first.

ehelms:
This is in a state of needing a rebase

README.md
```diff
@@ -1,11 +1,43 @@
...
-Puppet module for setting up Pulp 3 as part of Katello installation
 
-# Pulpcore 3.2
+Puppet module to set up Pulp 3, also known as Pulpcore and its various plugins. The goal of this module is to provide a Puppet alternative to [pulp_installer](https://github.com/pulp/pulp_installer). The primary goal is to allow consumption within [foreman-installer](https://github.com/theforeman/foreman-installer), but should not be limited to that use case.
 
-We are adding a few new settings to make the installer compatible with pulpcore 3.2. These settings are not available in releases prior to 3.2 and should not be used in earlier versions.
+## Compatibility
 
-**ALLOWED_IMPORT_PATHS** : This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests. For more information on this, see [https://docs.pulpproject.org/settings.html#allowed-import-paths](https://docs.pulpproject.org/settings.html#allowed-import-paths).
+The module requires at least Pulpcore 3.6.
```
ekohl, line 7: @jlsherrill I recall https://github.com/theforeman/puppet-foreman_proxy_content/commit/bd6654b038c6dc41518fa59469808a4a28a88a73 (which is sort of included in this PR) only worked with Pulpcore 3.6 so this would math it. Can you confirm this?

Going forward I would like this module to complete define the Pulp compatibility and puppet-foreman_proxy_content should only decide the compatibility on consuming some puppet-pulpcore version.

manifests/init.pp
ehelms: Given the intent of the module is to primarily support the `foreman-installer` use case, *and* we have a desire to expose this module as a top level module, what is the value in exposing these directories as primary inputs the module that a user could configure?

ekohl:
Note to self: per https://github.com/pulp/pulpcore-selinux/pull/21#discussion_r502465225 it appears like there's `/var/lib/pulp/pulpcore_static` that the Ansible installer uses as a document root for similar purposes. We should align on the exact directory name.

ehelms:
This can now be rebased as the required commit is merged.

templates/plugin-pulp2-content-routing.epp
```diff
@@ -0,0 +1,8 @@
+<%- | String[1] $path, String[1] $url | -%>
+  ProxyPass <%= $path %> <%= $url %>
```
ehelms, line 2: Is it possible to have a pattern of validation, empty line, template code? e.g.

```
<%- | String[1] $path, String[1] $url | -%>

  ProxyPass <%= $path %> <%= $url %>
```

This pattern of declaration would improve readability for me which can often be hard to follow when we are trying to combine ERB + indentation that is designed to render the *final* file properly nested.

ekohl:
Note to self: testing pipeline using forklift
```
LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 ansible-playbook pipelines/install_pipeline.yml -e pipeline_os=centos7 -e pipeline_type=katello -e pipeline_version=nightly -e foreman_installer_module_branches="['ekohl/pulpcore/apache-management','ekohl/foreman_proxy_content/remove-pulpcore-templates']"
```

ekohl:
I've started doing testing and this appears to work well now. I'm doing a final EL7 pipeline to verify it end to end. EL8 is untested, but I noticed that on EL8's Apache we need to do more work. https://pulp-certguard.readthedocs.io/en/latest/reverse_proxy_config.html#apache-2-4-10-config-example states we need to encode it.

templates/apache-fragment.epp
```diff
@@ -0,0 +1,53 @@
+<%- |
```
ehelms, line 1: This is painful to read and parse as a developer. Is there empty lines we could add to separate blocks more? Or processing that can happen inside the puppet class rather than in the template? I think, if we are going to go this route of trying to have a generic fragment that handles *all* cases, rather than having easy to read, specific templates we need to make the template easier to read and accept that the rendered version may not look like you would exactly design an Apache config file indentation and grouping of parameters wise.

ekohl: I should have mentioned this, but it's a trimmed down version of https://github.com/puppetlabs/puppetlabs-apache/blob/main/templates/vhost/_directories.erb with strict input checking. I could probably avoid it altogether and use `template('apache/vhost/_directories.erb')` but I wasn't quite sure about scoping.

I also wasn't quite sure about best practices. Room for improvement but I liked the strict typing checks.

I'll probably submit this upstream in puppetlabs-apache, but in the short term I don't think we can rely on that.

ekohl:
I think the tests should now be green, but Travis is backed up. After that, I think this is pretty much ready to be merged.

templates/apache-fragment.epp
```diff
@@ -0,0 +1,60 @@
...
+  Optional[Array[Hash[String, String]]] $proxy_pass = undef,
+| -%>
+<%-
+$directories.each |$directory| {
+  if $directory['provider'] =~ /^(.*)match$/ {
+    $provider = $1.capitalize + 'Match'
+  } else {
+    $provider = $directory['provider'].capitalize
+  }
+-%>
```
ehelms, line 23: This is hard to parse in this in this syntax. I think it could benefit from:

```
<%- $directories.each |$directory| { -%>
  <%- if $directory['provider'] =~ /^(.*)match$/ {
      $provider = $1.capitalize + 'Match'
    } else {
      $provider = $directory['provider'].capitalize
    }
  -%>
```

templates/apache-fragment.epp
```diff
@@ -0,0 +1,60 @@
...
+    Optional[request_headers] => Array[String[1]],
+    Optional[proxy_pass] => Array[Struct[{
+      url => String[1],
+      Optional[params] => Hash[String, Variant[String, Integer]],
+      Optional[reverse_urls] => Array[String[1]],
+    }]],
+  }]] $directories,
+  Optional[Array[Hash[String, String]]] $proxy_pass = undef,
+| -%>
+<%-
```
ehelms, line 16: Could this be:

```
| -%>

<%-
```

ehelms:
I had green install pipeline with this change. I just have questions regarding the EPP template formatting

ekohl:
How about this? I kept it as a separate commit to make the diff easier to read, but will squash if you agree.

ehelms:
Yes, that helps a lot, thank you! I can squash on merge once the tests run.

ehelms:
Trying to kick Travis

ehelms (approved):

PR #106, opened 2020-06-26
Use the non-deprecated namedspaced pg function

ekohl:


wbclark:
> I'm not sure if this is backwards incompatible - I'd say enhancement. What makes you think it's incompatible? The min version bump?

Exactly. Someone using this module with puppetlabs/postgresql v4.2.0 would no longer be able to do that.

PR #110, opened 2020-07-09
Introduce a curl_command helper

ekohl:
This aims to reduce the duplication in curl verifications and also allows checking both the body and the response code.

Since this is also usable outside of puppet-pulpcore, I aim to add this in a reusable place. This PR is here to verify it works in CI and share.

### DIFF
--- a/spec/support/acceptance/curl_command.rb
+++ b/spec/support/acceptance/curl_command.rb
@@ -1,0 +1,63 @@
+# https://github.com/mizzy/serverspec/pull/611 was rejected so adding it here.
+begin
+  require 'serverspec'
+
+  module Serverspec
+    module Type
+      class CurlCommand < Command
+        def response_code
+          m = /Response-Code: (?<code>\d+)/.match(stderr)
+          return 0 unless m
+          m[:code].to_i
+        end
+
+        def body
+          command_result.stdout
+        end
+
+        def body_as_json
+          MultiJson.load(body)
+        end
+
+        private
+
+        def curl_command
+          command = "curl --silent --write-out '%{stderr}Response-Code: %{response_code}\\n' '#{@name}'"
+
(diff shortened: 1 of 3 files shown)


spec/support/acceptance/curl_command.rb
```diff
@@ -0,0 +1,48 @@
...
+      end
+
+      def body_as_json
+        MultiJson.load(body)
+      end
+
+      private
+
+      def curl_command
+        command = "curl --silent --write-out '%{stderr}Response-Code: %{response_code}\\n' '#{@name}'"
```
ekohl, line 21: curl 7.70.0 introduces `--write-out %{json}` to output everything as JSON but even Fedora 32 doesn't ship that yet

ekohl:
I intend to submit this to serverspec itself - this is now just a quick workaround.

ekohl:
Submitted as https://github.com/mizzy/serverspec/pull/611

spec/support/acceptance/curl_command.rb
```diff
@@ -0,0 +1,63 @@
+# https://github.com/mizzy/serverspec/pull/611 was rejected so adding it here.
```
wbclark, line 1: One of the rejected serverspec PRs had the recommendation to use https://github.com/ryotarai/infrataster for this instead

Might be worth checking out longer term

ekohl: That repo hasn't seen activity since 2017. That concerns me.

ekohl:
The benefits are in https://github.com/theforeman/puppet-pulpcore/pull/105/files#diff-52c5938dc73ceef10765c19ac5ae35bcR76-R92. And infrataster appears to be dead. Long term we may want to move away from serverspec entirely.

PR #113, opened 2020-07-16
Fixes #30423 - match default directories in pulp_installer

wbclark:


manifests/init.pp
```diff
@@ -102,8 +105,9 @@
   Stdlib::Port $api_port = 24817,
   Stdlib::Host $content_host = '127.0.0.1',
   Stdlib::Port $content_port = 24816,
-  Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
-  Stdlib::Absolutepath $pulp_static_root = '/var/lib/pulp/assets',
+  Stdlib::Absolutepath $apache_docroot = $user_home,
```
ekohl, line 108: Setting the docroot to the user home feels like a bad idea, but I guess it matches what Pulp does.

manifests/init.pp
```diff
@@ -102,8 +105,9 @@
...
+  Stdlib::Absolutepath $pulpcore_media_root = $user_home,
+  Stdlib::Absolutepath $pulpcore_static_root = "${user_home}/assets",
```
ekohl, line 110: I ran into this with my Apache PR as well. While it's the same, I think this would be clearer in the intent. It's because the assets need to be served to the end user.
```suggestion
  Stdlib::Absolutepath $pulpcore_static_root = "${apache_docroot}/assets",
```

jlsherrill:
can't speak for the puppet particulars, but this looks good to me

ekohl:
In https://github.com/theforeman/puppet-pulpcore/pull/105 I've started to how I think it should be stored and configure Apache to also serve the static root at the correct static url. By having these configured, we can also start to serve the static content in a Katello scenario. It's not working entirely even though I just had it working. Maybe it's time for the weekend now.

manifests/init.pp
```diff
@@ -102,8 +105,9 @@
   Stdlib::Port $api_port = 24817,
   Stdlib::Host $content_host = '127.0.0.1',
   Stdlib::Port $content_port = 24816,
-  Stdlib::Absolutepath $webserver_static_dir = '/var/lib/pulp/docroot',
-  Stdlib::Absolutepath $pulp_static_root = '/var/lib/pulp/assets',
+  Stdlib::Absolutepath $apache_docroot = $user_home,
+  Stdlib::Absolutepath $pulpcore_media_root = $user_home,
+  Stdlib::Absolutepath $pulpcore_static_root = "${apache_docroot}/assets",
```
ekohl, line 110: I just found out that Apache refuses to serve `pulpcore_var_lib_t` so this needs another SELinux policy change. We should agree on a layout and properly label that.

ekohl:
I've written https://github.com/theforeman/puppet-pulpcore/pull/115 that implements https://pulp.plan.io/issues/7178.

ekohl:
Closing in favor of https://github.com/theforeman/puppet-pulpcore/pull/115

PR #115, opened 2020-07-21
Fixes #30423 - Change the application layout

ekohl:
This explicitly sets the all directories and documents the layout. It diverges from the upstream defaults in that MEDIA_ROOT is set to a subdirectory and the directory permissions are stricter than upstream.  The assets are served in static since this works around some SELinux issue.

Upstream this layout is proposed as the default. https://pulp.plan.io/issues/7178 has been opened for that.

It's an alternative to https://github.com/theforeman/puppet-pulpcore/pull/113. It will need a migration in the installer to move any generated content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,3 +9,39 @@ The module provides no guarantee for multiple versions. Whenever a version is dr
 ### Pulpcore 3.6
 
 Due to the use of libexec wrappers, at least python3-pulpcore 3.6.3-2 must be installed
+
+Supports EL 7 & 8, following [pulpcore-packaging](https://github.com/theforeman/pulpcore-packaging) only RPM specs are present.
+
+## Installation layout
+
+Pulpcore doesn't mandate a specific layout so this module creates and manages this. There are some constraints, mostly due to SELinux support.
+
+As part of the installation, it creates a user (default `pulp`) and group (default `pulp`). This user gets a home directory (default `/var/lib/pulp`). There is also a config dir (default `/etc/pulp`) under which a `settings.py` file is created.
+
+The media root (default `/var/lib/pulp/media`) refers to the [MEDIA_ROOT setting](https://docs.djangoproject.com/en/2.2/ref/settings/#media-root). In Pulp this should not be served by Apache. Instead of [MEDIA_URL](https://docs.djangoproject.com/en/2.2/ref/settings/#media-url) Pulp has a dedicated `pulpcore-content` service which can also perform permission checks. Only the Pulp services need to read the files so directory permissions are set to `0750`. Note this default differs from [Pulp's default](https://docs.pulpproject.org/settings.html#media-root). A subdirectory of the home directory allows a stricter lockdown and avoids any risk of uploading media files into the wrong directory.
+
+There are also the [STATIC_ROOT](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-STATIC_ROOT) and [STATIC_URL](https://docs.djangoproject.com/en/2.2/ref/settings/#static-url) settings. These serve the static assets used by Pulp. This includes CSS and Javascript for the HTML pages. They're not needed for the application to function, but make browsing the API more convenient.
+
+These is also the `cache_dir` which is used to configure [WORKING_DIRECTORY](https://docs.pulpproject.org/settings.html#working-directory) and [FILE_UPLOAD_TEMP_DIR](https://docs.djangoproject.com/en/2.2/ref/settings/#file-upload-temp-dir). This defaults to `/var/lib/pulp/tmp`. It is strongly recommended that this is on the same filesystem as `MEDIA_ROOT`.
+
+There is also `chunked_upload_dir` to configure the undocumented `CHUNKED_UPLOAD_DIR`. This directory stores the temporary files used for files uploaded as chunks.
+
+Apache is configured to use an empty directory as docroot (`$apache_docroot`, default `/var/lib/pulp/docroot`). Doing so prevents Apache from bypassing the Pulp content app. When Apache is not managed, this directory is not managed.
+
+While Pulp can create most of these directories at runtime, they're explicitly managed to set the correct permissions and, if pulpcore-selinux is installed, enforce the correct labels.
+
+This results into the following structure, using `tree -pug`:
+```
+/
+├── [drwxr-xr-x root     root    ]  etc
+│   └── [drwxr-xr-x root     pulp    ]  pulp ($config_dir)
+│       └── [-rw-r----- root     pulp    ]  settings.py
+└── [drwxr-xr-x root     root    ]  var
+    └── [drwxr-xr-x root     root    ]  lib
+        └── [drwxrwxr-x pulp     pulp    ]  pulp ($user_home)
+            ├── [drwxr-xr-x pulp     pulp    ]  assets ($static_root)
+            ├── [drwxr-xr-x pulp     pulp    ]  docroot ($apache_docroot)
+            ├── [drwxr-x--- pulp     pulp    ]  media ($media_root)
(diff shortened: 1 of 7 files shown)


ekohl:
The migration can be done by moving `/var/lib/pulp/docroot` to `/var/lib/pulp/media` in `foreman-installer`, unless `media` already exists.

manifests/config.pp
```diff
@@ -20,16 +23,41 @@
     order   => '01',
   }
 
-  file { [$pulpcore::user_home, $pulpcore::webserver_static_dir, $pulpcore::cache_dir]:
+  file { $pulpcore::user_home:
     ensure => directory,
     owner  => $pulpcore::user,
     group  => $pulpcore::group,
     mode   => '0775',
```
ekohl, line 30: So I've learned this may give problems when sharing this directory with Pulp 2. We don't add the Apache user to the pulp group. Looking into this now.

ekohl: So it turns out Pulp 2's spec file does this: https://github.com/pulp/pulp-packaging/blob/7c05ec9c678c01aa9c6a40f8e3a8be4b1aa9c50d/packages/pulp/pulp.spec#L500

It's lovely that `/var/lib/pulp` is also owned by pulp-server but with different ownership:
https://github.com/pulp/pulp-packaging/blob/7c05ec9c678c01aa9c6a40f8e3a8be4b1aa9c50d/packages/pulp/pulp.spec#L484-L485

This means `rpm -qV pulpserver` will result in warnings.

wbclark: If I understand correctly, this means your implementation should work whether it's working alongside pulp2 or a future new install that is pulp3 only? Because the group membership is only required by Pulp2 and is handled by Pulp2 packaging.

ekohl: Yes, I wasn't aware Pulp 2 packaging did that and thought it was something the installer needed to do.

manifests/init.pp
```diff
@@ -89,21 +107,31 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
-# @example
+# @example Default configuration
 #   include pulpcore
+#
+# @example Disable static content
```
wbclark, line 113: @ekohl can you help me understand this use case? why would an user want to disable management of static content ?

ekohl: Static content is only the CSS/JS and you don't need that for the API. Right now in our deployment we don't even serve it and we've shipped that to users.

wbclark: @ekohl My understanding is that it does allow for a nicer presentation of the api docs... Is that something that we don't expect a Katello user to interact with directly?

wbclark: My thinking is that from a troubleshooting standpoint that would be nice to have in the Katello scenario, particularly given that there isn't a `pulp-admin` equivalent for Pulp3.

ekohl: I think so and it is a goal, but right now we don't serve them at all and this allows a lighter deployment.

README.md
```diff
@@ -1,11 +1,43 @@
...
+
+These is also the `cache_dir` which is used to configure [WORKING_DIRECTORY](https://docs.pulpproject.org/settings.html#working-directory) and [FILE_UPLOAD_TEMP_DIR](https://docs.djangoproject.com/en/2.2/ref/settings/#file-upload-temp-dir). This defaults to `/var/lib/pulp/tmp`. It is strongly recommended that this is on the same filesystem as `MEDIA_ROOT`.
+
+There is also `chunked_upload_dir` to configure the undocumented `CHUNKED_UPLOAD_DIR`. This directory stores the temporary files used for files uploaded as chunks.
+
+When the Apache vhost is managed, `apache_docroot` (default `/var/lib/pulp/docroot`) is also created. This doesn't contain any files, but this makes sure that there aren't any files being served.
```
ehelms, line 25: I think there is some nuance of Apache embedded in this wording that makes it hard to know what this means. I can sum this up with: why do I care if files are being served or not?

ekohl: You should care that no files are served. Turns out that serving any files from Apache is a misconfiguration.

manifests/init.pp
ekohl: After considering this and seeing that Pulp reads the static files on application startup, I've decided to drop the option to not generate these. I have moved it to its own class so it doesn't restart all applications if anything changes. Thinking about it, it should probably restart the API service if it reads the files.

ekohl:
I had some conversations with upstream. Right now I want to agree on the best layout and ensure we're not going to have to migrate twice. This PR has been very useful for me to get my thoughts straight, but moving back to a draft for now.

README.md
ehelms: Would another way to phrase this be:

Apache requires that a docroot be configured. By configuring this to an empty directory, we prevent Apache from being configured to accidentally server Pulp content outside of the Pulp content app?

ekohl: Yes, that is a much better way to phrase it.

wbclark:
@ekohl @jlsherrill how immediate is the need to get this straightened out?

Should I close out https://github.com/theforeman/foreman-installer/pull/543 , https://github.com/theforeman/puppet-foreman_proxy_content/pull/275 , and https://github.com/theforeman/puppet-pulpcore/pull/113 while we settle on a definitive layout?

Or is it better to move forward with those while this one is still in progress ?

ekohl:
https://github.com/pulp/pulpcore/pull/799#issuecomment-684957879 would be roughly the migration plan.

We can do the same in a `pre` hook. That should be after our SELinux hook, though it's probably not that relevant. On non-migrated systems the directory will not be present. On upgraded systems it should already be present.

Puppet will take care of the config update + application restarts.

The cleanup of the symlink should happen in a `post` hook.

ekohl:
Updated. I now believe this is mostly ready. I'm going to work on an installer migration to migrate the content.

manifests/config.pp
```diff
@@ -20,13 +23,20 @@
...
   }
 
+  file { [$pulpcore::cache_dir, $pulpcore::chunked_upload_dir, $pulpcore::media_root]:
+    ensure => directory,
+    owner  => $pulpcore::user,
+    group  => $pulpcore::group,
+    mode   => '0770',
```
wbclark, line 37: `README.md` states that the `media_root` directory has `0750` permissions. Does that need to be split off here, or updated in the `README.md` to `0770` instead?

ekohl: Good point. I think 750 should be sufficient and it was a copy paste error from the home dir (I did fix 5 -> 0, forgot about the group). Will update.

ekohl:
Migration in https://github.com/theforeman/foreman-installer/pull/583.

ekohl:
Note to self: check SELinux context on `/var/lib/pulp/media`

ekohl:
https://github.com/pulp/pulpcore-selinux/pull/23

manifests/static.pp
```diff
@@ -11,6 +11,6 @@
   pulpcore::admin { 'collectstatic --noinput':
     refreshonly => true,
     subscribe   => Concat['pulpcore settings'],
-    require     => File[$pulpcore::pulp_static_root],
+    require     => File[$pulpcore::static_root],
```
wbclark, line 14:
```suggestion
    subscribe     => File[$pulpcore::static_root],
```
As discussed on IRC, there is a very small possibility that owner/group could be changed. In that case we would want to re-run collectstatic so that the collected static content will also be changed to match.

In most regular use cases that is unlikely to happen, but that also means the cost of making this change (in terms of execution time) is negligible since the refresh would only get triggered in that edge case.

wbclark (approved):
This worked well in my testing

PR #116, opened 2020-07-22
Fixes #30465 - Use libexec wrappers for SELinux

ekohl:
In python3-pulpcore 3.7.1-2 the `/usr/libexec/pulpcore` wrappers have been introduced to enter the proper SELinux domain.

It has also been cherry picked to 3.6.3-2 but in the SELinux policy is incomplete so it has no effect. The main benefit of that cherry pick is to keep the module compatible with both 3.6 and 3.7.

The README is updated with a support policy. Since 3.7 is not being tested in CI yet, I have not yet added it to the supported versions.

### DIFF
--- a/templates/pulpcore-api.service.erb
+++ b/templates/pulpcore-api.service.erb
@@ -10,7 +10,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=~
 RuntimeDirectory=pulpcore-api
-ExecStart=/usr/bin/gunicorn pulpcore.app.wsgi:application \
+ExecStart=/usr/libexec/pulpcore/gunicorn pulpcore.app.wsgi:application \
           --bind '<%= scope['pulpcore::api_host'] %>:<%= scope['pulpcore::api_port'] %>' \
           --access-logfile -
 ProtectSystem=full
--- a/templates/pulpcore-resource-manager.service.erb
+++ b/templates/pulpcore-resource-manager.service.erb
@@ -11,7 +11,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=~
 RuntimeDirectory=pulpcore-resource-manager
-ExecStart=/usr/bin/rq worker \
+ExecStart=/usr/libexec/pulpcore/rq worker \
           -w pulpcore.tasking.worker.PulpWorker -n resource-manager \
           -c 'pulpcore.rqconfig' \
           --disable-job-desc-logging
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # puppet-pulpcore
-Puppet module for setting up Pulp 3 as part of Katello installation
 
-# Pulpcore 3.2
+Puppet module to set up Pulp 3. The primary goal of the maintainers is to set up Pulp 3 as part of Katello installation, but there's no reason it couldn't be used elsewhere.
 
-We are adding a few new settings to make the installer compatible with pulpcore 3.2. These settings are not available in releases prior to 3.2 and should not be used in earlier versions.
+## Support policy
 
-**ALLOWED_IMPORT_PATHS** : This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests. For more information on this, see [https://docs.pulpproject.org/settings.html#allowed-import-paths](https://docs.pulpproject.org/settings.html#allowed-import-paths).
+The module provides no guarantee for multiple versions. Whenever a version is dropped, the major version is increased. All supported versions are listed.
 
-**AUTHENTICATION_BACKENDS , REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES** :
-The defaults that katello uses are defined in [templates/settings.py.erb](https://github.com/theforeman/puppet-pulpcore/blob/master/templates/settings.py.erb). For more information on these authentication settings, see [https://docs.pulpproject.org/installation/authentication.html](https://docs.pulpproject.org/installation/authentication.html)
+### Pulpcore 3.6
+
+Due to the use of libexec wrappers, at least python3-pulpcore 3.6.3-2 must be installed
--- a/templates/pulpcore-content.service.erb
+++ b/templates/pulpcore-content.service.erb
@@ -10,7 +10,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=~
 RuntimeDirectory=pulpcore-content
-ExecStart=/usr/bin/gunicorn pulpcore.content:server \
+ExecStart=/usr/libexec/pulpcore/gunicorn pulpcore.content:server \
           --bind '<%= scope['pulpcore::content_host'] %>:<%= scope['pulpcore::content_port'] %>' \
           --worker-class 'aiohttp.GunicornWebWorker' \
           -w 2 \
--- a/templates/pulpcore-worker@.service.erb
+++ b/templates/pulpcore-worker@.service.erb
@@ -13,7 +13,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=~
 RuntimeDirectory=pulpcore-worker-%i
-ExecStart=/usr/bin/rq worker \
+ExecStart=/usr/libexec/pulpcore/rq worker \
           -w pulpcore.tasking.worker.PulpWorker \
           -c 'pulpcore.rqconfig' \
           --disable-job-desc-logging


ekohl:
So it turns out #115 is not needed for this. I verified this manually on an EL7 VM (using the beaker tests).

templates/pulpcore-resource-manager.service.erb
```diff
@@ -4,14 +4,15 @@ After=network-online.target
...
 Environment="PULP_SETTINGS=<%= scope['pulpcore::settings_file'] %>"
 User=<%= scope['pulpcore::user'] %>
-WorkingDirectory=/var/run/pulpcore-resource-manager/
+Group=<%= scope['pulpcore::group'] %>
+WorkingDirectory=%t/pulpcore-resource-manager
 RuntimeDirectory=pulpcore-resource-manager
-ExecStart=/usr/bin/rq worker \
+ExecStart=/usr/libexec/pulpcore/rq worker \
           -w pulpcore.tasking.worker.PulpWorker -n resource-manager \
-          --pid=/var/run/pulpcore-resource-manager/resource-manager.pid \
```
ekohl, line 14: These pid files gave SELinux errors and https://python-rq.org/patterns/systemd/ doesn't suggest it needs pid files.

Probably because `/var/run` is not being labeled correctly since it's a symlink to `/run` making https://github.com/pulp/pulpcore-selinux/blob/64b99b463186305a371498ce1da8d1f3bb956a20/pulpcore.fc#L12-L16 useless.

templates/pulpcore-api.service.erb
```diff
@@ -8,9 +8,10 @@ Environment="DJANGO_SETTINGS_MODULE=pulpcore.app.settings"
 Environment="PULP_SETTINGS=<%= scope['pulpcore::settings_file'] %>"
 Environment="PULP_STATIC_ROOT=<%= scope['pulpcore::pulp_static_root'] %>"
 User=<%= scope['pulpcore::user'] %>
-PIDFile=/run/pulpcore-api.pid
+Group=<%= scope['pulpcore::group'] %>
+WorkingDirectory=%t/pulpcore-api
```
ekohl, line 12: I *think* %t resolves to `/run` when it runs as a system service. I'm a bit unsure about this. I've considered setting it to `~` or explicit `/var/lib/pulp`. That way systemd also automatically ensures `/var/lib/pulp` is present and mounted. I'm not sure why these services should have a different workdir.

ekohl:
When running this, you do see AVCs:
```console
# grep type=AVC /var/log/audit/audit.log 
type=AVC msg=audit(1595519054.448:2345): avc:  denied  { search } for  pid=28778 comm="gunicorn" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
type=AVC msg=audit(1595519054.448:2346): avc:  denied  { search } for  pid=28778 comm="gunicorn" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
type=AVC msg=audit(1595519054.723:2347): avc:  denied  { read } for  pid=28822 comm="sh" name="meminfo" dev="proc" ino=4026532028 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:proc_t:s0 tclass=file permissive=0
type=AVC msg=audit(1595519058.657:2361): avc:  denied  { search } for  pid=28799 comm="rq" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
type=AVC msg=audit(1595519058.657:2362): avc:  denied  { search } for  pid=28799 comm="rq" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
```

manifests/service.pp
```diff
@@ -2,22 +2,37 @@
...
 class pulpcore::service {
 
+  ['gunicorn', 'rq'].each |$bin| {
+    file { "/usr/libexec/pulpcore/${bin}":
+      ensure  => file,
+      content => "#!/bin/bash\nexec ${bin} \"$@\"\n",
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0755',
+      # https://github.com/pulp/pulpcore-selinux/pull/13
```
ehelms, line 12: This is not clear to me if it means this exists until the pull request is merged or depends upon the pull request. Can you update the wording in the comment to clarify?

ekohl: That PR would add the default, now we need to explicitly set it.

ekohl:
@mikedep333 mind having a look as well? In particular the systemd unit files.

manifests/config.pp
```diff
@@ -30,6 +30,14 @@
   pulpcore::admin { 'collectstatic --noinput':
     refreshonly => true,
     subscribe   => Concat['pulpcore settings'],
+    require     => File[$pulpcore::pulp_static_root],
+  }
+
+  file { '/usr/libexec/pulpcore':
```
ehelms, line 36: Might be worth turning this path into an internal variable due to it's usage throughout.

manifests/service.pp
```diff
@@ -2,22 +2,37 @@
 # @api private
 class pulpcore::service {
 
+  ['gunicorn', 'rq'].each |$bin| {
+    file { "/usr/libexec/pulpcore/${bin}":
+      ensure  => file,
+      content => "#!/bin/bash\nexec ${bin} \"$@\"\n",
+      owner   => 'root',
+      group   => 'root',
+      mode    => '0755',
```
ehelms, line 11: Is worldable executable applicable in our case?

templates/pulpcore-api.service.erb
ehelms: I'm unclear from the comment what we want to end up with. If `/var/lib/pulp` is the proper working directory then I vote set it explicitly to that.

ehelms:
I'm still not clear on why the wrappers are needed versus using the system paths.

ekohl:
> I'm still not clear on why the wrappers are needed versus using the system paths.

What you execute needs to have some SELinux label. IMHO Pulpcore doesn't own `/usr/bin/{gunicorn,rq}` so a label there feels wrong.

manifests/config.pp
ekohl: I also thought about that but didn't manage to update it before the weekend started. Will update.

manifests/service.pp
ekohl: It doesn't hurt since there's nothing secret in there.

templates/pulpcore-api.service.erb
ekohl: I think `%t` resolves to `/run` for system services. This is to preserve the previous `/var/run` (and on all modern systems that's a symlink to `/run`. I'm leaning to `~` or `/var/lib/pulp` because systemd will ensure it's mounted which can be very useful if it comes from NFS or a separate partition (which is very common), but I might missing something.

Here I'd like some feedback from Pulp upstream since I don't know what their preference is.

manifests/service.pp
ehelms: Sorry, I'm still not clear the implications on this PR. Does this PR need the pulpcore-selinux PR?

manifests/service.pp
ehelms: Does it imply any other user could run those commands? Starting any rogue services?

ehelms:
This is just my lack of knowledge in the SELinux area. Is it common to create wrappers? It feels generally safe to assume, or rather to enforce that Pulpcore is the only thing using `/usr/bin/gunicorn,rq` on a box that it is installed.

This question is geared at:

 1) Are we following standard SELinux practices
 2) Are we deviating from upstream Pulp too much

I'm checking on these more than I am trying to comment on them one way or another.

manifests/service.pp
ekohl: Yes, but they can already do that today using `/usr/bin/gunicorn` so there is no difference. I have a suggestion to use unix sockets instead of TCP ports but the current SELinux policy in Pulpcore doesn't allow that now (https://pulp.plan.io/issues/7189). By using systemd socket activation and letting systemd create `/run/pulpcore-api.sock` (or similar) owned by `apache:root` and mode `0600` we can prevent any local user from spinning up a rogue service. It also prevents local users from impersonating admin by connecting directly to the TCP port.

manifests/service.pp
ekohl: No, it does not need that PR. After that PR is released, the line below it can be dropped.

ekohl:
> This is just my lack of knowledge in the SELinux area. Is it common to create wrappers?

Yes:
```console
# ls -lZ /usr/share/foreman/bin/rails /usr/libexec/foreman/sidekiq-selinux
-rwxr-xr-x. root root system_u:object_r:foreman_rails_exec_t:s0 /usr/libexec/foreman/sidekiq-selinux
-rwxr-xr-x. root root system_u:object_r:foreman_rails_exec_t:s0 /usr/share/foreman/bin/rails
```

Note that setting a context on a file is not security thing. As a normal non-privileged user you can set it:
```console
$ touch test
$ chcon -t foreman_rails_exec_t test
$ ls -lZ test
-rw-rw-r--. ekohl ekohl unconfined_u:object_r:foreman_rails_exec_t:s0 test
```

> It feels generally safe to assume, or rather to enforce that Pulpcore is the only thing using `/usr/bin/gunicorn,rq` on a box that it is installed.

I don't think it is. You may need another Python service with another SELinux context. By not claiming any specific resources in a shared path, you keep it open in the future.

> Are we following standard SELinux practices

I think we are.

> Are we deviating from upstream Pulp too much

I suggested the wrappers and upstream replied with "This is a really good idea IMO.". See https://pulp.plan.io/issues/7178#note-3 (last paragraph).

It also allows upstream to use the same wrapper locations in their setup with a virtualenv. This allows further standardization of the systemd unit files. I still hope that one day we can share those between both projects rather than having them duplicated in two places.

ehelms:
Thanks for the explanations. All sounds good. Are we waiting on Pulp to merge this?

ekohl:
I'm hoping to get an answer about [WorkingDirectory](https://github.com/theforeman/puppet-pulpcore/pull/116#discussion_r459559607) and the [generated denials](https://github.com/theforeman/puppet-pulpcore/pull/116#issuecomment-663092953).

Then I'll also introduce the [execdir variable](https://github.com/theforeman/puppet-pulpcore/pull/116#discussion_r460867067).

ekohl:
@dkliban @mikedep333 could you have a look at these 2? I should have mentioned this earlier today

> I'm hoping to get an answer about [WorkingDirectory](https://github.com/theforeman/puppet-pulpcore/pull/116#discussion_r459559607) and the [generated denials](https://github.com/theforeman/puppet-pulpcore/pull/116#issuecomment-663092953).

templates/pulpcore-api.service.erb
mikedep333: On Pulp upstream, we never considered the `$PWD` of the pulp processes.

I agree with `~` (pulp_installer defaults the pulp  home dir to `/var/lib/pulp`, but lets users change it), in case we do end up writing anything there, and so that it gets mounted.

mikedep333:
> When running this, you do see AVCs:
> 
> ```
> # grep type=AVC /var/log/audit/audit.log 
> type=AVC msg=audit(1595519054.448:2345): avc:  denied  { search } for  pid=28778 comm="gunicorn" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
> type=AVC msg=audit(1595519054.448:2346): avc:  denied  { search } for  pid=28778 comm="gunicorn" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
> type=AVC msg=audit(1595519054.723:2347): avc:  denied  { read } for  pid=28822 comm="sh" name="meminfo" dev="proc" ino=4026532028 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:proc_t:s0 tclass=file permissive=0
> type=AVC msg=audit(1595519058.657:2361): avc:  denied  { search } for  pid=28799 comm="rq" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
> type=AVC msg=audit(1595519058.657:2362): avc:  denied  { search } for  pid=28799 comm="rq" name="httpd" dev="vda1" ino=34201106 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_config_t:s0 tclass=dir permissive=0
> ```

@zpytela is working on updating our policies right now. He should be able to help.

I'm running pulplift now with your nightly RPM packages to see if I get this error.

templates/pulpcore-api.service.erb
ekohl: > (pulp_installer defaults the pulp home dir to `/var/lib/pulp`, but lets users change it)

So does this module, but since it also manages the user, the homedir will be updated and that causes service restarts. Technically I think that Linux doesn't allow changing the homedir if a user is logged in, so you need to manually stop all services before, but after that it should work. Of course there's no data migration either.

ekohl:
Updated. There is now a parameter `libexecdir` and the `WorkingDirectory` in the services is set to `~`, meaning systemd will now also ensure the FS is mounted before starting services.

With this, I think it's ready to merge. I think those SELinux AVCs don't need to block merging.

Maybe we first want to get nightly green with 3.6 before adding more changes though. @jlsherrill?

jlsherrill:
Yeah that's probably a good idea, hopefully it'll all be fixed in a couple
days time

ekohl:
Rebased to resolve the merge conflict, no real change.

ekohl:
Looks like I had a bad rebase: it managed the assets dir twice. Should be fixed now.

ekohl:
@wbclark tests are green now and the pipeline is ready for it. Please have a look and approve/merge it if you think it's good.

ehelms:
I get errors testing this on current nightly. The issues look SELinux related:

```
[root@pipeline-katello-server-nightly-centos7 vagrant]# audit2allow -a


#============= pulpcore_t ==============
allow pulpcore_t httpd_config_t:dir search;
allow pulpcore_t httpd_sys_content_t:file getattr;
allow pulpcore_t httpd_sys_rw_content_t:dir search;
allow pulpcore_t proc_t:file read;
```

Or possibly this actually broken in nightly:

```
Sep 21 14:07:43 pipeline-katello-server-nightly-centos7.war.example.com pulpcore-worker-1[4730]: File "/usr/lib/python3.6/site-packages/pulpcore/app/settings.py", line 250, in <module>
Sep 21 14:07:43 pipeline-katello-server-nightly-centos7.war.example.com pulpcore-worker-1[4730]: "CONTENT_ORIGIN is a required setting but it was not configured. This may be caused "
Sep 21 14:07:43 pipeline-katello-server-nightly-centos7.war.example.com pulpcore-worker-1[4730]: django.core.exceptions.ImproperlyConfigured: CONTENT_ORIGIN is a required setting but it was not configured. This may be caused by invalid read permissions of the settings file. Note that CONTENT_ORIGIN is set by the installer automatically.
```

ekohl:
There was a big pulpcore-selinux update last week, but I'm not sure if that was already released. I'd like to test with that.

zpytela:
@ehelms adding these permissions should be fairly easy. I just need to ask: does pulpcore need the access to objects with the specified types? `httpd_config_t`  `httpd_sys_content_t` `httpd_sys_rw_content_t`
httpd_config_t can be /etc/pulp as well as /etc/httpd.
Could you share the complete AVC denials?

ekohl:
@zpytela I don't think your patch made it into nightly RPMs yet so I think that would be a good first step. Since it now labels `/usr/bin/{rq,unicorn}`, the libexec wrappers aren't even needed. I still don't think it's great to claim those files for Pulp only though.

ehelms:
Here ya go:

```
type=SYSCALL msg=audit(1600707522.684:93449): arch=c000003e syscall=4 success=no exit=-13 a0=7f202b7c1638 a1=7ffda81f5e50 a2=7ffda81f5e50 a3=24 items=0 ppid=1 pid=6277 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="rq" exe="/usr/bin/python3.6" subj=system_u:system_r:pulpcore_t:s0 key=(null)
type=AVC msg=audit(1600707522.684:93449): avc:  denied  { getattr } for  pid=6277 comm="rq" path="/etc/pulp/settings.py" dev="vda1" ino=2491257 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_sys_content_t:s0 tclass=file
```

```
type=SYSCALL msg=audit(1600707522.338:93434): arch=c000003e syscall=2 success=no exit=-13 a0=7f59a26eb3a6 a1=80000 a2=1b6 a3=24 items=0 ppid=6277 pid=6283 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="sh" exe="/usr/bin/bash" subj=system_u:system_r:pulpcore_t:s0 key=(null)
type=AVC msg=audit(1600707522.338:93434): avc:  denied  { read } for  pid=6283 comm="sh" name="meminfo" dev="proc" ino=4026532028 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:proc_t:s0 tclass=file
```

```
type=SYSCALL msg=audit(1600707521.865:93432): arch=c000003e syscall=4 success=no exit=-13 a0=7f203aa7d578 a1=7ffda81fbd70 a2=7ffda81fbd70 a3=24 items=0 ppid=1 pid=6277 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="rq" exe="/usr/bin/python3.6" subj=system_u:system_r:pulpcore_t:s0 key=(null)
type=AVC msg=audit(1600707521.865:93432): avc:  denied  { search } for  pid=6277 comm="rq" name="pulp" dev="vda1" ino=790580 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_sys_rw_content_t:s0 tclass=dir
```

manifests/init.pp
```diff
@@ -94,6 +98,7 @@
 class pulpcore (
   Stdlib::Absolutepath $cache_dir = '/var/lib/pulp/tmp',
   Stdlib::Absolutepath $config_dir = '/etc/pulp',
+  Stdlib::Absolutepath $libexecdir = '/usr/libexec/pulpcore',
```
mikedep333, line 101: If these are not part of the package, the FHS states that they should be under `/usr/local/libexec/pulpcore` instead.

zpytela:
> Here ya go:
> 
> ```
> type=SYSCALL msg=audit(1600707522.684:93449): arch=c000003e syscall=4 success=no exit=-13 a0=7f202b7c1638 a1=7ffda81f5e50 a2=7ffda81f5e50 a3=24 items=0 ppid=1 pid=6277 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="rq" exe="/usr/bin/python3.6" subj=system_u:system_r:pulpcore_t:s0 key=(null)
> type=AVC msg=audit(1600707522.684:93449): avc:  denied  { getattr } for  pid=6277 comm="rq" path="/etc/pulp/settings.py" dev="vda1" ino=2491257 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_sys_content_t:s0 tclass=file
> ```
So should we label settings.py with the new label pulpcore_etc_t as it was in the policy previously? If this file is for pulp3 only, the answer is probably yes. What about the keys, private_key.pem and public_key.pem?

> 
> ```
> type=SYSCALL msg=audit(1600707522.338:93434): arch=c000003e syscall=2 success=no exit=-13 a0=7f59a26eb3a6 a1=80000 a2=1b6 a3=24 items=0 ppid=6277 pid=6283 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="sh" exe="/usr/bin/bash" subj=system_u:system_r:pulpcore_t:s0 key=(null)
> type=AVC msg=audit(1600707522.338:93434): avc:  denied  { read } for  pid=6283 comm="sh" name="meminfo" dev="proc" ino=4026532028 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:proc_t:s0 tclass=file
> ```
This one is to add, likely with kernel_read_all_proc().

> 
> ```
> type=SYSCALL msg=audit(1600707521.865:93432): arch=c000003e syscall=4 success=no exit=-13 a0=7f203aa7d578 a1=7ffda81fbd70 a2=7ffda81fbd70 a3=24 items=0 ppid=1 pid=6277 auid=4294967295 uid=1001 gid=1001 euid=1001 suid=1001 fsuid=1001 egid=1001 sgid=1001 fsgid=1001 tty=(none) ses=4294967295 comm="rq" exe="/usr/bin/python3.6" subj=system_u:system_r:pulpcore_t:s0 key=(null)
> type=AVC msg=audit(1600707521.865:93432): avc:  denied  { search } for  pid=6277 comm="rq" name="pulp" dev="vda1" ino=790580 scontext=system_u:system_r:pulpcore_t:s0 tcontext=system_u:object_r:httpd_sys_rw_content_t:s0 tclass=dir
> ```
No information about the path; also search may not be the only permission needed. The permissive= entry was stripped off the record? Supposing 0 anyway; could you run again with `setenforce 0` and if possible also with full path auditing enabled?
And in general, should pulpcore have read access to http content, labeled httpd_sys_content_t and/or httpd_sys_rw_content_t?

ekohl:
I extracted the service changes to https://github.com/theforeman/puppet-pulpcore/pull/125. With the updated policy in 3.7 the libexec change isn't really needed and those service benefits should go in already.

ekohl:
I'm not entirely happy with the libexec changes. Marking it as a draft now.

manifests/init.pp
ekohl: You could make the same argument about `/var/lib/pulp`. I also would thing that managing the pulp user as part of the RPM and the service files would be good, but I think the ansible-based installer cares about pip installs so it probably won't happen.

manifests/service.pp
```diff
@@ -1,23 +1,37 @@
 # configure, enable, and start pulpcore services
 # @api private
 class pulpcore::service {
+  ['gunicorn', 'rq'].each |$bin| {
+    file { "${pulpcore::libexecdir}/${bin}":
```
evgeni, line 5: Let's package that instead: https://github.com/theforeman/pulpcore-packaging/pull/37

ekohl: https://github.com/theforeman/pulpcore-packaging/pull/41 is the cherry pick to 3.6.

ekohl:
Updated the PR to use the libexec wrappers provided by packaging.

ehelms (approved):

PR #117, opened 2020-08-06
Release 1.2.0

wbclark:


HISTORY.md
```diff
@@ -1,4 +1,42 @@
-# Changelog
+## [1.1.0](https://github.com/theforeman/puppet-pulpcore/tree/1.1.0) (2020-06-30)
```
ekohl, line 1: I only add things to the changelog if otherwise the tool can't generate a correct changelog. Was this the case here? (noticed the same thing in puppet-pulp).

HISTORY.md
```diff
@@ -1,4 +1,42 @@
-# Changelog
```
ekohl, line 1: Good catch

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+- Use pulpcore-manager [\#111](https://github.com/theforeman/puppet-pulpcore/pull/111) ([ekohl](https://github.com/ekohl))
+- Refs [\#30057](https://projects.theforeman.org/issues/30057) - Configure Pulpcore Worker Count [\#100](https://github.com/theforeman/puppet-pulpcore/pull/100) ([wbclark](https://github.com/wbclark))
+
+**Fixed bugs:**
+
+- Add the GPL license text [\#109](https://github.com/theforeman/puppet-pulpcore/pull/109) ([ekohl](https://github.com/ekohl))
+
+**Closed issues:**
+
+- Figure out how to make pulp-worker count configurable [\#37](https://github.com/theforeman/puppet-pulpcore/issues/37)
```
ekohl, line 19: I've added the skip-changelog label since there's already a PR that implements this, making this line redundant. In general issues don't really show up in our changelogs and I've considered just skipping them unconditionally.

HISTORY.md
wbclark: I've been burned by weird stuff showing up in changelogs enough times that I just made updating HISTORY.md a part of my process. I think it makes sense to keep it updated to the (n-1) release anyway since it just feels weird if we are on version 9.0.0 of a module but HISTORY.md only contains changes through 5.0.0

CHANGELOG.md
wbclark: Thanks, I reviewed the PRs but missed the closed issues. I'll update the changelogs to reflect this.

HISTORY.md
ekohl: I don't mind it, but I'm wondering if it makes sense to go all the way and copy the newest one as well (1.2.0 in this case). Thoughts?

PR #119, opened 2020-09-04
Refs #30780 - set proper content origin setting

jlsherrill:
CONTENT_ORIGIN needs to be poitned at https:// for the
/pulp/container/ redirect to work properly


ekohl:
I already ran into this in https://github.com/theforeman/puppet-pulpcore/pull/105. The thing is that right now this module doesn't know if the Apache vhost is HTTP or HTTPS. Its built in CI is HTTP only. Now we don't really test any content (we should) and the module is probably not used outside of our installer so we can probably merge this to unblock the pipeline, but I really need to find time to get back to the pulpcore fix. I wish the Apache code wasn't merged in puppet-foreman_proxy_content but that's the price you pay for going on vacation.

ehelms (approved):

PR #120, opened 2020-09-07
Handle X-Forwarded-Proto headers

ekohl:
In a reverse proxy setup, the process is running on HTTP only, but Apache is running on both HTTP and HTTPS. To let Django know, Apache sends the X-Forwarded-Proto header[1]. The Django settings documentation suggests this solution[2].

Apache sets these headers by default[3], but I don't know what happens if a clients sets them as well. Does Apache use those or explicitly sets its own?

[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto
[2]: https://docs.djangoproject.com/en/2.2/ref/settings/#secure-proxy-ssl-header
[3]: https://httpd.apache.org/docs/current/mod/mod_proxy.html#proxyaddheaders

### DIFF
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -37,3 +37,6 @@ REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
 
 <%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
 ALLOWED_IMPORT_PATHS = <%= scope['pulpcore::allowed_import_path'] %>
+
+# Derive HTTP/HTTPS via the X-Forwarded-Proto header set by Apache
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')


evgeni:
I *think* Apache doesn't allow to set these by clients, but I am not 100% sure

ekohl:
Open question: should we also set https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-USE_X_FORWARDED_HOST to true unconditionally?

ekohl:
@jlsherrill I'd appreciate a look at this.

@mikedep333 is this something that we should set in pulpcore's default settings?

It also looks like Django has no built in middleware for processing X-Forwarded-For to set `REMOTE_ADDR`.

evgeni:
FWIW, the Apache docs you linked only mention X-Forwarded-For, X-Forwarded-Host and X-Forwarded-Server, and not **-Proto**

ekohl:
Right. In pupet-foreman we set it ([here](https://github.com/theforeman/puppet-foreman/blob/e3bf011200d2519895627e29f0cb458dc47338b7/manifests/config/apache.pp#L195) and [here](https://github.com/theforeman/puppet-foreman/blob/e3bf011200d2519895627e29f0cb458dc47338b7/manifests/config/apache.pp#L218)).

In our case we can't really do that in this module because the Apache config lives all over the place. I think that in the common Katello setup it will actually be able to use the puppet-foreman headers. In https://github.com/theforeman/puppet-pulpcore/pull/105 I'd prefer to make that explicit. Otherwise we need to implement it in puppet-foreman_proxy_content.

jlsherrill:
i think this makes sense, but let me do a little testing to confirm nothing breaks related to fetching content over http or with cert guard.

jlsherrill (approved):
all my testing worked so this LGTM

jlsherrill:
i would think we should set USE_X_FORWARDED_HOST, or at least i can't think of any reason to not to?

ekohl:
No, I just see I didn't push the commit that did that even though I did it locally.

ekohl:
I pushed an additional commit. I think it's fine to combine it in the same PR.

ekohl:
Btw, there is also X-Forwarded-For but Django has no built in middleware for that. That's something upstream needs to implement first.

wbclark:
> Btw, there is also X-Forwarded-For but Django has no built in middleware for that. That's something upstream needs to implement first.

This is super useful for a Katello admin to have, particularly if any clients are behind a content proxy, as it can save a lot of time in troubleshooting and log analysis. Let's open a Pulp issue for it.

wbclark (approved):
Checking the Django 2.2 docs, both settings look correctly configured. I don't currently see any problem with having these statically configured. Considering this worked in Justin's testing, it looks good to me.

jlsherrill:
actually i just discovered that 'USE_X_FORWARDED_HOST = True' seems to break a podman pull:

```
pulp: django.security.DisallowedHost:ERROR: Invalid HTTP_HOST header: 'centos7-koji.windhelm.example.com, centos7-koji.windhelm.example.com'. The domain name provided is not valid according to RFC 1034/1035.
```
```

ekohl:
Interesting. That suggests it's proxied twice.

ekohl:
Rebased, for now I dropped `USE_X_FORWARDED_HOST`.

ekohl:
Based on https://github.com/theforeman/puppet-pulpcore/pull/120#pullrequestreview-484208754 I'm merging this now. Nightly pipelines are green so we'll see if this breaks.

PR #121, opened 2020-09-14
Manage the static assets via a class

ekohl:
Since 8c66bff5f2146139d2882ef75ac742235fe253b1 the pulpcore-manager command runs as root. On a fresh installation this is fine, but on upgrades the user doesn't own /var/lib/pulp/assets. This explicitly manages the static root.

It is split off to a separate class to better capture the dependencies.  The database doesn't need to refresh due to static files, nor any services.

### DIFF
--- a/manifests/static.pp
+++ b/manifests/static.pp
@@ -1,0 +1,16 @@
+# @summary Manage the static files (assets)
+# @api private
+class pulpcore::static {
+  file { $pulpcore::pulp_static_root:
+    ensure => directory,
+    owner  => $pulpcore::user,
+    group  => $pulpcore::group,
+    mode   => '0755',
+  }
+
+  pulpcore::admin { 'collectstatic --noinput':
+    refreshonly => true,
+    subscribe   => Concat['pulpcore settings'],
+    require     => File[$pulpcore::pulp_static_root],
+  }
+}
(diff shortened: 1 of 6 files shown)


ekohl:
> The database doesn't need to refresh due to static files, nor any services.

I didn't verify this - it's true for "standard" Django, but with [whitenoise](https://pypi.org/project/whitenoise/) this may not be true.

ekohl:
Note I extracted this from https://github.com/theforeman/puppet-pulpcore/pull/115, which is why I assumed it was already done.

manifests/static.pp
```diff
@@ -0,0 +1,16 @@
+# @summary Manage the static files (assets)
+# @api private
+class pulpcore::static {
+  file { $pulpcore::static_root:
```
wbclark, line 4: I think this parameter `$pulpcore::static_root` was present in your other PR but didn't get included here, which is breaking things

Otherwise it looks good to me

ekohl: Ah yes, here it was still `pulp_static_root`.

ekohl:
I added a commit that sets `STATIC_ROOT` in settings.py. This was also taken from the new layout PR and I'd consider it a bugfix.

wbclark (approved):
LGTM

PR #123, opened 2020-09-30
Add Pulpcore repository class

ekohl:
This class uses the new yum.theforeman.org repos that are versions per Pulpcore release. This allows easy switching between versions. It also properly sets the GPG key validation.

It does switch acceptance testing from staging repositories to stable repositories. Another implication is that it switches to Pulpcore 3.6 by default. This was chosen since 3.7 is not complete yet.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -1,0 +1,24 @@
+# Configure the Pulpcore repo
+#
+# @param version
+#   The Pulpcore version to use
+class pulpcore::repo (
+  Pattern['^\d+\.\d+$'] $version = '3.6',
+) {
+  $context = {
+    'version'   => $version,
+    'dist_tag' => "el${facts['os']['release']['major']}",
+  }
+
+  file { '/etc/yum.repos.d/pulpcore.repo':
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
+    content => epp('pulpcore/repo.epp', $context),
+    notify  => Anchor['pulpcore::repo'],
+  }
+
+  # An anchor is used because it can be collected
+  anchor { 'pulpcore::repo': } # lint:ignore:anchor_resource
+}
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -23,26 +23,14 @@ case $major {
   default: {}
 }
 
-# Defaults to staging, for release, use
-# $baseurl = "https://fedorapeople.org/groups/katello/releases/yum/nightly/pulpcore/el${major}/x86_64/"
-$baseurl = "http://koji.katello.org/releases/yum/katello-nightly/pulpcore/el${major}/x86_64/"
-
-$pulpcore_repo_conf = @("CONF")
-[pulpcore]
-baseurl=${baseurl}
-gpgcheck=0
-module_hotfixes=1
-CONF
-
-file { '/etc/yum.repos.d/pulpcore.repo':
-  ensure  => file,
-  content => $pulpcore_repo_conf,
+class { 'pulpcore::repo':
+  version => fact('pulpcore_version'),
 }
 
 # Needed as a workaround for idempotency
 if $facts['os']['selinux']['enabled'] {
   package { 'pulpcore-selinux':
     ensure  => installed,
-    require => File['/etc/yum.repos.d/pulpcore.repo'],
+    require => Class['pulpcore::repo'],
   }
 }
--- a/templates/repo.epp
+++ b/templates/repo.epp
@@ -1,0 +1,16 @@
+<%- | String[1] $version, String[1] $dist_tag | -%>
+[pulpcore]
+name=Pulpcore <%= $version %>
+baseurl=https://yum.theforeman.org/pulpcore/<%= $version %>/<%= $dist_tag %>/$basearch
+enabled=1
+gpgcheck=1
+gpgkey=https://yum.theforeman.org/pulpcore/<%= $version %>/GPG-RPM-KEY-pulpcore
+module_hotfixes=1
+
+[pulpcore-source]
+name=Pulpcore <%= $version %> Source
+baseurl=https://yum.theforeman.org/pulpcore/<%= $version %>/<%= $dist_tag %>/source
+enabled=0
+gpgcheck=1
+gpgkey=https://yum.theforeman.org/pulpcore/<%= $version %>/GPG-RPM-KEY-pulpcore
+module_hotfixes=1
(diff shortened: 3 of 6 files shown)


manifests/repo.pp
```diff
@@ -0,0 +1,40 @@
...
+    module_hotfixes=1
+
+    [pulpcore-source]
+    name=Pulpcore ${version} Soure
+    baseurl=${base}/source
+    enabled=0
+    gpgcheck=1
+    gpgkey=${base}/GPG-RPM-KEY-pulpcore
+    module_hotfixes=1
+    | REPO
```
ehelms, line 27: Can you share why you chose to inline this here instead of using a file or template?  
Is this puppet DSL version of HEREDOC?

ekohl: I actually copied it from the setup_acceptance_node file without thinking too much. It is indeed the Puppet DSL version of heredoc but a separate template probably makes more sense.

ekohl:
@evgeni this probably interests you. I do wonder if we should have some first stage where we test a reduced matrix and then a second stage where we test the rest. For example, 3.7 on Puppet 6 (probably both EL7 and EL8) and then all other variations on it.

evgeni:
Do you mean interests me to test/stage pulpcore, or puppet-pulpcore?

ekohl:
Also note that currently it doesn't appear to work yet and it doesn't find the custom fact, hence the draft status.

ekohl:
> Do you mean interests me to test/stage pulpcore, or puppet-pulpcore?

I mostly hope to get some help on the Github actions part but also verification of the RPM packages in the repo.

.github/workflows/acceptance.yml
```diff
@@ -0,0 +1,51 @@
...
+        setfile:
+          - centos7-64{hostname=centos7-64.example.com}
+          - centos8-64{hostname=centos8-64.example.com}
+        puppet:
+          - "6"
+          - "5"
+        pulpcore:
+          - '3.4'
+          - '3.6'
+          - '3.7'
```
ehelms, line 22: This implies that the development branch is attempting to maintain support for 3 versions of Pulp at a time. Is that are true desire? That seems more than standard.

ekohl: It is what we now state in the README. I'm not sure where we should land, but I think this is a pattern I want to apply in more places, such as puppet-foreman.

ekohl:
Note to self: testing with 3.6 shows
```
System check identified some issues:
  
  WARNINGS:
  ?: (guardian.W001) Guardian authentication backend is not hooked. You can add this in settings as eg: `AUTHENTICATION_BACKENDS = ('django.contrib.auth.backends.ModelBackend', 'guardian.backends.ObjectPermissionBackend')`.
```

ekohl:
Ok, the conclusion is that the 3.7 repo is not installable yet but the mechanism works with 3.4 and 3.6. Caching of gems works well, about 10 seconds on a cache hit is nice.

.github/workflows/acceptance.yml
ehelms: The thing I want to make sure we think about is how much choosing to support some amount of previous versions could pigeon hole us into additional work down the road trying to keep the main branch compatible across versions (vs. utilizing module releases and stable branches to achieve that).

ekohl:
Updated. I dropped the multi version testing for later and also pointed the repo to 3.6 instead of 3.7 (which isn't done yet).

ekohl:
Tests should now also pass.

spec/setup_acceptance_node.pp
```diff
@@ -23,26 +23,14 @@
...
-baseurl=${baseurl}
-gpgcheck=0
-module_hotfixes=1
-CONF
-
-file { '/etc/yum.repos.d/pulpcore.repo':
-  ensure  => file,
-  content => $pulpcore_repo_conf,
+class { 'pulpcore::repo':
+  version => fact('pulpcore_version'),
```
ekohl, line 27: Once https://github.com/voxpupuli/voxpupuli-acceptance/pull/13 is merged and released in the gem, you can use `BEAKER_FACTER_PULPCORE_VERSION=3.7` when running beaker to test out version 3.7. If it's undefined, it falls back to the class default (3.6).

templates/repo.epp
```diff
@@ -0,0 +1,16 @@
...
+[pulpcore]
+name=Pulpcore <%= $version %>
+baseurl=https://yum.theforeman.org/pulpcore/<%= $version %>/<%= $dist_code %>/$basearch
+enabled=1
+gpgcheck=1
+gpgkey=https://yum.theforeman.org/pulpcore/<%= $version %>/GPG-RPM-KEY-pulpcore
+module_hotfixes=1
+
+[pulpcore-source]
+name=Pulpcore ${version} Soure
```
ehelms, line 11:
```suggestion
name=Pulpcore <%= $version %> Source

File: manifests/repo.pp [hunk 5]
```diff
@@ -0,0 +1,24 @@
+# Configure the Pulpcore repo
+#
+# @param version
+#   The Pulpcore version to use
+class pulpcore::repo (
+  Pattern['^\d+\.\d+$'] $version = '3.6',
+) {
+  $context = {
+    'version'   => $version,
+    'dist_code' => "el${facts['os']['release']['major']}",
```
ehelms, line 10: The term `dist_code` seemed weird, but I get the thought behind naming it such. I most often see it called a `distirbution tag` or `dist tag` in related docs, so for congruence naming this `dist_tag` would align.

templates/repo.epp
```diff
@@ -0,0 +1,16 @@
...
+[pulpcore]
+name=Pulpcore <%= $version %>
+baseurl=https://yum.theforeman.org/pulpcore/<%= $version %>/<%= $dist_code %>/$basearch
+enabled=1
+gpgcheck=1
+gpgkey=https://yum.theforeman.org/pulpcore/<%= $version %>/GPG-RPM-KEY-pulpcore
+module_hotfixes=1
+
+[pulpcore-source]
+name=Pulpcore <%= $version %> Soure
```
ehelms, line 11:
```suggestion
name=Pulpcore <%= $version %> Source
```

ehelms (changes requested):

manifests/repo.pp
ekohl: With CentOS you can use `$releasever` in yum configs and that's `7` or `8` but with RHEL it can be `7Server`. That would have been my preferred solution if RHEL didn't complicate it. For now I'll go with `dist_tag` then.

ehelms (approved):

PR #124, opened 2020-10-02
Switch to using a Unix socket bind for API and Content services

ehelms:
The use of a Unix socket between the deployed service and the reverse
proxy provides tighter security as the only users who can access the
socket are root and the configured SocketUser. The introduction of
a systemd socket with a ListenStream also provides automatic activation
of the underlying service and safer restarts.

This change is backwards incompatible as it removes the host and port parameters
for the API and Content services in favor of a single bind parameter for each.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -10,10 +10,10 @@ class pulpcore::apache (
 ) {
   $vhost_priority = $pulpcore::apache_vhost_priority
   $api_path = '/pulp/api/v3'
-  $api_base_url = "http://${pulpcore::api_host}:${pulpcore::api_port}"
+  $api_base_url = "unix://${pulpcore::api_socket_path}|http://${pulpcore::servername}"
   $api_url = "${api_base_url}${api_path}"
   $content_path = '/pulp/content'
-  $content_base_url = "http://${pulpcore::content_host}:${pulpcore::content_port}"
+  $content_base_url = "unix://${pulpcore::content_socket_path}|http://${pulpcore::servername}"
   $content_url = "${content_base_url}${content_path}"
 
   $docroot_directory = {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -42,17 +42,11 @@
 #   apache_https_vhost, this will be used when attaching fragments to those
 #   vhosts. Note that this implies both vhosts need to have the same priority.
 #
-# @param api_host
-#   API service host
+# @param api_socket_path
+#   Path where the Pulpcore API service is listening. This is a unix socket.
 #
-# @param api_port
-#   API service port
-#
-# @param content_host
-#   Content service host
-#
-# @param content_port
-#   Content service port
+# @param content_socket_path
+#   Path where the Pulpcore Content service is listening. This is a unix socket.
 #
 # @param config_dir
 #   Pulp configuration directory. The settings.py file is created under this
@@ -158,10 +152,8 @@ class pulpcore (
   Optional[Stdlib::Absolutepath] $apache_https_ca = undef,
   Optional[Stdlib::Absolutepath] $apache_https_chain = undef,
   String[1] $apache_vhost_priority = '10',
-  Stdlib::Host $api_host = '127.0.0.1',
-  Stdlib::Port $api_port = 24817,
-  Stdlib::Host $content_host = '127.0.0.1',
-  Stdlib::Port $content_port = 24816,
+  Stdlib::Absolutepath $api_socket_path = '/run/pulpcore-api.sock',
+  Stdlib::Absolutepath $content_socket_path = '/run/pulpcore-content.sock',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_user = 'pulp',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -1,8 +1,22 @@
 # configure, enable, and start pulpcore services
 # @api private
 class pulpcore::service {
+  include apache
+
+  systemd::unit_file { 'pulpcore-api.socket':
+    content => template('pulpcore/pulpcore-api.socket.erb'),
+    active  => true,
+    enable  => true,
+  }
+
   systemd::unit_file { 'pulpcore-api.service':
     content => template('pulpcore/pulpcore-api.service.erb'),
+    active  => true,
+    enable  => true,
+  }
+
+  systemd::unit_file { 'pulpcore-content.socket':
+    content => template('pulpcore/pulpcore-content.socket.erb'),
     active  => true,
     enable  => true,
   }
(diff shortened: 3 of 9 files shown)


manifests/init.pp
```diff
@@ -98,10 +92,8 @@
   String $group = 'pulp',
   Stdlib::Absolutepath $user_home = '/var/lib/pulp',
   Boolean $manage_apache = true,
-  Stdlib::Host $api_host = '127.0.0.1',
-  Stdlib::Port $api_port = 24817,
-  Stdlib::Host $content_host = '127.0.0.1',
-  Stdlib::Port $content_port = 24816,
+  Pattern['^unix://'] $api_service_bind = 'unix:///run/pulpcore-api.sock',
```
ehelms, line 95: The Pulp project itself is using the path `/var/run/pulpcore-api` (https://github.com/pulp/pulp_installer/blob/d6fd0f468f7773d01d7eb146044daec5b9c74a18/molecule/source-static/group_vars/all#L4). Given this socket is shared between the pulpcore systemd socket and Apache, and is owned by the Apache user I leaned towards a top level location. However, I am open to congruency with the Pulp project.

ekohl: I personally stopped using `/var/run`. It's deprecated in favor of `/run` and just a symlink nowadays. However, in practice it means there is no real difference.

manifests/apache.pp
```diff
@@ -2,9 +2,9 @@
 # @api private
 class pulpcore::apache {
   $api_path = '/pulp/api/v3'
-  $api_url = "http://${pulpcore::api_host}:${pulpcore::api_port}${api_path}"
+  $api_url = "${pulpcore::api_service_bind}|http://${pulpcore::servername}${api_path}"
   $content_path = '/pulp/content'
-  $content_url = "http://${pulpcore::content_host}:${pulpcore::content_port}${content_path}"
+  $content_url = "${pulpcore::content_service_bind}|http://${pulpcore::servername}${content_path}"
```
ekohl, line 7: This will break things in puppet-foreman_proxy_content:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/743517dac849b5a0cd9bec56077f50f227a1f8c5/templates/pulpcore-docker-apache.conf.erb#L6-L10

In https://github.com/theforeman/puppet-pulpcore/pull/105 I hope to address that, but I haven't been able to sit down and even check out the branch to update it in the past weeks.

manifests/init.pp
ehelms: Sorry, I should have been more clear what I was asking here. The variation comes in the directory under which the socket is stored, Pulp uses the full path `/run/pulpcore-api/pulpcore-api.sock` (note the scoping on the `pulpcore-api` directory).

manifests/apache.pp
ehelms: Ahh yes indeed, you solved a testing mystery for me (there was one test that failed and that was pulling container content)

manifests/init.pp
ekohl: Ah, that difference. Given that it's owned by Apache and Apache may not have permission to read in `/run/pulpcore-api`, I'm also leaning to a top level location. Inside the Apache run dir may also not be correct since I'm not sure you can rely on that existing when the socket is started.

ekohl:
I would like to see a version of this based on https://github.com/theforeman/puppet-pulpcore/pull/105.

ekohl:
I rebased my own patch from July (661151f4d12ce0487c41aecb214baac2958596a5). Looks like I never pushed that work. I didn't test it extensively, but it does a few more tricks that you didn't. For example, it implements reload and uses type=notify. I also dropped the `--bind` since it's not needed if you pass in systemd sockets. I didn't implement the apache user correctly though.

ehelms:
I assume by that gunicorn does support systemd notify out of the box? I debated whether to keep the bind or not. The main reason I did is because I think it's easier to reason about what is going on by explicitly setting it for users inspecting the system.

ehelms:
This is  now based on https://github.com/theforeman/puppet-pulpcore/pull/105 -- so I suppose this is a bit of a draft and will now require that work to be completed.

spec/classes/pulpcore_spec.rb
```diff
@@ -259,6 +260,8 @@
         end
       end
 
+=======
+>>>>>>> 1e1cdfd... Switch to using a Unix socket bind for API and Content services
```
ekohl, line 264: Merge conflict here

manifests/service.pp
```diff
@@ -1,12 +1,25 @@
 # configure, enable, and start pulpcore services
 # @api private
 class pulpcore::service {
+
```
ekohl, line 4: A future lint (that's on the agenda to add) will complain about this empty line.

ekohl (commented):
> I assume by that gunicorn does support systemd notify out of the box?

Yes it does. The unit files were largely copied from https://docs.gunicorn.org/en/stable/deploy.html#systemd



> I debated whether to keep the bind or not. The main reason I did is because I think it's easier to reason about what is going on by explicitly setting it for users inspecting the system.

I think it may be confusing since gunicorn ignores all binds when there are systemd sockets:

https://github.com/benoitc/gunicorn/blob/548d5828da6b93fa6a14217742c6e6d2c7b2b900/gunicorn/sock.py#L142-L204

This may cause someone to try and use binds but being confused. I would leave it out.

If a user wants to inspect the system, they should see that it requires the socket and use `systemctl cat theservice.socket` to get the exact binds.

Also note that the documentation example also leaves it out.

manifests/service.pp
ehelms: Bummer, I like the readability the empty line creates.

ekohl:
Please rebase.

templates/pulpcore-api.socket.erb
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Pulp API Server socket
+
+[Socket]
+ListenStream=<%= scope['pulpcore::api_socket_path'] %>
+SocketUser=<%= scope['apache::user'] %>
```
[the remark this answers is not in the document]
ehelms: I had similar issues the first time round with the puppet-foreman implementation, will try that out

manifests/service.pp
```diff
@@ -1,12 +1,24 @@
 # configure, enable, and start pulpcore services
 # @api private
 class pulpcore::service {
+  systemd::unit_file { 'pulpcore-api.socket':
+    content => template('pulpcore/pulpcore-api.socket.erb'),
+    active  => true,
+    enable  => true,
+  }
+
   systemd::unit_file { 'pulpcore-api.service':
```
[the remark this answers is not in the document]
ehelms: I think this output implies this already is created:

```
        dependency cycles found: (Exec[systemctl-daemon-reload] => Class[Systemd::Systemctl::Daemon_reload] => Service[pulpcore-api.socket] => Systemd::Unit_file[pulpcore-api.socket] => Systemd::Unit_file[pulpcore-api.service] => File[/etc/systemd/system/pulpcore-api.service] => Class[Systemd::Systemctl::Daemon_reload] => Exec[systemctl-daemon-reload])
```

ehelms:
All :green_circle:  now

ekohl (approved):
Thanks! I do think this deserves a description in README so I'll do a follow up PR.

PR #125, opened 2020-10-05
Drop pid files and fix working dirs

ekohl:
When using systemd, the pid files are actually not important so they can be dropped.

The the WorkingDirectory is set to ~ which will resolve to the user's home directory. That means systemd will ensure it is mounted before starting services. Pulp itself doesn't use it so it's a cheap way for a more reliable service start.

Subscribing to pulpcore-worker@.service is another bugfix to ensure all workers use the correct service definition.

Explicitly setting the type and group is more for consistency and doesn't actually change anything.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -1,7 +1,6 @@
 # configure, enable, and start pulpcore services
 # @api private
 class pulpcore::service {
-
   systemd::unit_file { 'pulpcore-api.service':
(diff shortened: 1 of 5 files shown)


wbclark (approved):
Per the systemd.service manpage the PIDFiles are indeed unnecessary and only recommended when `type=Forking` which does not apply here.

Nice find with the change from `require` to `subscribe` for the resource `Systemd::Unit_file['pulpcore-worker@.service']`

PR #127, opened 2020-10-09
Support Pulpcore 3.7

ekohl:
This includes the switch to Github actions, but that needs some work. For example, currently in Travis we have a weekly cronjob to notify us of regressions, but it's still missing in the Github actions part.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -36,7 +36,10 @@ jobs:
         puppet:
           - "6"
           - "5"
-    name: Puppet ${{ matrix.puppet }} - ${{ matrix.setfile }}
+        pulpcore_version:
+          - '3.6'
+          - '3.7'
+    name: Puppet ${{ matrix.puppet }} - Pulp ${{ matrix.pulpcore_version }} - ${{ matrix.setfile }}
     steps:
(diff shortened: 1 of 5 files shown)


evgeni:
@ekohl see https://github.com/Apipie/apypie/pull/72/files#diff-4a2062e4c1189e390d76677db2b0f372R11 for an cron example on GH actions ;)

ekohl:
@evgeni The only thing I would like to express is "run crons only on theforeman/puppet-pulpcore". One way is to copy the file and add that if, but it does result in duplication. However, foreman-installer-modulesync can solve that for us. I just need a good way to deal with the multiple supported versions in a template.

ekohl:
Rebased, now includes https://github.com/theforeman/puppet-pulpcore/pull/136, based on https://github.com/theforeman/foreman-installer-modulesync/pull/106.

ehelms:
There are some errant looking GH action tests on the PR, seeing if closing and re-opening fixes that.

ehelms:
There were out of date requirements on merging that I have cleaned up.

ekohl:
Those requirements were easier with Travis where it just needed to pass. Now that names can change, it becomes harder.

ehelms (approved):

PR #128, opened 2020-10-09
Expand README with more about the support policy.

ekohl:


README.md
```diff
@@ -2,10 +2,16 @@
...
+It assumes package repositories are present on the system. For convenience there is a `pulpcore::repo` class which installs from [yum.theforeman.org](https://yum.theforeman.org/pulpcore) (built from [pulpcore-packaging](https://github.com/theforeman/pulpcore-packaging)) but users can obviously use other sources. Installation from pip is not supported.
+
 ## Support policy
 
-The module provides no guarantee for multiple versions. Whenever a version is dropped, the major version is increased. All supported versions are listed.
+The module provides no guarantee for multiple versions. Whenever a version is dropped, the major version is increased.
+
+All supported versions are listed. For every supported version, acceptance tests run in CI on every supported platform.
+
+Supported operating systems are listed in `metadata.json` but individual releases can divert from that. For example, if Pulpcore x.y drops EL7, it will still be listed in metadata.json until all versions have dropped it. Similar, if x.z adds support for EL9, it'll be listed in `metadata.json` and all versions that don't support EL9 will see a note.
```
wbclark, line 13:
```suggestion
Supported operating systems are listed in `metadata.json` but individual releases can divert from that. For example, if Pulpcore x.y drops EL7, it will still be listed in metadata.json until all versions have dropped it. Similarly, if x.z adds support for EL9, it will be listed in `metadata.json` and all versions that don't support EL9 will see a note.
```

I have the most minor of nits.

wbclark (approved):
See nits above but overall the extra clarity is good.

ekohl:
I'd like @ehelms to reply as well before I make adjustments.

README.md
```diff
@@ -2,10 +2,16 @@
 
 Puppet module to set up Pulp 3. The primary goal of the maintainers is to set up Pulp 3 as part of Katello installation, but there's no reason it couldn't be used elsewhere.
 
+It assumes package repositories are present on the system. For convenience there is a `pulpcore::repo` class which installs from [yum.theforeman.org](https://yum.theforeman.org/pulpcore) (built from [pulpcore-packaging](https://github.com/theforeman/pulpcore-packaging)) but users can obviously use other sources. Installation from pip is not supported.
```
ehelms, line 5: I would drop `obviously` and just state users can use other sources. The `it` at the start of the paragraph is ambiguous since there is no prior subject. I would opt for `The module assumes..`

README.md
```diff
@@ -2,10 +2,16 @@
...
+
 ## Support policy
 
-The module provides no guarantee for multiple versions. Whenever a version is dropped, the major version is increased. All supported versions are listed.
+The module provides no guarantee for multiple versions. Whenever a version is dropped, the major version is increased.
```
ehelms, line 9: I think it could benefit to be more specific here in each mentions of versions which versions we are talking about: module version or Pulpcore version.

ekohl:
Updated and I think I addressed all comments.

ehelms (approved):

PR #129, opened 2020-10-09
use absolute paths in unit files

wbclark:
To prevent:

`Oct 09 15:31:27 centos7-64-1 systemd[1]: [/etc/systemd/system/pulpcore-api.service:11] Not an absolute path, ignoring: ~`


ekohl (approved):
https://www.freedesktop.org/software/systemd/man/systemd.exec.html#WorkingDirectory= states
> If set to "~", the home directory of the user specified in User= is used.

I guess EL7's systemd is too old for that.

ekohl:
In hindsight, this should have used `Refs #30465` for proper tracking.

PR #137, opened 2020-10-27
Add ability to control if services are enabled or active

ehelms:


manifests/init.pp
```diff
@@ -130,6 +130,12 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param service_enable
+#   Enable/disable Pulp services at boot.
+
```
alexjfisher, line 135: This stray blank line would cause puppet-strings to only 'see' the lines below.

alexjfisher (changes requested):

manifests/init.pp
ehelms: Doh - thanks

alexjfisher (approved):

manifests/init.pp
```diff
@@ -171,6 +177,8 @@
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
+  Boolean $service_enable = true,
+  Boolean $service_ensure = true,
```
ekohl, line 181: We can also use running/stopped.
```suggestion
  Stdlib::Ensure::Service $service_ensure = 'running',
```

ehelms: Maybe I am missing a nuance, but it appears to validate that its a boolean for `systemd::unit_file`:

https://github.com/camptocamp/puppet-systemd/blob/master/manifests/unit_file.pp#L60

ekohl: While it would both work, by using the type alias that uses `running` and `stopped` you would avoid the casting in service.pp (around line 40). There also has been a long standing convention that `running` and `stopped` should be used. `ensure => running` reads a lot better than `ensure => true` since it's self-documenting.

ehelms: Sorry, I am confused still as to what the design change ask here is. Are you saying it's better to go with `running` and `stopped` and cast that to a Boolean inside the manifest for the `systemd::unit_file` given it requires a boolean?

ekohl: Ah, I missed that nuance and assumed it accepted running/stopped.

@alexjfisher what do you think is a cleaner API?

manifests/service.pp
```diff
@@ -5,42 +5,48 @@
...
 
   systemd::unit_file { 'pulpcore-worker@.service':
     content => template('pulpcore/pulpcore-worker@.service.erb'),
   }
 
+  if $pulpcore::service_ensure == false {
+    $worker_ensure = 'stopped'
+  } else {
+    $worker_ensure = 'running'
+  }
```
ekohl, line 44: I'm never a fan of checking for negative first so I'd normally write:
```suggestion
  if $pulpcore::service_ensure {
    $worker_ensure = 'running'
  } else {
    $worker_ensure = 'stopped'
  }
```

You can also simplify it:
```suggestion
  $worker_ensure = bool2str($pulpcore::service_ensure, 'running', 'stopped')
```

manifests/init.pp
ehelms: Seems this is waiting on your input @alexjfisher and I'd like to close this out to do the follow on work.

manifests/service.pp
ekohl: Oh, even easier, you can actually pass a boolean to `ensure` so no need to convert that. If you do that, then I'm fine with accepting a boolean in the class.

manifests/init.pp
```diff
@@ -130,6 +130,12 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param service_enable
+#   Enable/disable Pulp services at boot.
+#
+# @param service_ensure
+#   Specify if Pulp services should be running.
```
ekohl, line 137: This makes it clear that the service will actually be stopped (and isn't unmanaged on `false`)
```suggestion
#   Specify if Pulp services should be running or stopped.
```

manifests/service.pp
```diff
@@ -5,42 +5,44 @@
...
     content => template('pulpcore/pulpcore-worker@.service.erb'),
   }
 
+  $worker_ensure = bool2str($pulpcore::service_ensure, 'running', 'stopped')
+
   Integer[1, $pulpcore::worker_count].each |$n| {
     service { "pulpcore-worker@${n}.service":
-      ensure    => running,
-      enable    => true,
+      ensure    => $worker_ensure,
```
ekohl, line 44: My last suggestion was this:
```suggestion
      ensure    => $pulpcore::service_ensure,
```
(and drop `$worker_ensure`)

ekohl (changes requested):

ekohl (approved):

ekohl:
(looks like the tests need to be updated)

PR #138, opened 2020-10-27
Set docroot to pulpcore_static

ekohl:
6a88107e66607dbbbd008d5b2139ed538395f177 changed the docroot to /var/lib/pulp/docroot. While I still think this is a much more descriptive name what it is, Pulp upstream has chosen to go with pulpcore_static. A suggestion to change it was rejected in [7750].

This path will also be included in the SELinux policy so it's important it matches here.

[7750]: https://pulp.plan.io/issues/7750#note-3

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ These is also the `cache_dir` which is used to configure [WORKING_DIRECTORY](htt
 
 There is also `chunked_upload_dir` to configure the undocumented `CHUNKED_UPLOAD_DIR`. This directory stores the temporary files used for files uploaded as chunks.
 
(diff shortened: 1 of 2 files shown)


ehelms (approved):
Is there an installer change needed with this?

ekohl:
No, we haven't released this. We could clean up the old directory we had in 3.17 and earlier but that doesn't block this. Having this in before branching will at least have fresh 3.18 installs without any cruft and every 3.18+ installation will have the correct directory.

ehelms:
Tests need updating

wbclark:
Closed by #141

PR #140, opened 2020-10-29
Gh actions

wbclark:
I wasn't able to push the first commit directly to master so CI won't run for this PR until that is done and this gets rebased


wbclark:
#142 would unblock this

ehelms (approved):

PR #144, opened 2020-11-04
Set pulpcore_version in .sync.yml for modulesync

ekohl:
This uses the [new support in modulesync](https://github.com/theforeman/foreman-installer-modulesync/pull/133) for this. It makes sure a new modulesync wouldn't overwrite the config.

Currently a draft since I haven't figured out the name yet, but I want to see the result of how it is now.


ekohl:
Will come with an updated modulesync.

PR #145, opened 2020-11-05
support /pulp/deb proxying

jlsherrill:


manifests/plugin/deb.pp
```diff
@@ -1,5 +1,33 @@
 # @summary Pulp Deb plugin
+# @param use_pulp2_content_route
 class pulpcore::plugin::deb {
```
ekohl, line 3:
```suggestion
class pulpcore::plugin::deb (
```

manifests/plugin/deb.pp
```diff
@@ -1,5 +1,33 @@
 # @summary Pulp Deb plugin
+# @param use_pulp2_content_route
```
ekohl, line 2: Could you add a short description? It'll end up being rendered something like https://forge.puppet.com/theforeman/pulpcore/reference#pulpcorepluginrpm

ekohl (changes requested):
Does this also work with Pulpcore 3.6 or is it 3.7 only? I'm trying to keep https://github.com/theforeman/puppet-pulpcore#support-policy correct.

manifests/plugin/deb.pp
```diff
@@ -1,5 +1,33 @@
 # @summary Pulp Deb plugin
+# @param use_pulp2_content_route Whether to redirect the legacy (Pulp 2) URL, /pulp/deb/, to the content server
 class pulpcore::plugin::deb {
+  Boolean $use_pulp2_content_route = false,
+  } {
```
ekohl, line 5:
```suggestion
) {
```

ekohl (changes requested):
You're using the wrong braces for parameters, causing it to fail.

manifests/plugin/deb.pp
jlsherrill: oh shoot, sorry i got it backwards :)

manifests/plugin/deb.pp
```diff
@@ -1,5 +1,33 @@
 # @summary Pulp Deb plugin
-class pulpcore::plugin::deb {
+# @param use_pulp2_content_route Whether to redirect the legacy (Pulp 2) URL, /pulp/deb/, to the content server
+class pulpcore::plugin::deb (
+  Boolean $use_pulp2_content_route = false,
+  ) {
```
ekohl, line 5: Lint is unhappy
```suggestion
) {
```

jlsherrill:
```Does this also work with Pulpcore 3.6 or is it 3.7 only?```

It should work with both, nothing about it is specific to any pulpcore version

ekohl (approved):

PR #147, opened 2020-11-11
Fixes #30436 - add allowed_export_paths to settings.py

jeremylenz:
As part of the new content view import/export API work, we need to be able to add import & export paths to `settings.py`:

```python
ALLOWED_IMPORT_PATHS = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports'] # already exists
ALLOWED_EXPORT_PATHS = ['/var/lib/pulp/exports'] # new
```

This PR will be modeled on #79 and adds `ALLOWED_EXPORT_PATHS`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -123,10 +123,10 @@
 #   Django remote user environment variable
 #
 # @param allowed_import_path
-#   Allowed paths that pulp can use for content view imports, or sync from using file:// protocol
+#   Allowed paths that pulp can use for content imports, or sync from using file:// protocol
 #
 # @param allowed_export_path
-#   Allowed paths that pulp can use for content view exports
+#   Allowed paths that pulp can use for content exports
 #
 # @param worker_count
 #   Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller. Enabling more than 8 workers, even with additional CPU cores
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -177,7 +177,7 @@ class pulpcore (
   String $django_secret_key = extlib::cache_data('pulpcore_cache_data', 'secret_key', extlib::random_password(32)),
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
-  Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports', '/var/lib/pulp/imports'],
+  Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Array[Stdlib::Absolutepath] $allowed_export_path = ['/var/lib/pulp/exports'],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -178,7 +178,7 @@ class pulpcore (
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
-  Array[Stdlib::Absolutepath] $allowed_export_path = ['/var/lib/pulp/exports'],
+  Array[Stdlib::Absolutepath] $allowed_export_path = [],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
   Boolean $service_enable = true,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -123,7 +123,10 @@
 #   Django remote user environment variable
 #
 # @param allowed_import_path
-#   Allowed paths that pulp can sync from using file:// protocol
+#   Allowed paths that pulp can use for content view imports, or sync from using file:// protocol
+#
+# @param allowed_export_path
+#   Allowed paths that pulp can use for content view exports
 #
 # @param worker_count
 #   Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller. Enabling more than 8 workers, even with additional CPU cores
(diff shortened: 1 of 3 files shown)


jeremylenz:
still working on tests, so set to Draft for now

wbclark (commented):
Hey @jeremylenz thanks for this PR!

I have a few questions to better understand:

1. Do you consider the change to default `$allowed_import_path` to fall under the same Redmine? Should it be split into a separate commit, or separate PR and Redmine entirely?

2. Should this configuration be the same for standalone pulp3, pulp3 on katello, and pulp3 on content proxy? This module is used for both the katello and foreman_proxy_content installer scenarios and unless the configuration should be the same in both cases, we should expose the new parameter in puppet-foreman_proxy_content so that the installer can pass different values of the parameter per scenario.

jeremylenz:
@wbclark good questions!

For item 1 the change is related and small enough that I think it can stay here in this PR and redmine.
Regarding item 2 -  It's definitely relevant for pulp3 on katello.  I don't think this is relevant for standalone pulp3, but I'm not sure about pulp3 on content proxy.   @parthaa can probably give a better answer.

wbclark (commented):
OK, thinking about it a bit, content proxies shouldn't need CV import/export since they sync directly from Katello

So I think we should make the following changes:

1. This PR should include only the new parameter for $allowed_export_path

2. A second PR should be opened in puppet-FPC which exposes $allowed_{import,export}_path parameters at https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L371-L394

3. A 3rd PR should be opened in foreman-installer which overrides the default $allowed_{import,export}_path parameters for the Katello scenario only, and also defines a migration for the Katello scenario that sets the proper values of the parameters for users who already installed

jeremylenz:
@wbclark, Thanks for guiding this newbie :+1: 

> This PR should include only the new parameter for $allowed_export_path

updated

> A second PR should be opened in puppet-FPC which exposes $allowed_{import,export}_path parameters at https://github.com/theforeman/puppet-foreman_proxy_content/blob/master/manifests/init.pp#L371-L394

Opened https://github.com/theforeman/puppet-foreman_proxy_content/pull/297

> A 3rd PR should be opened in foreman-installer which overrides the default $allowed_{import,export}_path parameters for the Katello scenario only, and also defines a migration for the Katello scenario that sets the proper values of the parameters for users who already installed

Opened https://github.com/theforeman/foreman-installer/pull/615

Let me know how they look..

manifests/init.pp
```diff
@@ -175,6 +178,7 @@
   Integer[0] $redis_db = 8,
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
+  Array[Stdlib::Absolutepath] $allowed_export_path = ['/var/lib/pulp/exports'],
```
wbclark, line 181:
```suggestion
  Array[Stdlib::Absolutepath] $allowed_export_path = [],
```
My thinking is that we should try as best as possible to stick with defaults within this module that would make sense for a user who just wants to install pulp3 with puppet.

Since we'll expose this parameter in puppet-FPC and in the installer answers files for Katello and FPC scenarios, those are more appropriate places to set values that make sense specifically for Katello / content proxy.

templates/settings.py.erb
```diff
@@ -38,8 +38,11 @@ REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
     'pulpcore.app.authentication.PulpRemoteUserAuthentication'
 )
 
-<%# This setting whitelists paths that can be used for repository sync with file protocol. Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
+<%# These settings specify paths that can be used for repository sync with file protocol, -%>
+<%# and for content view version import/export. -%>
+<%# Katello uses the path /var/lib/pulp/sync_imports/ to run tests -%>
```
ekohl, line 43: I think these comments are not that good. The settings are documented [here](https://docs.pulpproject.org/pulpcore/settings.html#allowed-import-paths) and [here](https://docs.pulpproject.org/pulpcore/settings.html#allowed-export-paths) which is much more important than what Katello does with it.

wbclark:
Hey @jeremylenz thanks for the effort here! It's coming together nicely.

I left one comment above about setting defaults that make sense for pulp3 on its own.

The other thing I want to mention, following a similar line of reasoning, is the parameter documentation making references to concepts like Katello and Content Views, which pulpcore itself has no concept of. I think having more neutral documentation (see @ekohl 's suggestion referencing the pulpcore docs) makes sense here, while in puppet-FPC we can document the relevance to Katello users.

I also realize I am saying this that some of those references were already here before your PR (in the ERB...) -- sorry that you inherited that and let us know if we can assist with cleaning it up. :)

templates/settings.py.erb
jeremylenz: @ekohl I was just updating the comment so it would remain accurate.  But I don't have a strong opinion about the proper place for Katello-specific comments, so I'll go ahead and just remove the comment entirely.  would you be satisfied with that?

ekohl: Yes, that's fine by me.

jeremylenz:
@wbclark @ehelms @ekohl I tried to address comments on all 3 PR's.. let me know what I missed.  thanks!

spec/classes/pulpcore_spec.rb
```diff
@@ -347,6 +347,30 @@
...
+            .with_content(%r{ALLOWED_EXPORT_PATHS = \["/test/path", "/test/path2"\]})
+
+        end
+      end
+
+      context 'with empty allowed export paths' do
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{ALLOWED_EXPORT_PATHS = \["/var/lib/pulp/exports"\]})
```
wbclark, line 369:
```suggestion
            .with_content(%r{ALLOWED_EXPORT_PATHS = \[]})
```
This test needs to be updated to reflect the default that we set.

manifests/init.pp
```diff
@@ -123,7 +123,10 @@
 #   Django remote user environment variable
 #
 # @param allowed_import_path
-#   Allowed paths that pulp can sync from using file:// protocol
+#   Allowed paths that pulp can use for content view imports, or sync from using file:// protocol
+#
+# @param allowed_export_path
+#   Allowed paths that pulp can use for content view exports
```
ehelms, line 129: I'd drop the `view` wording here so its just `content exports`

manifests/init.pp
```diff
@@ -123,7 +123,10 @@
 #   Django remote user environment variable
 #
 # @param allowed_import_path
-#   Allowed paths that pulp can sync from using file:// protocol
+#   Allowed paths that pulp can use for content view imports, or sync from using file:// protocol
```
ehelms, line 126: I'd drop the `view` wording here so its just `content exports`

spec/classes/pulpcore_spec.rb
```diff
@@ -347,6 +347,30 @@
...
+        end
+      end
+
+      context 'with empty allowed export paths' do
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{ALLOWED_EXPORT_PATHS = \[\]})
+
+        end
```
ekohl, line 371: This matches the default. I'd move this to the section that starts at line 19. We should probably do the same for import paths.

jeremylenz:
@ekohl updated

ehelms (approved):

ekohl (commented):
Not sure why this was merged. I still had comments about the specs and it was not addressed as I suggested. I also think this should have been squashed and not as 6 separate commits.

ehelms:
Sorry, it looked like your comment had been resolved. I do wish Github would warn me when there are multiple commits, that was my bad.

PR #152, opened 2020-12-04
Allow configuring the registry url for the container plugin

ehelms:
Katello needs to be able to set the registry URL to an API end
point on the Foreman server to control auth for the registry. This
defaults to the Pulpcore API but provides a parameter to configure
this to allow pointing to the Katello endpoint instead.

### DIFF
--- a/manifests/plugin/container.pp
+++ b/manifests/plugin/container.pp
@@ -8,7 +8,15 @@
 class pulpcore::plugin::container (
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
+  String $registry_base_url = '',
 ) {
+
+  if $registry_base_url == '' {
+    $_registry_base_url = $pulpcore::apache::api_base_url
+  } else {
+    $_registry_base_url = $registry_base_url
+  }
+
   $context = {
     'directories' => [
       {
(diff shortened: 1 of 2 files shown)


manifests/plugin/container.pp
```diff
@@ -8,15 +8,23 @@
 class pulpcore::plugin::container (
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
+  String $registry_base_url = '',
```
ekohl, line 11:
```suggestion
  Optional[String] $registry_base_url = undef,
```

manifests/plugin/container.pp
```diff
@@ -8,15 +8,23 @@
...
 ) {
+
+  if $registry_base_url == '' {
+    $_registry_base_url = $pulpcore::apache::api_base_url
+  } else {
+    $_registry_base_url = $registry_base_url
+  }
```
ekohl, line 18:
```suggestion
  $_registry_base_url = pick($registry_base_url, $pulpcore::apache::api_base_url)
```

manifests/plugin/container.pp
```diff
@@ -8,15 +8,23 @@
 class pulpcore::plugin::container (
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
+  String $registry_base_url = '',
 ) {
+
```
ekohl, line 13: Empty line here breaks linting in the future (once I get to updating `voxpupuli-test`)

ekohl (changes requested):
Please document the parameter.

ehelms:
This change might be the result of a wild goose chase for an issue we are seeing. Will close it if that turns out to be the case.

PR #153, opened 2020-12-07
Set the reverse proxy host to the name of the service

ehelms:


ehelms:
This change stems from the breakage and investigation here -- https://community.theforeman.org/t/katello-nightly-rpm-pipeline-769-failed/21468/4?u=ehelms

The shadowing of the `/v2/` in Pulpcore API by Katello to provide auth causes it to clash with the reverse proxying within the same Vhost to Foreman. This appears to be tied to how the unix socket based reverse proxy handles things.

evgeni (approved):

PR #156, opened 2020-12-09
Fixes #31468 - create import/export directories

ehelms:
(cherry picked from commit 485fea4fac7463a90a3348351a9cdf6b9961c078)


wbclark (approved):
This depends on some of @jeremylenz 's prior work which is already in the 2.2-stable branch so it looks good to me.

PR #159, opened 2020-12-23
Correct directory tree in README

ekohl:
23bff59eafb60c17ff04561ce55ca8faeb71ebd9 changed the docroot default, but I forgot this entry.


wbclark (approved):
ACK, nice catch

PR #161, opened 2021-01-11
Add proxy params to plugin Pulp 2 content routes

ehelms:
For legacy content access, the old Pulp 2 routes are used and proxy directly to the content app rather than redirecting. Thus, they also need to have the proxy parameters such as timeout specified on them.


ekohl (approved):

ekohl:
This completes da8f775dc4f144a6d1ddfa1f8a0cf7e55b9b9af9. It's always good to mention that.

PR #163, opened 2021-01-19
Refs #31670 - don't timeout DB migrations

evgeni:


manifests/database.pp
```diff
@@ -14,6 +14,7 @@
   }
 
   pulpcore::admin { 'migrate --noinput':
+    timeout     => 0,
```
ehelms, line 17: I assume we should adopt the same strategy as puppet-foreman here? Add a parameter to this class just in case, and default to 3600.

evgeni: yepp, updated to match the way we discussed for puppet-foreman.

manifests/admin.pp
```diff
@@ -28,6 +31,7 @@
   Array[Stdlib::Absolutepath] $path = ['/usr/bin'],
   String $user = $pulpcore::user,
   Stdlib::Absolutepath $pulp_settings = $pulpcore::settings_file,
+  Optional[Integer] $timeout = undef,
```
ekohl, line 34:
```suggestion
  Optional[Integer[0]] $timeout = undef,
```

ekohl (approved):

PR #164, opened 2021-01-20
Support Pulp 3.9, drop earlier versions

ekohl:
Pulp 3.9 changes CHUNKED_UPLOAD_DIR from an absolute path to a relative path. It must be within MEDIA_ROOT and there's no need to manage it.  Pulp defaults are correct and overriding it is not recommended upstream.

This does rely on a fix in Pulpcore 3.9.1 which is currently [being worked on](https://github.com/pulp/pulpcore/pull/1078). Hence why this is a draft.

Technically older versions will continue to work but the upload directory may not exist and the pulp user may not have permission to create it. Rather than documenting these, it's chosen to only support the latest version. This also saves CI resources.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -37,8 +37,7 @@ jobs:
           - "6"
           - "5"
         pulpcore_version:
-          - '3.6'
-          - '3.7'
+          - '3.9'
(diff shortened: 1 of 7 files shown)


ekohl:
cc @ianballou @jlsherrill

jlsherrill:
looks good to me, @ianballou does that match your understanding?

ehelms (approved):

ianballou (approved):
Looks good to me too.

PR #170, opened 2021-02-05
Fixes #31815 - Allow setting number of workers for content app

ehelms:


ekohl (commented):
In other places we have set an upper bound. This is important on large systems with 10s of CPUs. I think we should do the same here.

ehelms:
> In other places we have set an upper bound. This is important on large systems with 10s of CPUs. I think we should do the same here.

Which should be?

ekohl:
Now I had hoped you had an answer for that. We do see that `$worker_count` is limited to 8 CPUs so if we follow that logic, perhaps it should limit content as well on 8 CPUs (which will result in 17 workers).

ehelms:
Sounds reasonable to go with how we set this in other places, and then see how the tuning plays out.

manifests/init.pp
```diff
@@ -135,6 +135,10 @@
 # @param service_ensure
 #   Specify if Pulp services should be running or stopped.
 #
+# @param content_worker_count
+#   Number of pulpcore-content service workers for gunicorn to use. Defaults to (2 * CPU + 1).
```
ekohl, line 139: Typically we keep the defaults out of param strings since the generated reference documentation includes them. That prevents them from going out of sync, which already happened in the same PR.
```suggestion
#   Number of pulpcore-content service workers for gunicorn to use.
```

ekohl (changes requested):

wbclark:
Is there any documentation on the difference between pulpcore workers and pulpcore content app workers? I'm not finding it readily

I did find however some more notes about hardware requirements vs worker counts at https://docs.pulpproject.org/pulpcore/components.html#hardware-requirements

We currently consider CPUs but not memory, which looks like it can also be a significant bottleneck. In particular for the RPM content plugin a single worker can consume several GBs of memory while syncing. In the future we should at least update the docs here to include this information.

ehelms:
pulpcore workers are the background workers, content app workers are really gunicorn workers for the webserver piece of the content app.

ekohl:
I wouldn't be surprised if at some point we also need to be able to tune the API workers. If you have more jobs and end up polling more, a single worker may not be able to handle it anymore. It would be good to add a parameter for that as well, even if we hardcode it to a single worker now.

ehelms:
Related to what sparked this change I have a pulp-list thread started -- https://listman.redhat.com/archives/pulp-list/2021-February/msg00007.html

ekohl:
Regarding the mentioned issue: we can't use the event worker on EL7's httpd. In https://github.com/theforeman/foreman-installer/commit/0ed45f776fdbeb5d5dc814ffeab355da6aefce51 we reverted this. There's a bugzilla somewhere that's closed as WONTFIX so we either need to move to the httpd SCL or drop EL7 for that.

ehelms:
Added an entry for the API service to adjust worker counts as well.

manifests/init.pp
```diff
@@ -178,6 +186,8 @@
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
+  Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
+  Integer[0] $api_service_worker_count = 1,
```
ekohl, line 190: Note to self: this matches the default according to https://docs.gunicorn.org/en/latest/settings.html#workers

ekohl (approved):

PR #171, opened 2021-02-08
Fixes #31835: Add disablereuse=on for pulpcore-content service

ehelms:


ehelms:
See discussion for more details: https://listman.redhat.com/archives/pulp-list/2021-February/msg00007.html

wbclark (approved):
After reviewing the discussion on pulp-list, it looks good to me

Based on the discussion on the aiohttp issue it looks like adding the required feature to make it work without `disablereuse=on` is not expected any time soon. I understand those who have tested this configuration haven't seen any performance degradation. So I conclude that we probably don't need any kind of note to remove this configuration later

If it's still the case that some of our modules can deploy pulpcore without using the reverseproxy configuration from this module, then we may need similar changes in those modules

ekohl (approved):
Given we're using unix sockets, reusing them is not as big a gain as when the backend has a high latency.

PR #174, opened 2021-03-08
Support Puppet 7

ekohl:


wbclark:
Re-running tests. I expect some will be fixed by https://github.com/theforeman/puppet-pulpcore/commit/30d4ecacdc4a98db88e08c2efb5a387b31e7aa53 while others would still fail due to `::systemd::systemctl::daemon_reload` in puppet-redis

ekohl:
Rebased which also happens to trigger tests as well.

ekohl:
Rebased again, let's see if the tests do pass now or if we really did hit a bug.

ekohl:
It does look persistent.

ehelms (approved):

PR #175, opened 2021-03-11
Pulpcore logging format to include correlation id.

jjeffers:


jlsherrill:
This looks good to me, just need tests to be fixed

ekohl (commented):
Do we only have correlation IDs in the API service? Is this also something that should be logged in workers?

jlsherrill:
According to @daviddavis the workers are passed a correlation id but log it by default, no need for special configuration

daviddavis:
> Do we only have correlation IDs in the API service? Is this also something that should be logged in workers?

Assuming that you're using the default for LOGGING, [it should include the correlation id in the django logs](https://github.com/pulp/pulpcore/blob/00497f89e06e50a1b15854a9917f7bab1848a40f/pulpcore/app/settings.py#L195) which applies to the workers. This line here is to configure gunicorn to also output the correlation id.

ekohl:
Ok, that's good. I also don't think content has correlation IDs. Correct?

daviddavis:
That's correct: the content app does not support correlation id. If that's something you all need, I can take a look.

ekohl:
I don't think so but I would like to make sure this PR covers everything we need. I'd like to avoid doing multiple PRs, each for a single service.

jjeffers:
@ekohl @ehelms @jlsherrill anything else that should be changed for this PR?

jlsherrill (approved):

jlsherrill:
@ekohl okay to merge?

ehelms (approved):

PR #176, opened 2021-03-15
Support camptocamp/systemd 3.0

ekohl:
The development branch of camptocamp/systemd has dropped support for Puppet 5 but for now we keep our modules compatible with it. This allows us to remain compatible with both.

Technically it's 6.1.0 that automatically does daemon reloads if needed.


ekohl:
It appears the trick doesn't work since for the acceptance tests we always install Puppet 6 in the "host". It also looks like there's an issue loading the socket service. Going to look into that as well.

ekohl:
I've cleaned it up a bit but this still needs work. It may be better to revisit this once the module is actually out we need to start to consider this.

PR #178, opened 2021-03-17
Fixes #32112 - Configure api service gunicorn worker timeout

wbclark:


wbclark:
If there is a need to do so, we can configure the timeout for the content service gunicorn workers as well. However the specific issue I faced when syncing multiple large RPM repositories required the gunicorn worker timeout for the api service to be increased from the default 30 seconds.

manifests/init.pp
```diff
@@ -188,6 +191,7 @@
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
   Integer[0] $api_service_worker_count = 1,
+  Integer[1] $api_service_worker_timeout = 60,
```
ekohl, line 194: https://docs.gunicorn.org/en/stable/settings.html#timeout states:
> Value is a positive number or 0. Setting it to 0 has the effect of infinite timeouts by disabling timeouts for all workers entirely.
```suggestion
  Integer[0] $api_service_worker_timeout = 60,
```

ekohl (changes requested):

wbclark:
@ekohl updated, thanks.

ehelms (approved):

ekohl (approved):

PR #179, opened 2021-03-19
Refs #32112 - Match upstream pulp default api service worker timeout

wbclark:
Upstream pulp_installer uses a default gunicorn worker timeout of 90 seconds[1] for both the content service (which we previously did not configure) and the api service (which we configured to 60 seconds)

[1] https://github.com/pulp/pulp_installer/blob/645887df342f3d06a2aa8ec90740bd41179331c7/roles/pulp_content/templates/pulpcore-content.service.j2#L24

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -143,6 +143,9 @@
 #   Number of pulpcore-api service workers for gunicorn to use.
 #   Modifying this parameter should be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param content_service_worker_timeout
+#   Timeout in seconds of the pulpcore-content gunicorn workers.
+#
 # @param api_service_worker_timeout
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -191,7 +191,8 @@
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
   Integer[0] $api_service_worker_count = 1,
-  Integer[0] $api_service_worker_timeout = 60,
+  Integer[0] $content_service_worker_timeout = 90,
```
ehelms, line 194: Need to add to module param docs above

wbclark: Right, thanks

ehelms (approved):

PR #181, opened 2021-04-13
Fixes #32309 - pulpcore-manager fails from certain directories

ianballou:
Sets a specific directory for pulpcore-manager to be called from.

### DIFF
--- a/manifests/admin.pp
+++ b/manifests/admin.pp
@@ -23,6 +23,9 @@
 # @param timeout
 #   The command should timeout after so many seconds.
 #
+# @param working_dir
+#   The directory to run pulpcore-manager from.
+#
 # @see exec
 define pulpcore::admin(
   String $command = $title,
@@ -32,11 +35,13 @@ define pulpcore::admin(
   String $user = $pulpcore::user,
   Stdlib::Absolutepath $pulp_settings = $pulpcore::settings_file,
   Optional[Integer[0]] $timeout = undef,
+  Stdlib::Absolutepath $working_dir = $pulpcore::user_home,
 ) {
   Concat <| title == 'pulpcore settings' |>
   -> exec { "pulpcore-manager ${command}":
     user        => $user,
     path        => $path,
+    cwd         => $working_dir,
     environment => ["PULP_SETTINGS=${pulp_settings}"],
     refreshonly => $refreshonly,
     unless      => $unless,
--- a/spec/defines/admin_spec.rb
+++ b/spec/defines/admin_spec.rb
@@ -11,6 +11,7 @@ describe 'pulpcore::admin' do
           {
             pulp_settings: '/etc/pulpcore/settings.py',
             user: 'pulpcore',
+            working_dir: '/var/lib/pulp'
           }
         end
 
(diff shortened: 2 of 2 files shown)


manifests/admin.pp
```diff
@@ -32,11 +35,13 @@
   String $user = $pulpcore::user,
   Stdlib::Absolutepath $pulp_settings = $pulpcore::settings_file,
   Optional[Integer[0]] $timeout = undef,
+  String $working_dir = $pulpcore::user_home,
```
ekohl, line 38: I prefer to be stricter, even if it doesn't really make a difference.
```suggestion
  Stdlib::Absolutepath $working_dir = $pulpcore::user_home,
```

ekohl (commented):
The change itself makes sense to me. However, please make the commit message more descriptive. Like `Always run admin cmds from Pulp's home` (not sure how much will fit). Ideally with then a body that describes that the cwd might be an unreadable directory like `/root` and dynaconf appears to unconditionally read the current directory. Pulp's user home should always be readable for the user pulp.

Also, it might be good to link to a dynaconf bugreport. IMHO it shouldn't error out on unreadable CWD.

ianballou:
Not sure yet why the tests are complaining about `Unknown variable: 'pulpcore::user_home'.`

ekohl:
In that testcase we don't include the class `pulpcore` so it can't reference the variable. That's why we explicitly set those in the tests:
https://github.com/theforeman/puppet-pulpcore/blob/9f680a141dda160d7a21f562c7392ba93afae7e8/spec/defines/admin_spec.rb#L9-L15

ianballou:
@ekohl I think we're all good now.

spec/defines/admin_spec.rb
```diff
@@ -11,6 +11,7 @@
           {
             pulp_settings: '/etc/pulpcore/settings.py',
             user: 'pulpcore',
+            working_dir: '/var/lib/pulp'
```
wbclark, line 14: Hey @ianballou , could you please add to the test below, under `is_expected.to contain_exec('pulpcore-manager help')`,

`with_cwd('/var/lib/pulp')`

This way we are not only passing the param from the parent module which `pulpcore::admin` needs, we are also testing that the feature works properly at the business logic level (puppet catalogue) and continues working as expected with any future changes.

ianballou:
@wbclark okay, I added the test condition.

wbclark:
> @wbclark okay, I added the test condition.

Great, thank you. This looks good to me. Let's see what @ekohl says

ekohl (approved):
This is what I would have written as well

goosemania:
> The change itself makes sense to me. However, please make the commit message more descriptive. Like `Always run admin cmds from Pulp's home` (not sure how much will fit). Ideally with then a body that describes that the cwd might be an unreadable directory like `/root` and dynaconf appears to unconditionally read the current directory. Pulp's user home should always be readable for the user pulp.
> 
> Also, it might be good to link to a dynaconf bugreport. IMHO it shouldn't error out on unreadable CWD.

FWIW, tried to fix, this helped https://github.com/rochacbruno/dynaconf/pull/570

PR #183, opened 2021-04-15
Add support for ALLOWED_CONTENT_CHECKSUMS

jlsherrill:


manifests/init.pp
```diff
@@ -188,6 +191,7 @@
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Array[Stdlib::Absolutepath] $allowed_export_path = [],
+  Array[String] $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
```
evgeni, line 194: Should this be Array[Enum[…]] to limit possibilities to only known ones? Or would that be to limiting for future pulpcore releases?

jlsherrill: I think that makes sense?  we'd need to update it anyways to adjust the default i would think?  Its unlikely to be a frequent occurrence either :)

jlsherrill:
note that these are the pulp defaults.  Katello will likely want to default to allowing 'sha1' as well.  Going to open a pr to foreman_proxy_content to do that and make it user configurable

manifests/init.pp
jlsherrill: added!

wbclark (approved):
This looks good to me.

wbclark:
If there is an existing Redmine for the issue, could you please update the commit message with it?

This one could be "Refs #xxxxx - Add support for ALLOWED_CONTENT_CHECKSUMS" and the puppet-FPC PR could be "Fixes #xxxxx - expose pulpcore allowed_content_checksums"

manifests/init.pp
ehelms: If pulpcore releases a new supported checksum, instead of say puppet-foreman_proxy_content updating what it passes to this parameter to include that new parameter, a release of this module will have to be made to update the list of allowed values to then allow the update to puppet-foreman_proxy_content as a consumer. It is balancing specifying only "valid" values and future proofing. Does Pulp throw an error if an incorrect value is supplied?

jlsherrill: You get an error on startup of the workers:
django.core.exceptions.ImproperlyConfigured: ALLOWED_CONTENT_CHECKSUMS may only contain algorithms known to pulp - see constants.ALL_KNOWN_CONTENT_CHECKSUMS for the allowed list.

Also, sha256 is required, it won't startup without, not sure if we should handle that here.  @ehelms you bring up a good point, maybe we should drop the Enum here and only have it in the foreman_proxy_content change?

ehelms: I lean towards not having the Enum personally to allow easier future proofing. This is usually where @ekohl shows up and sets us all straight :)

ekohl: If you want to reuse it, you can define it a type alias. https://github.com/theforeman/puppet-foreman_proxy/blob/master/types/listenon.pp is probably the easiest example from our codebase. So you would end up with
```puppet
type Pulpcore::ContentChecksums = Array[Enum[...], 1]
```

Then you can use, both in this module and puppet-foreman_proxy_content, `Pulpcore::ContentChecksums $allowed_content_checksums`. Benefits are that you only need to maintain it in one place while still having strong validation.

https://rspec-puppet.com/documentation/type_aliases/ documents how to write simple tests for it.

It's not that often that a new checksum type is added and it needs actual code on the Pulp side. Given that there is a list of constants, I do lean to an enum. Especially if we plan to expose this to users. Debugging a failed service start is a much worse experience than an up front error.

manifests/init.pp
```diff
@@ -188,6 +191,7 @@
   Stdlib::Fqdn $servername = $facts['networking']['fqdn'],
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Array[Stdlib::Absolutepath] $allowed_export_path = [],
+  Array[Enum['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']] $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
```
ekohl, line 194: I did suggest to extract the `Array[Enum[]]` to a type alias. That makes it easy to reuse in another module.

ekohl (changes requested):

manifests/init.pp
jlsherrill: ahh, thanks for clarifying, will add that

ekohl: I should have linked back to https://github.com/theforeman/puppet-pulpcore/pull/183#discussion_r615239844 for that suggestion.

jlsherrill: added!

types/checksumtypes.pp
```diff
@@ -0,0 +1,2 @@
+# The supported checksum types by pulpcore
+type Pulpcore::ChecksumTypes = Enum['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
```
ekohl, line 2: I think if it's an enum (which results in a single value), I'd make this singular:
```suggestion
type Pulpcore::ChecksumType = Enum['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
```

You can also keep it plural by making it an array:
```suggestion
type Pulpcore::ChecksumTypes = Array[Enum['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512'], 1]
```

The last part enforces that there's at least one value and an empty array is invalid (which I think makes Pulp refuse to start up).

Given that last part of additional validation that you don't want to copy to multiple places, I'd suggest to at least have the plural type. If you choose to have a singular type too (which the plural type can use), but that may be premature optimization.

jlsherrill: I like the plural, i will adapt to that

jlsherrill:
If this looks good i'll update the foreman proxy content pr to match

ekohl (approved):
This was supported in 3.9, right? (which is currently the only version we support in this module)

jlsherrill:
yes, its supported all the way back to 3.7: https://github.com/pulp/pulpcore/blob/3.7/pulpcore/app/settings.py#L235

jlsherrill:
and if you want to wait for https://github.com/theforeman/puppet-pulpcore/pull/184 for a release, i think that makes sense

evgeni:
And now nightly fails with 

```
[ERROR ] [root] Parameter foreman-proxy-content-pulpcore-allowed-content-checksums invalid: Elements of the array are invalid: ['sha1', 'sha224', 'sha256', 'sha384', 'sha512'] must be one of md5, sha1, sha224, sha256, sha384, sha512
```

Why?!

ekohl:
Ah, it's https://projects.theforeman.org/issues/31565 again with this line here:
https://github.com/theforeman/puppet-foreman_proxy_content/blob/447cff81db66c6b66315a1dc5661cbe57f891539/manifests/init.pp#L116

evgeni:
Can we band-aid it somehow?

ekohl:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/346

PR #184, opened 2021-04-19
Fixes #31950 - support ansible plugin

jlsherrill:


ekohl (commented):
In https://github.com/theforeman/puppet-pulpcore/blob/master/spec/acceptance/plugins_spec.rb we also have code to install it during an acceptance test. This means an actual RPM installation which has proven very useful.

jlsherrill:
@ekohl thanks for the pointer, i added the ansible install to the acceptance test and getting:

seeing:
```
   Error: /Stage[main]/Pulpcore::Plugin::Ansible/Pulpcore::Plugin[ansible]/Package[python3-pulp-ansible]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install python3-pulp-ansible' returned 1: Error: 
   Problem: package python3-pulp-ansible-1:0.6.1-1.el8.noarch requires python3-galaxy-importer, but none of the providers can be installed
    - conflicting requests
    - nothing provides /usr/bin/ansible-test needed by python3-galaxy-importer-0.2.12-1.el8.noarch
    - nothing provides ansible needed by python3-galaxy-importer-0.2.12-1.el8.noarch
```
i assume i need to enable the ansible or epel repo, should i do that as part of this module? (That seems incorrect).  Should we just not include it in the acceptance test?
Thanks

ekohl:
https://github.com/theforeman/puppet-pulpcore/blob/master/spec/setup_acceptance_node.pp does feel like the best place for that and https://github.com/theforeman/puppet-foreman_proxy/blob/afb45386c3b36201bf0f95ad02151d15bca01fa5/spec/setup_acceptance_node.pp#L15-L19 is probably the right bit to include for that.

jlsherrill:
uggg i think i'm hitting https://github.com/theforeman/pulpcore-packaging/issues/138  which seems like a nasty issue :(

jlsherrill:
ahh, got it figured out.  the centos 7 docker container already had epel enabled which hit https://github.com/theforeman/pulpcore-packaging/issues/138  but centos8 either didnt' have it enabled, or epel 8 doesn't include ansible

spec/classes/plugin_ansible_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+          .that_subscribes_to('Class[Pulpcore::Install]')
+          .that_notifies(['Class[Pulpcore::Database]', 'Class[Pulpcore::Service]'])
+        is_expected.to contain_apache__vhost__fragment('pulpcore-https-plugin-ansible')
+                           .with_content(
+<<CONTENT
+
+  ProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
+  ProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
+CONTENT
+                           )
```
wbclark, line 24: Is there any way we can make this nicer to look at ?

I tried:

```
          .with_content(
            <<~CONTENT.chomp
            
              \s\sProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
              \s\sProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
            CONTENT
          )
```

But rpsec doesn't seem to be able to treat the `\s` sequences as spaces:

```
       expected that the catalogue would contain Apache::Vhost::Fragment[pulpcore-https-plugin-ansible] with content set to supplied string
       Diff:
         <The diff is empty, are your objects producing identical `#inspect` output?>
```

The `#strip_heredoc` method from rails would do what I'm looking for but it's not available here. I also tried interpolation to get the two spaces but rspec fails in the same way as above.

What finally worked for me is defining the `VHOST_CONTENT` at the top of the file:

```
VHOST_CONTENT=<<HEREDOC
  ProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
  ProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
HEREDOC
```

And then the test becomes:

```suggestion
          .with_content(VHOST_CONTENT)
```

Open to other ideas if you have them :)

ekohl: > But rpsec doesn't seem to be able to treat the `\s` sequences as spaces:

That's because heredoc is a string, not a regex.

> Open to other ideas if you have them :)

You can consider a let block which is probably better than a constant:
```ruby
let(:expected_vhost_content) do
  'the expected content'
end
```

Another option is to use `.with_content` twice if you don't care where it exactly is, just that it's there:
```ruby
.with_content(%r{ProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock\|http://pulpcore-api/pulp_ansible/galaxy/})
.with_content(%{ProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock\|http://pulpcore-api/pulp_ansible/galaxy/})

File: spec/setup_acceptance_node.pp [hunk 2]
```diff
@@ -23,6 +23,12 @@
   default: {}
 }
 
+if $facts['os']['name'] == 'CentOS' and $major == '8' {
+  package { 'centos-release-ansible-29':
+    ensure => present,
+  }
+}
```
ekohl, line 30: Looking at the `case $major` block, it appears we already assume it's CentOS for version 7 anyway. Perhaps the `package` definition should move to the block on line 19? Then the conditional part can be removed.

spec/classes/plugin_ansible_spec.rb
jlsherrill: I tried moving it to a let as a string:

```
      let(:expected_vhost_content) do
        "\n  ProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/\n" +
            "  ProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/"
      end
```

```
        is_expected.to contain_apache__vhost__fragment('pulpcore-https-plugin-ansible')
                           .with_content(expected_vhost_content)
```

But i get:
```

  1) pulpcore::plugin::ansible on redhat-7-x86_64 is expected to contain Apache::Vhost::Fragment[pulpcore-https-plugin-ansible] with content  supplied string
     Failure/Error:
       is_expected.to contain_apache__vhost__fragment('pulpcore-https-plugin-ansible')
                          .with_content(expected_vhost_content)
       expected that the catalogue would contain Apache::Vhost::Fragment[pulpcore-https-plugin-ansible] with content set to supplied string
       Diff:
         <The diff is empty, are your objects producing identical `#inspect` output?>
```

but defining the let with a HEREDOC works:
```
      let(:expected_vhost_content) do
<<CONTENT

  ProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
  ProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
CONTENT
      end
```

IDK i'm fine with the heredoc, in this case i think its important to verify the content exactly (and not in a location)

ekohl: I feel like a large inline block of texts makes it harder to read and would prefer a `let` block.

jlsherrill: to clarify, you mean  a let block with heredoc as well?

jlsherrill: went ahead and pushed that, let me know if that wasn't what you meant

ekohl (approved):
Thanks!

ekohl:
(tests are still running now)

spec/classes/plugin_ansible_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      it { is_expected.to compile.with_all_deps }
+      it { is_expected.to contain_package('python3-pulp-ansible') }
+      it { is_expected.to contain_pulpcore__plugin('ansible') }
+      it do
+        is_expected.to compile.with_all_deps
+        is_expected.to contain_pulpcore__plugin('ansible')
+          .that_subscribes_to('Class[Pulpcore::Install]')
+          .that_notifies(['Class[Pulpcore::Database]', 'Class[Pulpcore::Service]'])
+        is_expected.to contain_apache__vhost__fragment('pulpcore-https-plugin-ansible')
+                           .with_content(expected_vhost_content)
```
wbclark, line 25:
```suggestion
          .with_content(
            Regexp.new <<~CONTENT
              \n\s\sProxyPass /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
              \s\sProxyPassReverse /pulp_ansible/galaxy/ unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp_ansible/galaxy/
            CONTENT
          )
```

This works for me, has more readable indentation, and avoids the `let` block entirely. What do you think @jlsherrill @ekohl ?

Thanks to you both for going down this rabbit hole with me. I learned a lot from this :)

ekohl: I think that the current form is the most readable. We use the same pattern in multiple places.

wbclark: It looks horrible in the narrow suggested change window due to line wrapping, but much nicer inside any editor

jlsherrill: I'm not as big of a fan due to the \s characters personally.  All of this is kinda personal preference i guess

wbclark: IMO the `let` block is unnecessary because it's only used once. I am an rspec noob though so please `let` (har) me know if I am missing something. :)

With that said, while the current implementation is not my favorite due to the indentation, I see no reason to delay merging this further over style matters. If you are satisfied, please go ahead

wbclark (approved):
LGTM

as an aside, a friend suggested to me that creating a fixture is another strategy we could have taken to make the style nicer while properly testing for the whitespace in the configuration. that might be useful in the future

jlsherrill:
I think this is ready for merge & release?

ekohl:
> as an aside, a friend suggested to me that creating a fixture is another strategy we could have taken to make the style nicer while properly testing for the whitespace in the configuration. that might be useful in the future

Fixtures can help, but it depends. The benefit is that you can extract large blocks, but it also means that if you make a change, you now need to edit two files. It becomes worse if you do need a variable. It's all trade offs. As developers we need to know the possibilities, their upsides and downsides. Then we must make a choice on what to use. Often we do that by placing weights on the various upsides and downsides. Not everyone applies the same weights but also not everyone knows of all the possibilities. That's why in code review we must figure out whether the weights are different or they missed an option.

As an author, it can help speed up this process if you write down what your process was. I've learned the habit of doing so i the commit message (which Github helpfully copies to the PR description). This allows reviewers to bypass the guessing part and move straight to the validation part.

This doesn't just apply to relatively unimportant things. In fact, it's even more important for big choices. Architectural choices come to mind. If you get in the habit of writing things down in commit messages, it becomes second nature.

More on topic: thanks for the contribution and chance for some education @jlsherrill!

PR #186, opened 2021-04-27
Refs #32383: Configurable client certificate authentication to Pulp

ehelms:
…o Pulp as admin

Require a client certificate to be presented with a CN that matches
the supplied auth CN. If this is present, set the REMOTE_USER to
admin to pass along to Pulp. This changes from having to generate
aclient certificate with a valid user (e.g. admin) as the CN
to allowing to use a client certificate generated with a more standard
CN (e.g. FQDN) and act as the admin user.

Example usage:

```
class { 'pulpcore':
  api_client_auth_cn => $foreman_host
}
```

Questions:

 * This would allow us to drop the special purpose pulp_client certificates we generate today. Do we want to go this route?
 * Is this the best way implement how we enable this and configure it?
 * This forces all API paths to be behind authentication now. The `/pulp/api/v3/status/` does not require authentication by Pulp. Do we want an exception for that route? If so, how do we idenitfy that path in the Apache config to skip the require?

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -44,6 +44,16 @@ class pulpcore::apache (
   # Pulp has a default for remote header. Here it's ensured that the end user
   # can't send that header to spoof users.
   $remote_user_environ_header = $pulpcore::remote_user_environ_name.regsubst(/^HTTP_/, '')
+
+  $api_default_request_headers = [
+    "unset ${remote_user_environ_header}",
+    "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
+  ]
+
+  $api_additional_request_headers = $pulpcore::api_client_auth_cn_map.map |String $cn, String $pulp_user| {
+    "set ${remote_user_environ_header} \"${pulp_user}\" \"expr=%{SSL_CLIENT_S_DN_CN} == '${cn}'\""
+  }
+
   $api_directory = {
     'path'            => $api_path,
     'provider'        => 'location',
@@ -53,10 +63,7 @@ class pulpcore::apache (
         'params' => $api_proxy_params,
       },
     ],
-    'request_headers' => [
-      "unset ${remote_user_environ_header}",
-      "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
-    ],
+    'request_headers' => $api_default_request_headers + $api_additional_request_headers,
   }
 
   # Static content is served by the whitenoise application. SELinux prevents
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -152,6 +152,9 @@
 # @param api_service_worker_timeout
 #   Timeout in seconds of the pulpcore-api gunicorn workers.
 #
+# @param api_client_auth_cn_map
+#   Mapping of certificate common name and Pulp user to authenticate to Pulp API.
+#
 # @example Default configuration
 #   include pulpcore
 #
@@ -200,6 +203,7 @@ class pulpcore (
   Integer[0] $api_service_worker_count = 1,
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
+  Hash[String[1], String[1]] $api_client_auth_cn_map = {},
 ) {
   $settings_file = "${config_dir}/settings.py"
 
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -428,6 +428,51 @@ CONTENT
           is_expected.to contain_service("pulpcore-worker@1.service").with_ensure(false)
         end
       end
+
+      context 'with API client auth common names' do
+        let :params do
+          {
+            'api_client_auth_cn_map': {'foreman.example.com' => 'admin'}
+          }
+        end
+
+        it do
+          is_expected.to contain_apache__vhost('pulpcore-https')
+            .with_directories([
+              {
+                'provider'       => 'Directory',
+                'path'           => '/var/lib/pulp/pulpcore_static',
+                'options'        => ['-Indexes', '-FollowSymLinks'],
+                'allow_override' => ['None'],
+              },
+              {
+                'path'            => '/pulp/content',
+                'provider'        => 'location',
+                'proxy_pass'      => [{
+                  'url'    => 'unix:///run/pulpcore-content.sock|http://pulpcore-content/pulp/content',
+                  'params' => {'timeout' => '600', 'disablereuse' => 'on'},
+                }],
+                'request_headers' => [
+                  'unset X-CLIENT-CERT',
+                  'set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT',
+                ],
+              },
+              {
+                'path'            => '/pulp/api/v3',
+                'provider'        => 'location',
+                'proxy_pass'      => [{
+                  'url'    => 'unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3',
+                  'params' => {'timeout' => '600'},
+                }],
+                'request_headers' => [
+                  'unset REMOTE_USER',
+                  'set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN',
+                  'set REMOTE_USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == \'foreman.example.com\'"',
+                ],
+              }
+            ])
+        end
+      end
     end
   end
 end


ekohl:
> This would allow us to drop the special purpose pulp_client certificates we generate today. Do we want to go this route?

Yes please. I would love to get rid of those special purpose certificates.

> This forces all API paths to be behind authentication now. The `/pulp/api/v3/status/` does not require authentication by Pulp. Do we want an exception for that route? If so, how do we idenitfy that path in the Apache config to skip the require?

Some other pattern that I saw while reading some docs was this: https://httpd.apache.org/docs/trunk/expr.html#examples. While I haven't tried, it makes me think something like this may work.
```
Header unset REMOTE_USER
<If "%{SSL_CLIENT_S_DN_CN} == 'MY_CN'">
  Header set REMOTE_USER "admin"
</If>
```

ehelms:
> > This would allow us to drop the special purpose pulp_client certificates we generate today. Do we want to go this route?
> 
> Yes please. I would love to get rid of those special purpose certificates.
> 
> > This forces all API paths to be behind authentication now. The `/pulp/api/v3/status/` does not require authentication by Pulp. Do we want an exception for that route? If so, how do we idenitfy that path in the Apache config to skip the require?
> 
> Some other pattern that I saw while reading some docs was this: https://httpd.apache.org/docs/trunk/expr.html#examples. While I haven't tried, it makes me think something like this may work.
> 
> ```
> Header unset REMOTE_USER
> <If "%{SSL_CLIENT_S_DN_CN} == 'MY_CN'">
>   Header set REMOTE_USER "admin"
> </If>
> ```

Solid find, that did handle both situations.

manifests/init.pp
```diff
@@ -200,6 +203,7 @@
   Integer[0] $api_service_worker_count = 1,
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
+  String $api_client_auth_cn = 'admin',
```
ehelms, line 206: By setting this default to `admin` means that this change can be released without breaking consumers of this, primarily Katello which uses client certificates with the CN set to `admin`.

ekohl: Should we use a stricter data type?
```suggestion
  Stdlib::Fqdn $api_client_auth_cn = 'admin',
```

AFAIK, a CN field in a cert should always be a hostname. The FQDN type also allows unqualified names (such as `admin`).

ekohl:
I like that it's now shaping up as an enhancement. The commit message should reflect that.

ehelms:
> I like that it's now shaping up as an enhancement. The commit message should reflect that.

How would you like me to reflect that wording wise?

ekohl:
I'm thinking about:
> Configurable Certificate CN matching + mapping
>
> Prior to this, the Certificate CN was passed directly to Pulp. After this, Apache checks for a match and then sends the value admin.

Reading what I just wrote, it's not really backwards compatible for some use cases. Perhaps if the "destination" was also a parameter then it would really be a mapping. You can also think about `Hash $api_cn_map = { 'admin' => 'admin' }`.

ehelms:
I am confused as to how the current test failure works:

```
  1) basic installation Curl command "https://centos7-64.example.com/pulp/api/v3/" body is expected to contain "artifacts_list"
     Failure/Error: its(:body) { is_expected.to contain('artifacts_list') }
       expected "openapi: 3.0.3\ninfo:\n  title: Pulp 3 API\n  version: v3\n  description: Fetch, Upload, Organize, a...n      type: apiKey\n      in: cookie\n      name: Session\nservers:\n- url: http://pulpcore-api/\n" to contain "artifacts_list"
       
     # ./spec/acceptance/basic_spec.rb:92:in `block (3 levels) in <top (required)>'
```

When I look in a Pulp 3 install on a Katello nightly:

```
[root@pipe-katello-server-nightly-centos8 bats]# curl https://`hostname`/pulp/api/v3/ | grep artifact
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 42473  100 42473    0     0   117k      0 --:--:-- --:--:-- --:--:--  117k
      description: Queues a task that creates a new Collection from an uploaded artifact.
  /pulp_ansible/galaxy/{path}/api/v3/artifacts/collections/:
      operationId: pulp_ansible_galaxy_api_v3_artifacts_collections_create
      description: Create an artifact and trigger an asynchronous task to create Collection
        artifact:
        artifact:
        artifact:
```

I don't see `artifacts_list` in there.

ekohl:
https://github.com/theforeman/puppet-pulpcore/blob/5462f346da26df650290e9ab19cb6a522feb8632/spec/acceptance/basic_spec.rb#L83-L94

Note how first it's unauthenticated and expects it **not** to be there. Then it is authenticated and it's expected to be there. The status endpoint is very different for anonymous and (admin) users.

The client cert is generated here:
https://github.com/theforeman/puppet-pulpcore/blob/5462f346da26df650290e9ab19cb6a522feb8632/spec/setup_acceptance_node.pp#L61

ehelms:
Ahh, I see now. It is not the status API but the root API which returns the OpenAPI spec that differs based on authenticated vs. unauthenticated. That feels a bit odd to get back different API specs based on who you are but that's not what this PR is about -- thanks.

ekohl:
It makes sense that the OpenAPI spec describes what you can actually do vs what you could do if you had permissions. It's also what I used to determine if authentication actually works.

manifests/apache.pp
```diff
@@ -55,8 +55,11 @@
     ],
     'request_headers' => [
       "unset ${remote_user_environ_header}",
-      "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
     ],
+    'custom_fragment' => epp('pulpcore/apache-if.epp', {
```
ehelms, line 59: I had to switch to custom_fragment as that is what puppet-postgresql provides and thus I added it to https://github.com/theforeman/puppet-pulpcore/pull/186/files#diff-6c47f87fb2ecb452123d65c6aa6c6d16bf34eeb418e70b2a6fe2e04efff890caR55 

The resulting formatting of the Apache configuration is ugly no matter what I did. The current way I have it looked the best but makes the templates kinda ugly.

ehelms:
On a Katello nightly this comes out as:

```
  <Location "/pulp/api/v3">
    RequestHeader unset REMOTE_USER
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3 timeout=600
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3
    <If "%{SSL_CLIENT_S_DN_CN} == 'admin'">
      RequestHeader set REMOTE_USER "admin"
    </If>
  </Location>
```

On a proxy:

```
  <Location "/pulp/api/v3">
    Require all granted
      ## Request Header rules
            RequestHeader unset REMOTE_USER
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3 timeout=600
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/pulp/api/v3
              <If "%{SSL_CLIENT_S_DN_CN} == 'admin'">
      RequestHeader set REMOTE_USER "admin"
    </If>

  </Location>
```

ekohl:
Another thought I had (sorry to keep directing you all over the place) is using the conditional part of [mod_headers](https://httpd.apache.org/docs/current/mod/mod_headers.html):
```
RequestHeader unset REMOTE_USER
RequestHeader set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN
RequestHeader set REMOTE_USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == 'some-cn.example.com'"
```

It keeps the option open to use other identities but allows us to remap some. What do you think about this?

ehelms:
> ```
> RequestHeader set REMOTE_USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == 'some-cn.example.com'"
> ```

I went looking for something like this, glad you found it as that makes the code much simpler. Now I find myself asking:

 1) Do we allow only a single mapping? Or support an array of mappings?
 2) Do we always include "set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN"

ekohl:
> Do we allow only a single mapping? Or support an array of mappings?

I was thinking about a simple hash where the key is the source (it matches `SSL_CLIENT_S_DN_CN`) and the value is the destination (`admin`)). I don't know if we may end up wanting to support more than a single (admin) user but it would prepare us for that.

> Do we always include "set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN"

I think it would make sense. It keeps it backwards compatible and allows using multiple users in the future.

ehelms:
> > Do we always include "set REMOTE_USER "%{SSL_CLIENT_S_DN_CN}s" env=SSL_CLIENT_S_DN_CN"
> 
> I think it would make sense. It keeps it backwards compatible and allows using multiple users in the future.

The last thought I had with this was if we wanted, going forward, to support passing any CN presented to Pulp and letting Pulp handle things. In the pure mapping case we are providing a layer in Apache that both restricts the set of possible users and enforces an explicit declaration of who can get to that endpoint. But perhaps that is over reaching and we should just trust Pulp to handle whatever user is thrown at it.

ekohl:
> But perhaps that is over reaching and we should just trust Pulp to handle whatever user is thrown at it.

I'm leaning to this, but I had the same considerations.

ehelms:
The code is much simpler now which I like. This keeps the previous default and adds any specified mappings iteratively.

manifests/apache.pp
```diff
@@ -53,10 +63,7 @@
         'params' => $api_proxy_params,
       },
     ],
-    'request_headers' => [
-      "unset ${remote_user_environ_header}",
-      "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
-    ],
+    'request_headers' => concat($api_default_request_headers, $api_additional_request_headers),
```
ekohl, line 66: You can also write
```suggestion
    'request_headers' => $api_default_request_headers + $api_additional_request_headers,
```

Or keeping the diff smaller, you could keep the hardcoded array:
```suggestion
    'request_headers' => [
      "unset ${remote_user_environ_header}",
      "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
    ] + api_additional_request_headers,
```

(Github butchers the diff)

manifests/init.pp
```diff
@@ -200,6 +203,7 @@
   Integer[0] $api_service_worker_count = 1,
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
+  Optional[Hash] $api_client_auth_cn_map = {},
```
ekohl, line 206: There's no need to use `Optional` if there's a default. `Optional[Hash]` is equivalent to `Variant[Hash, Undef]`. You can also enforce stricter types.
```suggestion
  Hash[Stdlib::Fqdn, String[1]] $api_client_auth_cn_map = {},
```

ekohl: I'm not sure what you mean. One is the key, the other the value. Perhaps the value can be even stricter than `String[1]` since it must be a valid Django username, but Django usernames are looser than FQDNs. They could include `@` for example.

ekohl (approved):
I think this is now a pure enhancement rather than an actual change. Thanks!

PR #188, opened 2021-05-04
Release 3.4.0

wbclark:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## [3.4.0](https://github.com/theforeman/puppet-pulpcore/tree/3.4.0) (2021-05-04)
+
+[Full Changelog](https://github.com/theforeman/puppet-pulpcore/compare/3.3.0...3.4.0)
+
+**Implemented enhancements:**
+
+- Allow Puppet 7 compatible versions of mods [\#187](https://github.com/theforeman/puppet-pulpcore/pull/187) ([ekohl](https://github.com/ekohl))
```
ekohl, line 9: https://github.com/theforeman/puppet-foreman_proxy/issues/671

ehelms:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/350 does require a release of this module (or at least a version bump). Which would be preferred so that I can then update the metadata to require a new minimum version?

wbclark:
> [theforeman/puppet-foreman_proxy_content#350](https://github.com/theforeman/puppet-foreman_proxy_content/pull/350) does require a release of this module (or at least a version bump). Which would be preferred so that I can then update the metadata to require a new minimum version?

1. Do we want to get this into the next z-stream Foreman release ? (If so, I lean towards releasing here sooner)

2. Are there other features here that should be in the next z-stream release ? (If so, we could get those in as well to avoid needing 2 releases)

If we'd like to consider this for z-stream and we aren't currently aware of other features here that would be desired in the same z-stream, I would say let's go ahead and do it. Otherwise I don't see any problem with updating metadata in devel branches only for now. Let me know what you think @ehelms

ehelms:
> * Do we want to get this into the next z-stream Foreman release ? (If so, I lean towards releasing here sooner)
> *

No. This is a change for the next Foreman release not 2.5.

ekohl:
We can bump the module in git already but not release it. We've done this before. That way you can bump the metadata requirement already. In Voxpupuli it's also common to use a `-rc0` suffix in git to distinguish the two.

ehelms:
PR to just bump the version -- https://github.com/theforeman/puppet-pulpcore/pull/189

HISTORY.md
```diff
@@ -1,3 +1,20 @@
+## [3.3.0](https://github.com/theforeman/puppet-pulpcore/tree/3.3.0) (2021-04-21)
```
wbclark, line 1: the update here is to prevent changelog-generator from wanting to insert empty release notes for v2.2.3

ehelms (approved):

PR #191, opened 2021-05-18
Fixes #32309 - Run admin commands from Pulp's home

ehelms:
* Dynaconf seems to always be using the current directory. As such, the
  cwd could be a directory that is unreadable by the pulp user.  Pulp's
  user home must always be readable by the pulp user.

(cherry picked from commit 951a2f823cf57986623329af54f7d060d256e757)

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup ruby
-        uses: actions/setup-ruby@v1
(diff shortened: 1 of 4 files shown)


wbclark (commented):
I think failing tests would be fixed by https://github.com/theforeman/puppet-pulpcore/pull/192 , or at least it would reduce the amount of noise and allow us to properly test

ekohl (approved):

ehelms:
I think the cherry picks for this branch should include https://github.com/theforeman/puppet-pulpcore/pull/193

PR #192, opened 2021-05-18
Pin camptocamp/systemd to 2.12.0 in testing

wbclark:
Git master has dropped daemon reloading code which we currently depend
on. This pins to the latest released version. I tried to support both,
but currently there's no way to distinguish 2.x and 3.x.

(cherry picked from commit 30d4ecacdc4a98db88e08c2efb5a387b31e7aa53)


wbclark:
I believe this is needed to properly test https://github.com/theforeman/puppet-pulpcore/pull/191

ekohl (approved):

PR #193, opened 2021-05-18
Fixes #32622: Include StdEnvVars, ExportCertData SSL options in Apache

ehelms:
When deploying a stand-alone vhost, Pulp needs these SSL options
set in order for X-CLIENT-CERT to be set.

This fixes an issue for content proxies with Red Hat based content that receive a 403 currently. This should be backported to at least Foreman 2.3.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -123,6 +123,7 @@ class pulpcore::apache (
         ssl_chain         => $pulpcore::apache_https_chain,
         ssl_ca            => $pulpcore::apache_https_ca,
         ssl_verify_client => $ssl_verify_client,
+        ssl_options       => ['+StdEnvVars', '+ExportCertData'],
         *                 => $https_vhost_options,
       }
     }


manifests/apache.pp
```diff
@@ -123,6 +123,7 @@
         ssl_chain         => $pulpcore::apache_https_chain,
         ssl_ca            => $pulpcore::apache_https_ca,
         ssl_verify_client => $ssl_verify_client,
+        ssl_options       => '+StdEnvVars +ExportCertData',
```
wbclark, line 126:
```suggestion
        ssl_options       => ['+StdEnvVars', '+ExportCertData'],
```

Reading https://github.com/puppetlabs/puppetlabs-apache/blob/e14ec095534469f06b417041d7d6967f232eabc5/manifests/vhost.pp#L1640-L1657 , I think this is the correct way to handle multiple options

ekohl (approved):

ekohl:
Btw, how come we didn't see this in our acceptance tests? We deploy there standalone without Foreman's config but do use client certificates. Is our test incomplete?

ehelms:
This requires content being present that is locked behind entitlements and only applies in a stand-alone Pulp scenario (as Foreman's vhost already has this present).

PR #195, opened 2021-05-21
Switch to voxpupuli-acceptance-1.0

ekohl:


spec/acceptance/hieradata/os/RedHat/7.yaml
```diff
@@ -0,0 +1,11 @@
+---
```
ekohl, line 1: This was copied from puppet-foreman

ehelms (approved):

PR #197, opened 2021-05-26
Drop Puppet 5 support

ekohl:

PR #198, opened 2021-05-26
Support camptocamp/systemd 3.x

ekohl:
After [Puppet 5 is dropped](https://github.com/theforeman/puppet-pulpcore/pull/197) camptocamp/systemd 2.x support can be dropped. However, for now I've chosen to support both since I'm not entirely sure if all dependencies in the installer allow 3.x just yet.


ekohl:
Includes https://github.com/theforeman/puppet-pulpcore/pull/199. That PR is to verify it works on camptocamp/systemd 2.x while this PR tests it on 3.x.

PR #199, opened 2021-05-26
Introduce a private socket_service helper

ekohl:
This ensures the correct dependency ordering. This is submitted to camptocamp/systemd but we're currently still on 2.x so no update will be available. This also helps with the move to 3.x.

Submitting this now to see if it works in CI, then I can polish it up.

### DIFF
--- a/manifests/socket_service.pp
+++ b/manifests/socket_service.pp
@@ -1,0 +1,61 @@
+# @summary Create a systemd socket activated service
+# @api private
+#
+# Systemd socket activated services have their own dependencies. This is a
+# convenience wrapper around systemd::unit_file.
+#
+# @param name [Pattern['^[^/]+$']]
+#   The target unit file to create
+# @param socket_content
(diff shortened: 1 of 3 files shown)


manifests/socket_service.pp
```diff
@@ -0,0 +1,57 @@
...
+    assert_type(NotUndef, $socket_content)
+    assert_type(NotUndef, $service_content)
+  }
+
+  $active = $ensure ? {
+    'running' => true,
+    'stopped' => false,
+    'absent'  => false,
+    default   => undef,
+  }
```
ehelms, line 35: I don't see `present` in the list but it is in the Enum of the parameter.

ekohl: That's what `default` is for. I agree it's a bit ugly, but puppet-lint wants a `default` statement and this is how I dealt with it.

ehelms (approved):

PR #200, opened 2021-06-09
Fixes #32766 - adds scheme and pulp content path to setting value

jjeffers:
The current value omits the scheme and pulp content path, which causes issues when syncing content across capsules (and probably will cause other failures too).

### DIFF
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -16,7 +16,7 @@ class pulpcore::plugin::ansible(
   }
 
   pulpcore::plugin { 'ansible':
-    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${pulpcore::apache::external_content_url}\"",
+    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${pulpcore::apache::external_content_url}/\"",
     https_content => epp('pulpcore/apache-fragment.epp', $context),
   }
 }
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -16,7 +16,7 @@ class pulpcore::plugin::ansible(
   }
 
   pulpcore::plugin { 'ansible':
-    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${pulpcore::servername}\"",
+    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"https://${pulpcore::servername}/pulp/content/\"",
     https_content => epp('pulpcore/apache-fragment.epp', $context),
   }
 }
--- a/manifests/plugin/ansible.pp
+++ b/manifests/plugin/ansible.pp
@@ -24,7 +24,7 @@ class pulpcore::plugin::ansible(
   }
 
   pulpcore::plugin { 'ansible':
-    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${external_content_url}/\"",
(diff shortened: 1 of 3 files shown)


manifests/plugin/ansible.pp
```diff
@@ -16,7 +16,7 @@
   }
 
   pulpcore::plugin { 'ansible':
-    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${pulpcore::servername}\"",
+    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"https://${pulpcore::servername}/pulp/content/\"",
```
ekohl, line 19: I don't like constructing a URL again and would prefer to reuse.

We have this here:
https://github.com/theforeman/puppet-pulpcore/blob/22db71ca2c220f52f6651e93d3c60bf19ef0e319/manifests/apache.pp#L19

What we should do is construct the URL. Technically speaking it should also take `$https_port` into account and fail if `https` is turned off. So roughly:

```puppet
if ! $pulpcore::apache_https_vhost {
  # TODO: or can it be served over HTTP?
  fail('HTTPS must be turned on for Ansible content')
} elsif $pulpcore::apache::https_port != 443 {
  # Technically we can't determine this when this module doesn't manage the vhost
  # It's up to the user to keep them in sync
  $external_content_url = "https://${pulpcore::servername}:${pulpcore::apache::https_port}${pulpcore::apache::content_path}"
} else {
  $external_content_url = "https://${pulpcore::servername}${pulpcore::apache::content_path}"
}
```

We're not so complete here:

https://github.com/theforeman/puppet-foreman_proxy_content/blob/fb401944723c5b93941aa98d16f377bded5627c0/manifests/init.pp#L299

It may be an idea to add a `$external_content_url` in `apache.pp` so that it can be used in puppet-foreman_proxy_content too.

ekohl (changes requested):

spec/classes/plugin_ansible_spec.rb
```diff
@@ -25,7 +25,7 @@
                            .with_content(expected_vhost_content)
         is_expected.to contain_concat__fragment('plugin-ansible')
           .with_content(/^ANSIBLE_API_HOSTNAME = "foo.example.com"/)
-          .with_content(/^ANSIBLE_CONTENT_HOSTNAME = "foo.example.com"/)
+          .with_content(%r{^ANSIBLE_CONTENT_HOSTNAME = "https://foo.example.com/pulp/content/"})
```
wbclark, line 28: The percent literal is a lot nicer here. :+1:

manifests/plugin/ansible.pp
wbclark: > It may be an idea to add a `$external_content_url` in `apache.pp` so that it can be used in puppet-foreman_proxy_content too.

I like this idea, and I think it's best that we tackle the change in apache.pp it in a separate PR.

We could fix the immediate issue here, then improve the cross-module implementation details in a PR that is "Refs #32766 - ..." (really two separate PRs, since we'd also want to update puppet-FPC at that point)

manifests/plugin/ansible.pp
```diff
@@ -15,8 +15,16 @@
...
+    fail('HTTPS must be turned on for Ansible content')
+  } elsif $pulpcore::apache::https_port != 443 {
+    $external_content_url = "https://${pulpcore::servername}:${pulpcore::apache::https_port}${pulpcore::apache::content_path}"
+  } else {
+    $external_content_url = "https://${pulpcore::servername}${pulpcore::apache::content_path}"
+  }
+
   pulpcore::plugin { 'ansible':
-    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${pulpcore::servername}\"",
+    config        => "ANSIBLE_API_HOSTNAME = \"${pulpcore::servername}\"\nANSIBLE_CONTENT_HOSTNAME = \"${external_content_url}/\"",
```
ekohl, line 27: Just wondering: does it need a trailing slash?

ekohl (approved):
Minor question inline, but I'm ok with it.

manifests/plugin/ansible.pp
jjeffers: I assumed it was needed. However, I just tested this without a trailing slash. It appears that pulp3 handles this without the trailing slash without any issues.

ekohl (approved):

PR #201, opened 2021-06-14
Support version 3.11

ekohl:
This version is already being shipped by Katello 4.1 and should be supported.


ekohl:
It fails because 2to3 is incompatible with the Django version. I'm surprised repoclosure didn't catch this but it's probably a limitation of RPM.

ekohl:
Not for us, but do we have a way to uninstall it for users?

evgeni:
Uh, no, probably not. Good point.

ekohl:
https://github.com/pulp/pulp-2to3-migration/pull/381 should unpin it again.

ehelms (approved):

ekohl:
@evgeni do you think we can include such a PR in the update procedure? https://github.com/theforeman/tool_belt/blob/master/procedures/pulpcore/update.md.erb does have a `Update to a new pulpcore release` section.

PR #202, opened 2021-06-17
Support Pulp 3.13, drop earlier versions

ekohl:
Pulp 3.13 moves to a new wrapper for the worker. This isn't present in older releases so support for them is dropped.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,7 @@ jobs:
         setfile: ${{fromJson(needs.setup_matrix.outputs.beaker_setfiles)}}
         puppet: ${{fromJson(needs.setup_matrix.outputs.puppet_major_versions)}}
         pulpcore_version:
-          - '3.9'
-          - '3.11'
+          - '3.13'
     name: Acceptance / ${{ matrix.puppet.name }} - ${{ matrix.setfile.name }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


evgeni:
Which just happened. I re-triggered the jobs.

ekohl:
To be clear, we should do a minor release before we bump to a major.

evgeni:
ah, and there is no compatoble 2to3 migration for 3.13 yet.

given it's not used by katello anymore, shall we drop it?

ekohl:
Won't that break upgrades for users? Recently we had a chat with Tanya on IRC and she said it's being worked on.

ehelms:
Does 3.13 drop the RQ worker? Or are both provided as options?

ekohl:
I think the worker script is a wrapper that can call both, depending on the settings. Technically we can introduce a simple shell wrappper in pulpcore 3.11 RPM to remain compatible, but I'm not sure we should take that route.

evgeni:
Yeah, I think RQ is still the default, but you can make it use something else.

ehelms:
Do these new workers require any updates to Pulp's selinux policy?

ekohl:
I don't think they do since this covers both the old and new wrapper.
https://github.com/pulp/pulpcore-selinux/blob/064442e829310e6bab95aa639343a8cfaf90f219/pulpcore.fc#L11

evgeni:
That's my understanding as well.

I've opened https://github.com/theforeman/pulpcore-packaging/pull/180 to get in the new 2to3 migration plugin, which is compatible with 3.13 in, which should make the tests happy.

ekohl:
This is now green.

ehelms (approved):

PR #203, opened 2021-06-27
Fixes #32891 - Add feature to enable new tasking system and enable it by default

wbclark:
This introduces a new parameter `use_legacy_tasking_system` which
defaults to true and configures the application the same as before.
When set to false, pulpcore is configured with the new tasking system
instead. Acceptance tests are included to ensure that users can switch
between the two tasking systems. This is backwards compatible with
Pulpcore versions older than 3.13 ONLY when configured to use the
legacy tasking system. The newer tasking system requires Pulpcore
version >= 3.13.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -132,6 +132,11 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param use_rq_tasking_system
+#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
+#   Any benchmarking you did to optimize worker_count or other tasking related parameters will no longer be accurate after changing the tasking system.
+#   Do not modify this setting unless you understand the implications for performance and stability.
+#
 # @param service_enable
 #   Enable/disable Pulp services at boot.
 #
@@ -203,6 +208,7 @@ class pulpcore (
   Pulpcore::ChecksumTypes $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
+  Boolean $use_rq_tasking_system = false,
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -18,10 +18,18 @@ class pulpcore::service {
     service_content => template('pulpcore/pulpcore-content.service.erb'),
   }
 
-  systemd::unit_file { 'pulpcore-resource-manager.service':
-    content => template('pulpcore/pulpcore-resource-manager.service.erb'),
-    active  => $pulpcore::service_ensure,
-    enable  => $pulpcore::service_enable,
+  if $pulpcore::use_rq_tasking_system {
+    systemd::unit_file { 'pulpcore-resource-manager.service':
+      content => template('pulpcore/pulpcore-resource-manager.service.erb'),
+      active  => $pulpcore::service_ensure,
+      enable  => $pulpcore::service_enable,
+    }
+  } else {
+    systemd::unit_file { 'pulpcore-resource-manager.service':
+      ensure => 'absent',
+      active => false,
+      enable => false,
+    }
   }
 
   systemd::unit_file { 'pulpcore-worker@.service':
--- a/spec/acceptance/disable_new_tasking_system_spec.rb
+++ b/spec/acceptance/disable_new_tasking_system_spec.rb
@@ -1,0 +1,137 @@
+require 'spec_helper_acceptance'
+
+describe 'change configuration from postgresql tasking system to rq tasking system', :order => :defined do
+  certdir = '/etc/pulpcore-certs'
+
+  context 'initial configuration with newer postgresql tasking system' do
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) do
+        <<-PUPPET
+        class { 'pulpcore':
+          worker_count          => 1,
+          use_rq_tasking_system => false,
+        }
+        PUPPET
+      end
+    end
+  
+    describe service('httpd') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-api') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-content') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe file('/etc/systemd/system/pulpcore-resource-manager.service') do
+      it { is_expected.not_to exist }
+    end
+  
+    describe service('pulpcore-resource-manager') do
+      it { is_expected.not_to be_enabled }
+      it { is_expected.not_to be_running }
+    end
+  
+    describe service('pulpcore-worker@1') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe port(80) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe port(443) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+      its(:response_code) { is_expected.to eq(200) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
+      its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = True/) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+      its(:stdout) { is_expected.to match(/^0 workers, /) }
+      its(:stdout) { is_expected.not_to match(/^resource-manager /) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  end
+  
+  context 'reconfigure pulpcore to use older rq worker tasking system' do
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) do
+        <<-PUPPET
+        class { 'pulpcore':
+          worker_count          => 1,
+          use_rq_tasking_system => true,
+        }
+        PUPPET
+      end
+    end
+  
+    describe service('httpd') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-api') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-content') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe file('/etc/systemd/system/pulpcore-resource-manager.service') do
+      it { is_expected.to exist }
+    end
+  
+    describe service('pulpcore-resource-manager') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-worker@1') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe port(80) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe port(443) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+      its(:response_code) { is_expected.to eq(200) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
+      its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = False/) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+      its(:stdout) { is_expected.to match(/^2 workers, /) }
+      its(:stdout) { is_expected.to match(/^resource-manager /) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  end
+end
--- a/spec/acceptance/enable_new_tasking_system_spec.rb
+++ b/spec/acceptance/enable_new_tasking_system_spec.rb
@@ -1,0 +1,137 @@
+require 'spec_helper_acceptance'
+
+describe 'change configuration from rq tasking system to postgresql tasking system', :order => :defined do
+  certdir = '/etc/pulpcore-certs'
+
+  context 'initial configuration with older rq worker tasking system' do
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) do
+        <<-PUPPET
+        class { 'pulpcore':
+          worker_count          => 1,
+          use_rq_tasking_system => true,
+        }
+        PUPPET
+      end
+    end
+  
+    describe service('httpd') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-api') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-content') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe file('/etc/systemd/system/pulpcore-resource-manager.service') do
+      it { is_expected.to exist }
+    end
+  
+    describe service('pulpcore-resource-manager') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-worker@1') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe port(80) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe port(443) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+      its(:response_code) { is_expected.to eq(200) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
+      its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = False/) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+      its(:stdout) { is_expected.to match(/^2 workers, /) }
+      its(:stdout) { is_expected.to match(/^resource-manager /) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  end
+  
+  context 'reconfigure pulpcore to use newer postgresql tasking system' do
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) do
+        <<-PUPPET
+        class { 'pulpcore':
+          worker_count          => 1,
+          use_rq_tasking_system => false,
+        }
+        PUPPET
+      end
+    end
+  
+    describe service('httpd') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-api') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe service('pulpcore-content') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe file('/etc/systemd/system/pulpcore-resource-manager.service') do
+      it { is_expected.not_to exist }
+    end
+  
+    describe service('pulpcore-resource-manager') do
+      it { is_expected.not_to be_enabled }
+      it { is_expected.not_to be_running }
+    end
+  
+    describe service('pulpcore-worker@1') do
+      it { is_expected.to be_enabled }
+      it { is_expected.to be_running }
+    end
+  
+    describe port(80) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe port(443) do
+      it { is_expected.to be_listening }
+    end
+  
+    describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+      its(:response_code) { is_expected.to eq(200) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
+      its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = True/) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  
+    describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+      its(:stdout) { is_expected.to match(/^0 workers, /) }
+      its(:stdout) { is_expected.not_to match(/^resource-manager /) }
+      its(:exit_status) { is_expected.to eq 0 }
+    end
+  end
+end
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -23,6 +23,8 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 
+USE_NEW_WORKER_TYPE = <%= scope['pulpcore::use_rq_tasking_system'] ? "False" : "True" %>
+
 MEDIA_ROOT = "<%= scope['pulpcore::media_root'] %>"
 STATIC_ROOT = "<%= scope['pulpcore::static_root'] %>"
 STATIC_URL = "<%= scope['pulpcore::static_url'] %>"
--- a/spec/acceptance/basic_spec.rb
+++ b/spec/acceptance/basic_spec.rb
@@ -29,8 +29,8 @@ describe 'basic installation' do
   end
 
   describe service('pulpcore-resource-manager') do
-    it { is_expected.to be_enabled }
-    it { is_expected.to be_running }
+    it { is_expected.not_to be_enabled }
+    it { is_expected.not_to be_running }
   end
 
   describe service('pulpcore-worker@1') do
@@ -73,6 +73,17 @@ describe 'basic installation' do
     its(:body) { is_expected.to contain('artifacts_list') }
     its(:exit_status) { is_expected.to eq 0 }
   end
+
+  describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
+    its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = True/) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+
+  describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+    its(:stdout) { is_expected.to match(/^0 workers, /) }
+    its(:stdout) { is_expected.not_to match(/^resource-manager /) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
 end
 
 describe 'reducing worker count' do
@@ -102,8 +113,8 @@ describe 'reducing worker count' do
   end
 
   describe service('pulpcore-resource-manager') do
-    it { is_expected.to be_enabled }
-    it { is_expected.to be_running }
+    it { is_expected.not_to be_enabled }
+    it { is_expected.not_to be_running }
   end
 
   describe service('pulpcore-worker@1') do
--- a/spec/acceptance/plugins_spec.rb
+++ b/spec/acceptance/plugins_spec.rb
@@ -37,8 +37,8 @@ describe 'basic installation' do
   end
 
   describe service('pulpcore-resource-manager') do
-    it { is_expected.to be_enabled }
-    it { is_expected.to be_running }
+    it { is_expected.not_to be_enabled }
+    it { is_expected.not_to be_running }
   end
 
   describe port(80) do
--- a/spec/acceptance/use_pulp2_content_route_spec.rb
+++ b/spec/acceptance/use_pulp2_content_route_spec.rb
@@ -40,8 +40,8 @@ describe 'basic installation' do
   end
 
   describe service('pulpcore-resource-manager') do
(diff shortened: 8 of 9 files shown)


wbclark:
NOTE: While the default (`use_legacy_tasking_system = true`) should be fully backwards compatible and consistent with all behavior prior to this commit, this also introduces tests to enable and disable the new tasking system, and those will FAIL until pulpcore >= 3.13 is available in foreman-packaging.

spec/acceptance/disable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+  describe port(443) do
+    it { is_expected.to be_listening }
+  end
+
+  describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+    its(:response_code) { is_expected.to eq(200) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+
+  describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
```
wbclark, line 55: This is somewhat of a hack. `diffsettings` will show the difference between default settings for Django, so pulpcore's `USE_NEW_WORKER_TYPE` setting (introduced with v3.13) should be visible here.

spec/acceptance/disable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+  describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+    its(:stdout) { is_expected.to match(/^0 workers, /) }
+    its(:stdout) { is_expected.not_to match(/^resource-manager /) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+
+  describe command("ps -e -o command") do
+    its(:stdout) { is_expected.to match(%r(^/usr/libexec/pulpcore/pulpcore-worker$)) }
+    its(:stdout) { is_expected.to match(%r(^/usr/libexec/pulpcore/pulpcore-worker --resource-manager$)) }
```
wbclark, line 68: This is consistent with the `ExecStart` configuration in the unit files but may not display as such in `ps` output; depending on how this is handled in packaging it may be something more like `/usr/bin/python3 /path/to/new/pulpcore-worker`

templates/settings.py.erb
```diff
@@ -40,6 +40,15 @@ REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
 ALLOWED_IMPORT_PATHS = <%= scope['pulpcore::allowed_import_path'] %>
 ALLOWED_EXPORT_PATHS = <%= scope['pulpcore::allowed_export_path'] %>
 ALLOWED_CONTENT_CHECKSUMS = <%= scope['pulpcore::allowed_content_checksums'] %>
+<% if scope['pulpcore::use_legacy_tasking_system'] %>
+
+# This setting is necessary but not sufficient to use the new tasking system;
+# the systemd unit files must also be properly configured for either worker type.
+# As such you should NEVER directly modify this setting. Refer to the documentation
+# (from your software vendor, if any, otherwise to the community documentation) for
+# the correct procedure to change between the new and legacy worker types.
```
wbclark, line 49: I struggled with writing this comment. I want to issue a warning to users who happen upon the `settings.py` not to modify this setting directly, as updating the setting but not the unit files will result in a broken configuration. But finding the correct way to make the configuration change would be quite different for an upstream user who used the puppet modules directly vs. an upstream user who used foreman-installer vs. an user of a downstream product like Satellite.

ekohl:
Should this be rebased on https://github.com/theforeman/puppet-pulpcore/pull/202?

manifests/init.pp
```diff
@@ -197,6 +202,7 @@
   Pulpcore::ChecksumTypes $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
+  Boolean $use_legacy_tasking_system = true,
```
ehelms, line 205: To avoid the ole `legacy` wording, perhaps we name it `use_postgresql_tasking_system` or `use_rq_tasking_system.

templates/settings.py.erb
ehelms: I suppose warning them is extra helpful. This is a file controlled by Puppet, they should not change *anything* in this file directly. That's a general rule of all things under our control.

ehelms: We should look to add a header to this file if we do not already that it's under Puppet control?

spec/acceptance/disable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+
+  describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
+    its(:stdout) { is_expected.to match(/^0 workers, /) }
+    its(:stdout) { is_expected.not_to match(/^resource-manager /) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+
+  describe command("ps -e -o command") do
```
ekohl, line 66: Would [process](https://serverspec.org/resource_types.html#process) be a better type to use?

templates/pulpcore-resource-manager.service.erb
```diff
@@ -11,10 +11,14 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=<%= scope['pulpcore::user_home'] %>
 RuntimeDirectory=pulpcore-resource-manager
+<% if scope['pulpcore::use_legacy_tasking_system'] %>
```
ekohl, line 14: https://github.com/theforeman/puppet-pulpcore/pull/202 already takes care of this. Also, if you're using lines like this, you should strip trailing whitespace to avoid empty lines
```suggestion
<% if scope['pulpcore::use_legacy_tasking_system'] -%>
```

templates/settings.py.erb
ekohl: With https://github.com/theforeman/puppet-pulpcore/pull/202, is it really needed? The systemd unit will then simply always use the new wrapper that can do both.

manifests/init.pp
evgeni: or an Enum of `['rq', 'postgresql']` or `['rq', 'new']`?

If I understand https://github.com/pulp/pulpcore/commit/0e1c6292f6601be4ea93417d688bb929b8a88212 correctly, there are two changes in that commit:
1. add a script ("entrypoint" in Python lingo) `pulpcore-worker` which should be used instead of `rq`
2. allow for two different tasking systems: old RQ and new PostgreSQL based

So switching to the old one doesn't *require* us to switch to calling `rq` directly (like this PR does right now) and I think this is actually officially unsupported by Pulpcore (or *should* be unsupported)

And if you look at the commit in `pulp_installer` (https://github.com/pulp/pulp_installer/commit/39a886965dca3cc30778329d7b1ccfbd32bb7b18) they switch to the new binary for all 3.13+ installs, unrelated to the used tasking system

spec/acceptance/disable_new_tasking_system_spec.rb
wbclark: The part I struggle with is this bit: "When several processes match, only the parameters of the first one are available" so this could return the parameters of any worker or even the resource manager -- as the new pulpcore-worker and resource manager share the same executable name and the resource manager is differentiated by appending the `--resource-manager` argument.

That may not be a big deal, as the new "resource manager" as I understand it is a dummy process which does nothing, is not required for the new tasking system to function, and seems to exist to allow switching between the two tasking systems without breaking tools like foreman-maintain that expect to find a `pulpcore-resource-manager.service`

The way I have it currently written tests that at least one resource manager and at least one non resource manager worker are running. I think I wouldn't be able to do that with https://serverspec.org/resource_types.html#process

templates/settings.py.erb
wbclark: > I suppose warning them is extra helpful. This is a file controlled by Puppet, they should not change _anything_ in this file directly. That's a general rule of all things under our control.

I am concerned about users falling into the trap of "yes, I know this is managed by the installer, but I don't care about making it persistent, I just want to try out the new workers so I'll change the configuration in settings.py and restart services."

I've seen both end users and support engineers use this type of approach, particularly when troubleshooting (for example to make a temporary change to the # of pulp 2 workers, worker timeout, or download concurrency)

wbclark: > With #202, is it really needed? The systemd unit will then simply always use the new wrapper that can do both.

I think you're right.

I initially misread https://github.com/pulp/pulp_installer/pull/616 / missed https://pulp.plan.io/issues/8721

spec/acceptance/disable_new_tasking_system_spec.rb
wbclark: Looking back at this with a better understanding of https://pulp.plan.io/issues/8721 , this test might be entirely redundant anyway.

ehelms:
To recap, the first big question is whether we take a progressive enhancement, keep supporting Pulp 3.11 and 3.13+ or just support 3.13+ only with the new wrapper. https://github.com/theforeman/puppet-pulpcore/pull/202 takes the latter approach.

I lean towards the 3.13+ only approach, there are enough changes coming that I think its safe to cut off 3.11 from the main branch, and create a stable release branch if we need to backport to supporting 3.11 for any reason. The big changes, supporting new tasking system, content caching are only applicable to 3.13+ and there is no plan to backport any of those so I think its safe for our module to look forward only.

wbclark:
@ehelms @ekohl @evgeni thanks for all comments and feedback!

I've updated this with the following changes:
- rebased on https://github.com/theforeman/puppet-pulpcore/pull/202 (support only pulpcore >= 3.13, always use new worker entrypoint)
- renamed the parameter to `use_rq_tasking_system` (kept this a boolean instead of moving to an enum)
- removed a redundant test of `ps -e -o command` output from the new acceptance tests (since always using new worker entrypoint)
- removed the new inaccurate comment from settings.py
- always include `USE_NEW_WORKER_TYPE` in settings.py (don't rely on default) and set its value based on `use_rq_tasking_system`

ekohl:
> To recap, the first big question is whether we take a progressive enhancement, keep supporting Pulp 3.11 and 3.13+ or just support 3.13+ only with the new wrapper. #202 takes the latter approach.

Just to throw it out there, we could introduce the same wrapper in 3.11 if we really cared about this.

manifests/init.pp
wbclark: I went with `use_rq_tasking_system` for now but I'm open to other ideas.

Some thoughts: at a base level it would be nice for this module to be consistent with the setting name in pulpcore, but `USE_NEW_WORKER_TYPE` has a problem in that the postgresql tasking system won't be "new" forever, and something that is described as "new" could be perceived by users as being unstable, unsafe, etc.

I like being specific and descriptive about what technology is actually used, so `Enum['rq', 'postgresql']` is appealing for that reason, but that could be more difficult for `foreman-installer` users ("what were the options again? I need to find that in the docs somewhere") vs. a simple boolean which can be set true or false.

So I went with `use_rq_tasking_system`, added a note in the parameter documentation that this refers to the older rq workers tasking system and the alternative is the postgresql tasking system introduced in pulpcore 3.13.

When we later reach a point where `use_rq_tasking_system` is false by default, this seems particularly intuitive as an interface to allow users to opt-in to use the older tasking system instead.

evgeni:
> > To recap, the first big question is whether we take a progressive enhancement, keep supporting Pulp 3.11 and 3.13+ or just support 3.13+ only with the new wrapper. #202 takes the latter approach.
> 
> Just to throw it out there, we could introduce the same wrapper in 3.11 if we really cared about this.

Given the current plans around pulpcore upgrades, I don't think we need to invest in 3.11 compatibility too long anymore?

manifests/init.pp
```diff
@@ -132,6 +132,11 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param use_rq_tasking_system
+#   Use the older rq workers tasking system instead of the newer postgresql tasking system introduced in Pulpcore 3.13.
```
evgeni, line 136:
```suggestion
#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.13.
```

evgeni:
A question on how this will play out when we decide to flip the switch and use the new (PostgreSQL) system:

Right now this is a direct parameter to the class (`pulpcore::use_rq_tasking_system`) and thus will be recorded in the answers file as `true`.
Once we decide to flip the switch, we'll edit `init.pp` and set `use_rq_tasking_system` to `false`.
Given the class parameters are overridden by answers, users that had it set to `true` before, will remain on the RQ system.
When we now introduce a migration, it would migrate all users from `true` to `false`, regardless whether they have set this manually or because it's the default. Wouldn't this migration thus make it impossible to flip back to RQ, as every time the installer is run, the value gets migrated?

ekohl:
> > > To recap, the first big question is whether we take a progressive enhancement, keep supporting Pulp 3.11 and 3.13+ or just support 3.13+ only with the new wrapper. #202 takes the latter approach.
> > 
> > 
> > Just to throw it out there, we could introduce the same wrapper in 3.11 if we really cared about this.
> 
> Given the current plans around pulpcore upgrades, I don't think we need to invest in 3.11 compatibility too long anymore?

The case I can imagine is that we have update the pulpcore repo definition in katello-repos.rpm. So the user runs:
* yum update
* foreman-installer

Now what the user should have done:
* yum update katello-repos
* yum update
* foreman-installer

By shipping the wrapper in 3.11, at least the service will continue to start up.

ekohl:
> Wouldn't this migration thus make it impossible to flip back to RQ, as every time the installer is run, the value gets migrated?

Migrations only run once. There's a file (IIRC `.applied` in `$scenario.migrations`) that lists names of migrations that already ran.

evgeni:
> The case I can imagine is that we have update the pulpcore repo definition in katello-repos.rpm. So the user runs:
> 
>     * yum update
>     * foreman-installer

Ooooh, you mean for users already on 4.1 that is using 3.11? Fair. But at the same time, this will only happen once?

ekohl:
> Ooooh, you mean for users already on 4.1 that is using 3.11? Fair. But at the same time, this will only happen once?

Yes, but it's certainly a case I worry about. We should very carefully approach it to make it a smooth and painless migration.

ehelms:
You can now rebase this on master

wbclark:
> You can now rebase this on master

Rebased

templates/settings.py.erb
```diff
@@ -23,6 +23,8 @@ REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 
+USE_NEW_WORKER_TYPE=<%= scope['pulpcore::use_rq_tasking_system'] ? "False" : "True" %>
```
ehelms, line 26:
```suggestion
USE_NEW_WORKER_TYPE = <%= scope['pulpcore::use_rq_tasking_system'] ? "False" : "True" %>
```

Style nit.

ehelms:
Code looks good, can you fix up the tests?

ekohl (commented):
A discussion we had on IRC: should we only include Redis if RQ is used? That will mean we don't have to deploy Redis on content proxies. Fewer services is certainly better for users.

IMHO we don't need to clean up an existing installation.

ehelms:
> A discussion we had on IRC: should we only include Redis if RQ is used? That will mean we don't have to deploy Redis on content proxies. Fewer services is certainly better for users.
> 
> IMHO we don't need to clean up an existing installation.

I would agree with that simplicity of it. If new tasking system, which would be the default, do not set up Redis. If switching to old system, setup Redis. If switching from old to new, leave Redis alone.

wbclark:
> Code looks good, can you fix up the tests?

Recapping IRC discussion:

This necessarily must wait until pulpcore 3.14 is installable. version 3.13 has the new worker entrypoints but doesn't actually contain the new tasking system.

Currently:

```
# rpm -q python3-pulpcore
python3-pulpcore-3.13.0-2.el7.noarch
```

ehelms:
This can now be rebased on master and see if tests are successful.

spec/acceptance/disable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+
+describe 'initial configuration with newer postgresql tasking system' do
+  certdir = '/etc/pulpcore-certs'
+
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      class { 'pulpcore':
+        worker_count          => 1,
+        use_rq_tasking_system => false,
```
ehelms, line 11: Can drop as this is the default

spec/acceptance/disable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe service('pulpcore-content') do
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe service('pulpcore-resource-manager') do
```
ehelms, line 32: I thought this service went away with the new system

wbclark: the new entrypoint `pulpcore-worker --resource-manager` points to the old resource manager when the app is configured for it, else it is basically a null command

wbclark: I've been waiting to see how this behaves once the new tasking system is available, to see if tests need updating

ehelms: I see. I think I am OK with that behavior for the first pass of this. We should clean that up later so the service is gone in the new tasking system.

ekohl: @mdellweg @bmbouter can you enlighten us?

bmbouter: This service is no longer needed at all with the new tasking system. Starting it will not error and it will just sleep. At some point starting it will yield an error.

spec/acceptance/disable_new_tasking_system_spec.rb
wbclark: I think dropping it ensures that we are properly testing the default behavior, but at the expense of making the tests less readable.

I'd rather test default behavior by adding that test to the `basic_spec.rb`

spec/acceptance/disable_new_tasking_system_spec.rb
ekohl: Thanks! Given that we've declared the module 3.14 only, we can ensure it's present & running if RQ is used and absent & stopped when new tasking is used. This will save some resources but more importantly, it's easier to understand for users.

wbclark: > Thanks! Given that we've declared the module 3.14 only, we can ensure it's present & running if RQ is used and absent & stopped when new tasking is used. This will save some resources but more importantly, it's easier to understand for users.

By removing the service under those circumstances, we'd be creating an issue in `foreman-maintain service` and in any other scripts that check service status. I'm appreciative of the grace period we've been given by the current transitionary design and I think it would be beneficial to proceed carefully here.

ekohl: You can then just ensure the service is present but stopped. I think it should deal with disabled services now. Also, foreman-maintain shouldn't hold us back and if it can't deal with it, we should fix it.

wbclark:
Updated with style nits.

@ehelms I kept explicit declaration of `use_rq_tasking_system` in all cases of the enable/disable tests, and I added a test that the postgresql tasking system is used by default without declaration in the basic_spec.rb -- please let me know if that's agreeable.

spec/acceptance/disable_new_tasking_system_spec.rb
ehelms: I agree foreman-maintain should handle it, the question that I think is a good one being raised by @wbclark is the radius of affects for this first change. So we have two options:

 Option 1:
   * Leave resource manager alone
   * Update foreman-maintain and any other code
   * Add another patch that better manages resource-manager
  
Option 2:
 * Handle resource-manager from the start
 * Fix up foreman-maintain and other areas
 
I do agree that resource-manager being present and running presents the biggest over head to debugging as it may lead folks down incorrect paths when investigating. That is something I do think is important to avoid where we can given the newness of this.

wbclark: It's been mentioned that there is a resource overhead for having the extraneous service, but my understanding is that it should be quite negligible.

It's been mentioned as well that there is an overhead to debugging as the presence of the resource manager could lead people down the wrong path in troubleshooting. I would argue that the _absence_ of the resource manager could have the very same effect.

For me, it comes down to two things: 1. the radius of effects, as @ehelms called it, and 2. what I read to be the Pulp team's design intent in implementing an idle "resource manager" under the PostgreSQL tasking system (I infer they made this decision for essentially the same reasons as the 1st point, but applied to an even wider set of users).

So my strong preference, given the context in which we are transitioning to pulpcore 3.14, is to minimize the risk and take advantage of the opportunity to easily switch between tasking systems without having to manage a different set of services.

wbclark: > Update foreman-maintain and any other code

It's the "and any other code" part here that really concerns me. Maybe that's no code at all, maybe that's end user scripts and in house monitoring tools that we aren't responsible for, or maybe there are other things which we're responsible for which could break. I'm not 100% confident about it, and that's the reason I'd prefer to defer the change.

ehelms (approved):
I'd like @ekohl to also weigh in. For me, while I like the clean approach, this feels like a slightly easier ecosystem transition for the moment.

wbclark:
> I'd like @ekohl to also weigh in.

:+1:

I appreciate all of the feedback from the team, and particularly so when it challenges my design decisions.

I hope I've been persuasive in defending my intent around the pulpcore-resource-manager.service , and that we can move forward with the support for switching between tasking systems as currently written and make the change in that area at a slightly later time, but let's give a chance for additional feedback and hopefully find a consensus before merging.

wbclark:
Updated commit message to reflect the current implementation

spec/acceptance/disable_new_tasking_system_spec.rb
evgeni: I lean towards saying: look, we changed the tasking system, if your local custom code broke by that, fix it now, please.

Pretending the manager is still there would just break it later and for now give the users the false feeling of their old code still working (it is not, it is looking at a dummy).

I *think* f-maintain should be able to handle this situation today, and if not we can fix it. Any other code we shouldn't care about breaking, not at a point like this were we are actively replacing big parts of the stack.

spec/acceptance/disable_new_tasking_system_spec.rb
evgeni: Is there a reason not to just assert the contents of `settings.py`?

spec/acceptance/disable_new_tasking_system_spec.rb
ekohl: Given that this is the subject of the test, I don't mind being explicit here.

spec/acceptance/disable_new_tasking_system_spec.rb
ekohl: I lean the same way as @evgeni. We're going to have to deal with it at some point anyway and as a user I would be very unhappy with a service that does nothing. I also think that the customer support aspect should weigh heavy. In theory it can fail and does generate output so it means people will check it. That's only wasting time.

spec/acceptance/disable_new_tasking_system_spec.rb
ekohl: I think you can also use dynaconf to dump the settings. I think that won't break when Pulp switches its default.

ekohl (changes requested):
We discussed ensuring pulpcore-resource-manager is absent when using the new tasking system and agreed that's best.

wbclark:
We discussed with @upadhyeammit and settled on disabling the pulpcore-resource-manager.service and removing the unit file when not using the RQ tasking system.

foreman_maintain will then use the presence or absence of the unit file to determine whether the pulpcore-resource-manager.service should be included in its services definition.

wbclark:
Current implementation is dealing with this issue when enabling the new tasking system:

```
# systemctl status pulpcore-resource-manager.service
● pulpcore-resource-manager.service
   Loaded: not-found (Reason: Unit pulpcore-resource-manager.service not found.)
   Active: active (running) since Tue 2021-07-06 18:32:08 UTC; 3min 56s ago
 Main PID: 26471 (pulpcore-worker)
    Tasks: 2 (limit: 5970)
   Memory: 46.2M
   CGroup: /system.slice/pulpcore-resource-manager.service
           └─26471 /usr/libexec/platform-python /usr/bin/pulpcore-worker --resource-manager

Jul 06 18:32:08 centos8-64-1 systemd[1]: Started Pulp Resource Manager.
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: pulpcore.tasking.entrypoint:INFO: Starting rq type worker
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: pulpcore.tasking.worker_watcher:INFO: Worker 'resource-manager' shutdown
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: pulpcore.tasking.worker_watcher:INFO: Cleaning up shutdown worker 'resource-manager'.
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: rq.worker:INFO: Worker rq:worker:resource-manager: started, version 1.8.1
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: rq.worker:INFO: Subscribing to channel rq:pubsub:resource-manager
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: rq.worker:INFO: *** Listening on resource-manager...
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: pulpcore.tasking.worker_watcher:INFO: New worker 'resource-manager' discovered
Jul 06 18:32:12 centos8-64-1 pulpcore-resource-manager[26471]: pulp [None]: rq.worker:INFO: Cleaning registries for queue: resource-manager
Jul 06 18:32:54 centos8-64-1 systemd[1]: pulpcore-resource-manager.service: Current command vanished from the unit file, execution of the command list won't be resumed.
```

manifests/service.pp
```diff
@@ -18,10 +18,16 @@
...
-    enable  => $pulpcore::service_enable,
+  if $pulpcore::use_rq_tasking_system {
+    systemd::unit_file { 'pulpcore-resource-manager.service':
+      content => template('pulpcore/pulpcore-resource-manager.service.erb'),
+      active  => $pulpcore::service_ensure,
+      enable  => $pulpcore::service_enable,
+    }
+  } else {
+    systemd::unit_file { 'pulpcore-resource-manager.service':
+      ensure => 'absent',
```
ekohl, line 29: I think you should also add `active` and `enable` to make sure they're stopped.

ekohl (changes requested):

manifests/service.pp
wbclark: Thanks!

I think I initially misread https://github.com/voxpupuli/puppet-systemd/blob/97dd16fa32886b5b0f77a6f38a4953d4c1511c0e/manifests/unit_file.pp#L90-L115 as implying that `ensure => absent` was sufficient

spec/acceptance/disable_new_tasking_system_spec.rb
wbclark: > Is there a reason not to just assert the contents of `settings.py`?

To me, reading the running configuration of the application seems closer to testing the true, desired outcome vs. reading the contents of a file. For that reason I see it as a better approach in an "acceptance" test.

With that said, we could easily test both and treat one as an in memory test for the currently running process and the other as a test of the persistent configuration.

wbclark: > I think you can also use dynaconf to dump the settings. I think that won't break when Pulp switches its default.

I'll have to look into how to do this

spec/acceptance/disable_new_tasking_system_spec.rb
wbclark: > Given that this is the subject of the test, I don't mind being explicit here.

This sums up my feeling as well.

wbclark: `spec/acceptance/basic_spec.rb` now tests that the PostgreSQL tasking system is used with default parameters.

wbclark:
To my knowledge, this is now ready to go, as it provides the agreed upon design and has complete test coverage.

Requesting a fresh round of reviews. Thanks for all the feedback and testing so far.

spec/acceptance/disable_new_tasking_system_spec.rb
ehelms: This test looks good enough for now. Or just yank it since we test this in the unit tests and we are asserting the behaviors we expect anyway. That's my TLDR way of saying, let's not hold this up on that sorta detail.

templates/settings.py.erb
```diff
@@ -1,3 +1,11 @@
+########################################################################################
+# WARNING: This file is externally managed and you should never make manual edits.     #
+#                                                                                      #
+# Not only are your edits likely to be overwritten, there is also a strong possibility #
+# of breaking your system. Refer to documentation for the software that installed      #
+# Pulpcore to determine the safe and persistent way to modify the configuration.       #
+########################################################################################
```
ehelms, line 7: If we are going to solve this problem *now* (as this is a change that is nice but not needed by the target of this PR and could be addressed later) could we stick to the Foreman header for consistency? https://github.com/theforeman/puppet-foreman/blob/master/templates/_header.erb

spec/acceptance/disable_new_tasking_system_spec.rb
ekohl: > I'll have to look into how to do this

FYI: sosreport has this:
https://github.com/sosreport/sos/blob/c7d3644c0c64e9e5439806250592a55c8e2de26f/sos/report/plugins/pulpcore.py#L99-L102

spec/acceptance/enable_new_tasking_system_spec.rb
```diff
@@ -0,0 +1,137 @@
+require 'spec_helper_acceptance'
+
+describe 'initial configuration with older rq worker tasking system' do
```
ekohl, line 3: rspec guidelines say you should only have a single top level `describe` while this file has 2. Maybe you also need to add `order: :defined` so it really runs in the order you expect.

templates/settings.py.erb
ekohl: I think it would make a good separate PR. There's no reason it really needs to be here in this PR.

ekohl (changes requested):

templates/settings.py.erb
wbclark: I am wary of removing the unit file for pulpcore-resource-manager.service without including this warning in settings.py

wbclark: Specifically that change invites the possibility of someone modifying the `USE_NEW_WORKER_TYPE` setting without having the necessary service definitions in place

wbclark: Regarding the existing template which mentions that the file is managed by a puppet module, I am also concerned this is not clear to foreman-installer users, or downstream users.

manifests/init.pp
```diff
@@ -132,6 +132,11 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param use_rq_tasking_system
+#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.13.
```
evgeni, line 136:
```suggestion
#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
```

I think you were the one who pointed out that the new system is only really available in 3.14 ;)

evgeni (commented):
i'll let the others do the nitpicks, looks good to me!

templates/settings.py.erb
ehelms: Thing is, this is a problem at that already exists. So we should fix it separately and consider consistency. We run the risk of bike shedding this PR and getting behind schedule. I think it will be more productive to focus this PR on the actual change, and then make the proposed change here separately. You are also banking on users actually *reading* as well ;)

ekohl: My main motivation is that we can then merge this PR without bikeshedding on the exact message. It's also nice in the changelog to have a dedicated item.

wbclark: Opened https://github.com/theforeman/puppet-pulpcore/pull/210

ehelms (approved):
Updates look good, thanks @wbclark for addressing

ekohl (approved):

PR #204, opened 2021-06-29
Refs #32910: Add ability to enable content caching

ehelms:
Adds a flag to enable content caching via Redis. This is disabled
by default as its a new optional feature in Pulp 3.14.

Requires https://github.com/pulp/pulpcore/pull/1356 being present which is targeted for Pulpcore 3.14. This also updates the README to reflect that minimum version.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -155,6 +155,12 @@
 # @param api_client_auth_cn_map
 #   Mapping of certificate common name and Pulp user to authenticate to Pulp API.
 #
+# @param cache_enabled
+#   Enables Redis based content caching within the Pulp content app.
+#
+# @param cache_expires_ttl
+#   The number of seconds that content should be cached for. Specify 'None' to never expire the cache.
+#
 # @example Default configuration
 #   include pulpcore
 #
@@ -204,6 +210,8 @@ class pulpcore (
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},
+  Boolean $cache_enabled = false,
+  Optional[Variant[Integer[1], Enum['None']]] $cache_expires_ttl = undef,
 ) {
   $settings_file = "${config_dir}/settings.py"
 
--- a/spec/acceptance/basic_spec.rb
+++ b/spec/acceptance/basic_spec.rb
@@ -117,3 +117,54 @@ describe 'reducing worker count' do
   end
 
 end
+
+describe 'with content cache enabled' do
+  certdir = '/etc/pulpcore-certs'
+
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      class { 'pulpcore':
+        cache_enabled => true,
+      }
+      PUPPET
+    end
+  end
+
+  describe service('httpd') do
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe service('pulpcore-content') do
+    it { is_expected.to be_enabled }
+    it { is_expected.to be_running }
+  end
+
+  describe port(80) do
+    it { is_expected.to be_listening }
+  end
+
+  describe port(443) do
+    it { is_expected.to be_listening }
+  end
+
+  describe port(6379) do
+    it { is_expected.to be_listening }
+  end
+
+  describe service('rh-redis5-redis'), if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
+    it { is_expected.to be_running }
+    it { is_expected.to be_enabled }
+  end
+
+  describe service('redis'), unless: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
+    it { is_expected.to be_running }
+    it { is_expected.to be_enabled }
+  end
+
+  describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/status/", cacert: "#{certdir}/ca-cert.pem") do
+    its(:response_code) { is_expected.to eq(200) }
+    its(:exit_status) { is_expected.to eq 0 }
+  end
+end
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -43,3 +43,10 @@ ALLOWED_CONTENT_CHECKSUMS = <%= scope['pulpcore::allowed_content_checksums'] %>
 
 # Derive HTTP/HTTPS via the X-Forwarded-Proto header set by Apache
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
+CACHE_ENABLED = <%= scope['pulpcore::cache_enabled'] ? 'True' : 'False' %>
+<% if scope['pulpcore::cache_enabled'] && scope['pulpcore::cache_expires_ttl'] -%>
+CACHE_SETTINGS = {
+    'EXPIRES_TTL': <%= scope['pulpcore::cache_expires_ttl'] %>,
+}
+<% end -%>
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -23,6 +23,7 @@ describe 'pulpcore' do
             .with_content(%r{ALLOWED_EXPORT_PATHS = \[\]})
             .with_content(%r{ALLOWED_IMPORT_PATHS = \["/var/lib/pulp/sync_imports"\]})
             .with_content(%r{ALLOWED_CONTENT_CHECKSUMS = \["sha224", "sha256", "sha384", "sha512"\]})
+            .with_content(%r{CACHE_ENABLED = False})
             .without_content(/sslmode/)
           is_expected.to contain_file('/etc/pulp')
           is_expected.to contain_file('/var/lib/pulp')
@@ -479,6 +480,21 @@ CONTENT
                 ],
               }
             ])
+        end
+      end
+
+      context 'can enable content caching and set an expires' do
+        let :params do
+          {
+            cache_enabled: true,
+            cache_expires_ttl: 60,
+          }
+        end
+
+        it do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{CACHE_ENABLED = True})
+            .with_content(%r{CACHE_SETTINGS = \{\n    'EXPIRES_TTL': 60,\n\}})
         end
       end
     end


ehelms:
There is one other setting that is being discussed on how it should be exposed: https://github.com/pulp/pulpcore/pull/1356/files#r660160352

I have not included it for now as my read is that it should not be user facing.

manifests/init.pp
```diff
@@ -155,6 +155,12 @@
 # @param api_client_auth_cn_map
 #   Mapping of certificate common name and Pulp user to authenticate to Pulp API.
 #
+# @param pulp_cache_enable
```
wbclark, line 158: If the goal is to have an exact correspondence with the name of the settings in settings.py (which is a fine approach) then I suppose it should be `pulp_cache_enabled` ?

Personally I am a bit more biased towards names that are concise and simple to the user, and would suggest `enable_content_caching` (the "pulp" part is clear from context) or simply `content_caching`

ehelms: I can go with `pulp_cache_enabled` to match the setting. I think discoverability wise it would be easier for those that look at Pulp for these to match. Now, over on the puppet-foreman_proxy_content side where we wrap these options, that can be a bit more nuanced.

ekohl: I left a note at https://github.com/pulp/pulpcore/pull/1356/files#r661030668 that the `pulp_` prefix is weird to me. I see @ehelms' point about matching names so let's make sure upstream picks a good one.

ehelms:
I have updated this to reflect the merged patch to Pulpcore that will be in 3.14.

README.md
```diff
@@ -10,7 +10,7 @@ All supported versions are listed below. For every supported version, acceptance
 
 Supported operating systems are listed in `metadata.json` but individual releases can divert from that. For example, if Pulpcore x.y drops EL7, it will still be listed in metadata.json until all versions supported by the module have dropped it. Similarly, if x.z adds support for EL9, it'll be listed in `metadata.json` and all versions that don't support EL9 will have a note.
 
-### Pulpcore 3.13
+### Pulpcore 3.14
```
ekohl, line 13: This should be a separate patch IMHO. For example, 3.14 also means we should update our CI.

ehelms: Works for me, I wanted to capture that this change requires this update. Should @wbclark do tha tin the worker PR or should it be the first PR, stand-alone, that goes in?

ekohl: For what it's worth, the alternative is to mention that it does nothing on 3.13 (where unknown options are just ignored).

manifests/init.pp
```diff
@@ -204,6 +210,8 @@
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},
+  Boolean $cache_enabled = false,
+  Variant[Integer[0], Enum['None']] $cache_expires_ttl = 60,
```
ekohl, line 214: Should we make this an optional parameter? Then if you don't set this, the whole `CACHE_SETTINGS` will not be included in `settings.py` and it'll use the application built in defaults.

I also wonder what `None` means. No TTL expiration and thus infinite caching?

ehelms: Correct on the None front, found that in their docs. I'll make this optional.

ekohl: We can include the `Use None for infinite caching` in our doc string.

spec/acceptance/basic_spec.rb
```diff
@@ -117,3 +117,41 @@ class { 'pulpcore':
...
 end
+
+describe 'with content cache enabled' do
+  certdir = '/etc/pulpcore-certs'
+
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      class { 'pulpcore':
+        worker_count => 2,
```
ekohl, line 128: Is the number of workers relevant for this test? If not, I'd keep it at the default and not specify it.

spec/acceptance/basic_spec.rb
```diff
@@ -117,3 +117,41 @@ class { 'pulpcore':
   end
 
 end
+
+describe 'with content cache enabled' do
```
ekohl, line 121: Should this also verify Redis is running? Perhaps even perform some content request and verify if it's requested twice that a cache hit is returned (which I think should be in the response header)

ehelms: I thought about the checking the content cache itself, and this would be a good enhancement to our testing but I'd like to do that as a follow up as it may take more time to figure out getting all the right API calls lined up. I've added the Redis check.

ekohl: I also figured a real in depth check would be complex. It does mean adding actual content into our instance which we don't do today. Perhaps we can leverage pulp-cli for this when we do get to it.

spec/acceptance/basic_spec.rb
```diff
@@ -117,3 +117,44 @@ class { 'pulpcore':
...
+    it { is_expected.to be_listening }
+  end
+
+  describe port(443) do
+    it { is_expected.to be_listening }
+  end
+
+  describe port(6379) do
+    it { is_expected.to be_listening }
+  end
```
ekohl, line 154: In addition to the port, I'd also like to check the `service('redis')` (or is it `rh-redis5-redis` on EL7?).

And on a related note: I think we should think about using a unix socket for Redis.

ekohl (commented):
Small nit, otherwise :+1:

spec/acceptance/basic_spec.rb
ehelms: Done.

templates/settings.py.erb
```diff
@@ -43,3 +43,10 @@ ALLOWED_CONTENT_CHECKSUMS = <%= scope['pulpcore::allowed_content_checksums'] %>
 
 # Derive HTTP/HTTPS via the X-Forwarded-Proto header set by Apache
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
+CACHE_ENABLED = <%= scope['pulpcore::cache_enabled'] ? 'True' : 'False' %>
+<% if scope['pulpcore::cache_expires_ttl'] -%>
```
wbclark, line 48:
```suggestion
<% if scope['pulpcore::cache_enabled'] && scope['pulpcore::cache_expires_ttl'] -%>
```

I assume including this setting when caching is disabled would not harm anything, but I think it would be better to omit unused settings from the configuration file.

manifests/init.pp
```diff
@@ -204,6 +210,8 @@
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},
+  Boolean $cache_enabled = false,
+  Optional[Variant[Integer[0], Enum['None']]] $cache_expires_ttl = undef,
```
wbclark, line 214: What does it mean to have `CACHE_SETTINGS = { 'EXPIRES_TTL': 0 }` and is that a valid configuration? Is it meaningfully different from `CACHE_SETTINGS = { 'EXPIRES_TTL': None }` ?

The docs string calls out `None` as an acceptable value but doesn't explain what happens if this is `0` instead; therefore should we use `Integer[1]` here?

wbclark: Based on https://github.com/pulp/pulpcore/commit/81684b206e3fd8bceb104e4a77c0d717387667f9#diff-a8f18c64735759c7f007d734b8b40e45938d03bc54327381fa2c9cf5fa2df544R42-R43 ,

`0` should have the exact same behavior as `None`

ehelms: That is a good point. I was aiming for following their docs rather than the code: https://github.com/pulp/pulpcore/commit/81684b206e3fd8bceb104e4a77c0d717387667f9#diff-9243f057fa8b6379f8ecb2570d457b290cf2e455e6d591541022caefc0b81448R215

We should probably point this out to them. As I would interpret 0 as meaning effectively this never gets cached.

templates/settings.py.erb
ehelms: Looking at the code, it would appear there is no default here other than if this does not exist, then treat it as infinite cache -- https://github.com/pulp/pulpcore/commit/81684b206e3fd8bceb104e4a77c0d717387667f9#diff-a8f18c64735759c7f007d734b8b40e45938d03bc54327381fa2c9cf5fa2df544R42-R43

Then I would ask, is that what we want? Would that fill up Redis unknowingly?

wbclark: I think that code path would not be traversed with `CACHE_ENABLED = False`

ekohl: There are defaults for settings, so if you don't specify it, those will be used:
https://github.com/pulp/pulpcore/blob/9c1f115e9bc01e74381f09c5f3e28b06d2ae1ce1/pulpcore/app/settings.py#L242-L246

But @wbclark is right that if the cache is disabled, there really is no point in setting a TTL.

manifests/init.pp
ekohl: Big :+1: for expanding upstream documentation. That should really be specified. If we can stick to just integers then that would be my preference.

spec/acceptance/basic_spec.rb
```diff
@@ -117,3 +117,54 @@ class { 'pulpcore':
...
+
+  describe service('rh-redis5-redis'), if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
+    it { is_expected.to be_running }
+    it { is_expected.to be_enabled }
+  end
+
+  describe service('redis'), if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 8 do
```
ekohl, line 161: Perhaps this should be `unless` with the previous condition? That way we know that exactly one of the two runs. When we look ahead to EL9, it'll be nice to have.
```suggestion
  describe service('redis'), unless: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
```

manifests/init.pp
gerrod3: You are right,`0` and `None` have the exact same behavior currently, I'll add the clarification to the docs.

templates/settings.py.erb
gerrod3: It should use the default of 1 day if you don't set it explicitly. You might be interested in setting the `maxmemory` option for Redis just to make sure that Redis doesn't grow to eat up all the RAM (unlikely IMO since the entries will expire and are very small to begin with). https://redis.io/topics/lru-cache

spec/acceptance/basic_spec.rb
```diff
@@ -117,3 +117,54 @@ class { 'pulpcore':
...
+  describe port(6379) do
+    it { is_expected.to be_listening }
+  end
+
+  describe service('rh-redis5-redis'), unless: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 8 do
+    it { is_expected.to be_running }
+    it { is_expected.to be_enabled }
+  end
+
+  describe service('redis'), unless: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
```
ekohl, line 161: What I meant in https://github.com/theforeman/puppet-pulpcore/pull/204#discussion_r663782249 was that one entry would be `if` and the other `unless` with the same condition in both places.
```suggestion
  describe service('rh-redis5-redis'), if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
    it { is_expected.to be_running }
    it { is_expected.to be_enabled }
  end

  describe service('redis'), unless: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
```

ekohl (commented):
Small clarification on the test part, other than that :+1:

ekohl (approved):
@wbclark want to have a last look as well?

wbclark (approved):

PR #205, opened 2021-06-30
Generate db_encrypted_fields_keyfile

wbclark:
Pulpcore 3.15 will require a Fernet symmetric encryption
key to encrypt certain sensitive database fields.
This is expected to contain 32 pseudorandom bytes in
url-safe base64-encoded format, with padding.


wbclark:
TODO: Update settings.py (depends on implementation details in pulpcore 3.15)

manifests/init.pp
```diff
@@ -205,7 +205,9 @@
   Integer[0] $api_service_worker_timeout = 90,
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},
 ) {
-  $settings_file = "${config_dir}/settings.py"
+  $settings_file               = "${config_dir}/settings.py"
```
ekohl, line 208: I've stopped aligning because it really makes diffs bigger and reviews harder.

manifests/init.pp
```diff
@@ -205,7 +205,9 @@
...
+  $db_encrypted_fields_keyfile = "${config_dir}/db_encrypted_fields_key"
+  $db_encrypted_fields_key     = extlib::cache_data('pulpcore_cache_data', 'db_encrypted_fields_key', generate_fernet_key())
```
ekohl, line 210: This must be a parameter. If you have multiple machines behind a load balancer you want a way have them share the same key.

lib/puppet/functions/generate_fernet_key.rb
```diff
@@ -0,0 +1,12 @@
+require 'securerandom'
+
+Puppet::Functions.create_function(:generate_fernet_key) do
```
ekohl, line 3: This should be namespaced to pulpcore
```suggestion
Puppet::Functions.create_function(:'pulpcore::generate_fernet_key') do
```

lib/puppet/functions/generate_fernet_key.rb
```diff
@@ -0,0 +1,12 @@
...
+  # @return 32 byte url-safe base64-encoded (with padding) Fernet symmetric encryption key
+  dispatch :generate_fernet_key do
+    return_type 'String'
```
ekohl, line 6: You can even give a stronger guarantee. It's at least 32 characters, right?
```suggestion
    return_type 'String[32]'
```

ekohl (changes requested):

manifests/init.pp
wbclark: Good catch. I've updated to make it a parameter with the type `Pattern[/\A([a-zA-Z]|\d|-|_){43}=\z/]`

lib/puppet/functions/generate_fernet_key.rb
wbclark: I'll try `Pattern[/\A([a-zA-Z]|\d|-|_){43}=\z/]` since this matches exactly what are the allowed values

Well, technically, python's `cryptography.fernet` could also hande the standard, non url-safe base64 encoding, because it uses `base64.urlsafe_b64decode` and that performs the translation at https://github.com/python/cpython/blob/86eeeb425936ba67d79f32bfbd5c5f8002819438/Lib/base64.py#L108-L133

With that said `cryptography.frenet` claims that it doesn't support this, for example here https://github.com/pyca/cryptography/blob/77fb53c75e47f50e09b1b3be3a4d10c7e4e34dc2/src/cryptography/fernet.py#L37-L41 and also stated in the docs https://github.com/pyca/cryptography/blob/77fb53c75e47f50e09b1b3be3a4d10c7e4e34dc2/docs/fernet.rst

lib/puppet/functions/generate_fernet_key.rb
wbclark: :+1:

lib/puppet/functions/generate_fernet_key.rb
wbclark: It does indeed produce the same result to load in the key using either alphabet:

```
>>> from cryptography.fernet import Fernet
>>> urlsafe_key = b'xTjQtHgU6OsktqtDGAG5ogEb8xNjT2DQGytK3hWMg-_='
>>> nonurlsafe_key = b'xTjQtHgU6OsktqtDGAG5ogEb8xNjT2DQGytK3hWMg+/='
>>> f1 = Fernet(key = urlsafe_key)
>>> f2 = Fernet(key = nonurlsafe_key)
>>> f1._encryption_key == f2._encryption_key
True
>>> f1._signing_key == f2._signing_key
True
```

But since the actual documentation says only the url-safe alphabet should be used, let's keep it limited to that one.

PR #206, opened 2021-06-30
#32917 - Deploy Redis only when required by enabled features

wbclark:
Redis is now only required by this module when using at least
    one of the `use_rq_tasking_system` or `pulp_cache_enable` features.
    This commit introduces a new parameter `manage_redis`, default to true,
    which should be used when this module is responsible for managing Redis
    deployment. When false, this module will never attempt to deploy Redis,
    although it will still configure Pulpcore's connection to Redis if
    required by one or more enabled features.

Requires: https://github.com/theforeman/puppet-pulpcore/pull/203
Requires: https://github.com/theforeman/puppet-pulpcore/pull/204

### DIFF
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -45,3 +45,8 @@ ALLOWED_CONTENT_CHECKSUMS = <%= scope['pulpcore::allowed_content_checksums'] %>
 
 # Derive HTTP/HTTPS via the X-Forwarded-Proto header set by Apache
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+
+PULP_CACHE_ENABLED = <%= scope['pulpcore::pulp_cache_enable'] ? 'True' : 'False' %>
+PULP_CACHE_SETTINGS = {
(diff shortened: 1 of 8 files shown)


ehelms:
I could see doing a version of this PR without the content caching. Since we discussed what to do about Redis with respect to the worker PR, and the workers change is the one we absolutely want to have, I'd prefer to aim for a layering of changes that is like:

 * worker change
 * installing redis change
 * content cache change
 * manage redis change

templates/settings.py.erb
```diff
@@ -19,9 +19,14 @@ DATABASES = {
 <% end -%>
     },
 }
+
+<% if scope['pulpcore::configure_redis_connection'] -%>
 REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
+<% end -%>
+
```
ekohl, line 28: Slight nit: this leaves 2 empty lines in the generated file so I'd suggest to include empty line inside the if/end block. Github doesn't render this well, but roughly
```suggestion

<% end -%>
```

ehelms:
The other PR has been merged.

PR #207, opened 2021-07-01
Refs #32917 - Don't deploy or configure Redis with new tasking system

wbclark:
This is a more gradual alternative to https://github.com/theforeman/puppet-pulpcore/pull/206

In particular, it doesn't include the ability to enable or disable management of Redis, nor does it consider the content caching feature.

it simply includes the redis class and configures Pulpcore for Redis, based on the value of `use_rq_tasking_system`

### DIFF
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -23,8 +23,7 @@ describe 'pulpcore' do
             .with_content(%r{ALLOWED_EXPORT_PATHS = \[\]})
             .with_content(%r{ALLOWED_IMPORT_PATHS = \["/var/lib/pulp/sync_imports"\]})
             .with_content(%r{ALLOWED_CONTENT_CHECKSUMS = \["sha224", "sha256", "sha384", "sha512"\]})
-            .with_content(%r{CACHE_ENABLED = False})
-            .without_content(/sslmode/)
+            .without_content(%r{sslmode})
           is_expected.to contain_file('/etc/pulp')
           is_expected.to contain_file('/var/lib/pulp')
           is_expected.to contain_file('/var/lib/pulp/sync_imports')
@@ -41,7 +40,7 @@ describe 'pulpcore' do
           is_expected.to contain_exec('pulpcore-manager collectstatic --noinput')
         end
 
-        it 'configures the database' do
+        it 'configures the PostgreSQL database' do
           is_expected.to contain_class('pulpcore::database')
           is_expected.to contain_class('postgresql::server')
           is_expected.to contain_postgresql__server__db('pulpcore')
@@ -49,6 +48,30 @@ describe 'pulpcore' do
           is_expected.to contain_exec('pulpcore-manager migrate --noinput')
           is_expected.to contain_pulpcore__admin('reset-admin-password --random')
           is_expected.to contain_exec('pulpcore-manager reset-admin-password --random')
+        end
+
+        it 'does not install Redis' do
+          is_expected.not_to contain_class('redis')
+        end
+
+        it 'does not configure Pulpcore connection to Redis' do
+          is_expected.to contain_concat__fragment('base')
+            .without_content(%r{REDIS_HOST})
+            .without_content(%r{REDIS_POST})
+            .without_content(%r{REDIS_DB})
+        end
+
+        it 'does not configure content caching' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(/CACHE_ENABLED = False/)
+            .without_content(%r{CACHE_SETTINGS})
+            .without_content(%r{'EXPIRES_TTL':})
+        end
+
+        it 'configures pulpcore to use PostgreSQL tasking system' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{USE_NEW_WORKER_TYPE = True})
+          is_expected.to contain_systemd__unit_file('pulpcore-resource-manager.service').with_ensure('absent')
         end
 
         it 'configures apache' do
@@ -128,7 +151,6 @@ describe 'pulpcore' do
           is_expected.to contain_systemd__unit_file('pulpcore-content.socket')
           is_expected.to contain_systemd__unit_file('pulpcore-content.service')
           is_expected.to contain_file('/etc/systemd/system/pulpcore-content.socket').that_comes_before('Service[pulpcore-content.service]')
-          is_expected.to contain_systemd__unit_file('pulpcore-resource-manager.service').with_ensure('absent')
           is_expected.to contain_systemd__unit_file('pulpcore-worker@.service')
           is_expected.to contain_service("pulpcore-worker@1.service").with_ensure(true)
           is_expected.not_to contain_service("pulpcore-worker@2.service")
@@ -491,10 +513,45 @@ CONTENT
           }
         end
 
-        it do
+        it 'configures content caching' do
           is_expected.to contain_concat__fragment('base')
             .with_content(%r{CACHE_ENABLED = True})
             .with_content(%r{CACHE_SETTINGS = \{\n    'EXPIRES_TTL': 60,\n\}})
+        end
+
+        it 'installs Redis' do
+          is_expected.to contain_class('redis')
+        end
+
+        it 'configures Pulpcore connection to Redis' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{REDIS_HOST})
+            .with_content(%r{REDIS_PORT})
+            .with_content(%r{REDIS_DB})
+        end
+      end
+
+      context 'can enable RQ tasking system' do
+        let :params do
+          {
+            use_rq_tasking_system: true,
+          }
+        end
+
+        it 'configures RQ tasking system' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{USE_NEW_WORKER_TYPE = False})
+        end
+
+        it 'installs Redis' do
+          is_expected.to contain_class('redis')
+        end
+
+        it 'configures Pulpcore connection to Redis' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{REDIS_HOST})
+            .with_content(%r{REDIS_PORT})
+            .with_content(%r{REDIS_DB})
         end
       end
     end
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -19,10 +19,13 @@ DATABASES = {
 <% end -%>
     },
 }
+
+<% if scope['pulpcore::uses_redis'] -%>
 REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
 
+<% end -%>
 USE_NEW_WORKER_TYPE = <%= scope['pulpcore::use_rq_tasking_system'] ? "False" : "True" %>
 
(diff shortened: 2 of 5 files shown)


wbclark:
rebased and added logic to handle content caching

ekohl:
Please rebase now that https://github.com/theforeman/puppet-pulpcore/pull/203 is merged.

wbclark:
```
 Failures:

  1) basic installation Command "DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig" stdout is expected to match /^0 workers, /
     Failure/Error: its(:stdout) { is_expected.to match(/^0 workers, /) }
       expected "Error 99 connecting to localhost:6379. Cannot assign requested address.\n" to match /^0 workers, /
       Diff:
       @@ -1 +1 @@
       -/^0 workers, /
       +Error 99 connecting to localhost:6379. Cannot assign requested address.
       
     # ./spec/acceptance/basic_spec.rb:83:in `block (3 levels) in <top (required)>'

  2) basic installation Command "DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig" exit_status is expected to eq 0
     Failure/Error: its(:exit_status) { is_expected.to eq 0 }
       
       expected: 0
            got: 1
       
       (compared using ==)
       
     # ./spec/acceptance/basic_spec.rb:85:in `block (3 levels) in <top (required)>'

Finished in 8 minutes 54 seconds (files took 1 minute 7.13 seconds to load)
176 examples, 2 failures
```


This I can fix, but what I want to point out is, we only don't see similar results in some of the enable/disable rq tasking tests, because of the order in which those tests are running.... once redis gets installed once, the failure won't occur.

ekohl:
That sounds like Pulp has a hard dependency on Redis and something we should take upstream.

templates/settings.py.erb
```diff
@@ -19,9 +19,12 @@ DATABASES = {
 <% end -%>
     },
 }
+
+<% if scope['pulpcore::use_rq_tasking_system'] || scope['pulpcore::cache_enabled'] -%>
 REDIS_HOST = "localhost"
 REDIS_PORT = "<%= scope['redis::port'] %>"
 REDIS_DB = <%= scope['pulpcore::redis_db'] %>
+<% end -%>
```
ekohl, line 28: I'd have the empty line within the block. That way `settings.py` can't end up with 2 blank lines:
```suggestion

<% end -%>
```

ekohl (commented):
You can also consider a line in `init.pp` like `$uses_redis = $use_rq_tasking_system or $cache_enabled` and use that variable in `database.pp` and `settings.py`.

wbclark:
> That sounds like Pulp has a hard dependency on Redis and something we should take upstream.

That isn't how I interpreted it. If you look closely at what happens:

```
[root@centos8-64-1 ~]# DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig
Error 111 connecting to localhost:6379. Connection refused.
```

I'll update that test.

Additionally, I'm going to move the 'enable content caching' test out of basic_spec.rb where we have multiple top-level describe blocks. I'd like to move the reducing worker count test out into its own test as well, but there is already a PR open for it (#209 )

wbclark:
> You can also consider a line in `init.pp` like `$uses_redis = $use_rq_tasking_system or $cache_enabled` and use that variable in `database.pp` and `settings.py`.

Probably a good idea to consolidate the logic in this way. Thanks

I was actually taking a similar approach with #206 and I'm not sure the reason I didn't follow it here. Probably a simple oversight

spec/classes/redis_spec.rb
```diff
@@ -0,0 +1,89 @@
+require 'spec_helper'
```
ehelms, line 1: You could simplify your testing approach here by adding an assertion to https://github.com/theforeman/puppet-pulpcore/blob/master/spec/classes/pulpcore_spec.rb#L486 that `redis` is included, and then in the defaults (https://github.com/theforeman/puppet-pulpcore/blob/master/spec/classes/pulpcore_spec.rb#L8) that redis is not included. 

Then you could add a small block for the enabling of rq tasking system similar to https://github.com/theforeman/puppet-pulpcore/blob/master/spec/classes/pulpcore_spec.rb#L486 and include a similar assertion of redis.

In this way you'd keep all of the assertions logically together, and they would reflect the parameters being passed in and how that affects things. This method, while it does work, I think it's downside is it separates the expectations.

wbclark: After trying this out, I see what you mean. It's a very nice approach.

Updated with the change. There is no longer a separate `spec/classes/redis_spec.rb` -- now the cases of default arguments, RQ tasking system, and content caching, are all covered in the `spec/classes/pulpcore_spec.rb`

spec/classes/pulpcore_spec.rb
```diff
@@ -51,6 +51,29 @@
...
+            .without_content(/REDIS_HOST/)
+            .without_content(/REDIS_POST/)
+            .without_content(/REDIS_DB/)
+        end
+
+        it 'does not configure content caching' do
+          is_expected.to contain_concat__fragment('base')
+            .without_content(/CACHE_ENABLED = True/)
+            .without_content(%r{CACHE_SETTINGS = \{\n    'EXPIRES_TTL': 60,\n\}})
+        end
```
ehelms, line 69: This is already covered by https://github.com/theforeman/puppet-pulpcore/pull/207/files#diff-257fe3136ecfdbfcb123b6445ff2de6c21d1fe08b0d6af54c08dde40637d2594R26

spec/classes/pulpcore_spec.rb
```diff
@@ -51,6 +51,29 @@
...
+
+        it 'configures pulpcore to use PostgreSQL tasking system' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(/USE_NEW_WORKER_TYPE = True/)
+          is_expected.to contain_systemd__unit_file('pulpcore-resource-manager.service').with_ensure('absent')
+        end
```
ehelms, line 75: This are OK here, but they feel better up here https://github.com/theforeman/puppet-pulpcore/pull/207/files#diff-257fe3136ecfdbfcb123b6445ff2de6c21d1fe08b0d6af54c08dde40637d2594R19 for me as it connects all of the configuration file together.

spec/classes/pulpcore_spec.rb
```diff
@@ -491,11 +513,46 @@
           }
         end
 
-        it do
+        it 'configures content caching' do
           is_expected.to contain_concat__fragment('base')
-            .with_content(%r{CACHE_ENABLED = True})
+            .with_content(/CACHE_ENABLED = True/)
```
ehelms, line 518: Why this change? I think the linter prefers `%r{}` syntax

spec/classes/pulpcore_spec.rb
```diff
@@ -51,6 +51,29 @@
           is_expected.to contain_exec('pulpcore-manager reset-admin-password --random')
         end
 
+        it 'does not install Redis' do
+          is_expected.not_to contain_class('redis')
+        end
+
+        it 'does not configure Pulpcore connection to Redis' do
+          is_expected.to contain_concat__fragment('base')
+            .without_content(/REDIS_HOST/)
```
ehelms, line 60: It's a bit of a bikeshed, but for consistency within the file the `%r{}` is used for regex already and I want to say that is preferred by the linter (I swear something yelled at me recently for not using it).

spec/classes/pulpcore_spec.rb
wbclark: My understanding was, `%r{}` is preferred only when matching more than one `/` character. Our usage throughout is pretty inconsistent though, I noticed. So I was planning to come back and make a larger pass at it in my test refactor PR.

I made the change here since I copied that code to another part of the file where I made the change, so it felt right to make it here as well for consistency.

spec/classes/pulpcore_spec.rb
wbclark: I started to think that it makes sense to structure the tests around the real world use case rather than the specific object being tested..

So instead of having all tests for the configuration file in the same place, have a test for 'content caching is disabled', 'postgresql tasking system is used', and so on.

I see two primary benefits: 1. This feels more maintainable in the longer term (when we make modifications to a specific feature, the tests for that feature are all located in that block, rather than spread out across different components the feature touches). 2. This should lead to more readable test output, and particularly failures. So instead of getting a failure like `default parameters - it configures pulpcore - it is expected to contain a file with (lots of stuff, including all kinds of things that aren't relevant to the specific thing you broke)` you would instead see something more like `default parameters - it uses new tasking system - it is expected to contain a configuration file with specific things that are relevant to the tasking system`

I want to pursue this idea more consistently in my planned refactoring. For this PR I thought, it makes sense to write the net-new tests using this style, to minimize the amount of re-writing I have to do later.

spec/classes/pulpcore_spec.rb
wbclark: I can revert this for now, use the same style even in my new code, and tackle this in the other PR.

spec/classes/pulpcore_spec.rb
wbclark: Indeed you're correct.

I thought we were mixing `//` and `%r{}` already in this file, BUT it turns out the instances of `//` I saw are only the ones I introduced.

I'll update to use `%r{}` consistently for now, and we can work out in the larger test PR I have planned whether one or both styles should actually be used.

spec/classes/pulpcore_spec.rb
wbclark: Nice catch, thank you.

wbclark:
Thanks for feedback @ehelms !

I pushed an update to address these items. In particular I use `%r{}` consistently throughout `spec/classes/pulpcore_spec.rb` (I thought we were already mixing `//` often but I was mistaken -- there was only a single, very small instance of that before the code I introduced here)

Regarding the overall structure of the tests, I did opt to keep for now the change of moving some tests of the configuration file contents out of the `it 'configures pulpcore'` block and into feature-specific blocks. Arguably that should be handled in my other test refactoring PR, but I'm balancing that fact vs. the fact that I already need to add these feature-specific blocks here now, and it's less work for me overall to not write it one way now, knowing that I'm planning to re-write it a different way soon.

With that said, if any of those changes are really controversial, I can revert them and tackle them later. The goal in not writing the same code twice is to save time. If it's not actually saving me time in the end, I'll go ahead and do what is asked of me here, and re-write it the way I think it really should be in the other PR.

ehelms:
If you are curious where I got the using `%r{}` everywhere, this comes from some initial work @ekohl did to add Rubocop. You can see what that looks like for example here, https://github.com/theforeman/puppet-qpid/pull/167/files. This uses an inherited Rubocop configuration from `voxpupuli-test` which has:

https://github.com/voxpupuli/voxpupuli-test/blob/master/rubocop.yml#L467-L470

Thus, they have chosen to enforce the use of `%r{}` as the way to define regex.

ehelms (approved):

ekohl:
> That sounds like Pulp has a hard dependency on Redis and something we should take upstream.

Looks like we're running into this now: https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1027-failed/24403

PR #209, opened 2021-07-02
Refactor pulpcore worker_count tests and add additional tests

wbclark:
This commit splits the acceptance tests for worker count changes out
into a separate file, spec/acceptance/worker_count.rb. It also adds
new cases to spec tests to ensure proper interaction of configured
worker_count, previously configured worker_count, and cpu count.

### DIFF
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -526,6 +526,66 @@ CONTENT
             .with_content(%r{^WORKER_TTL = 60$})
         end
       end
+
+      context 'with 24 processors from facts and default worker_count' do
+        let(:facts) { override_facts(os_facts, processors: {count: 24}) }
+
+        it 'enables only 8 pulpcore workers' do
+          (1..8).each do |i|
+            is_expected.to contain_service("pulpcore-worker@#{i}.service")
+              .with_ensure(true)
+              .with_enable(true)
+          end
+
+          (9..24).each do |i|
+            is_expected.not_to contain_service("pulpcore-worker@#{i}.service")
+          end
+        end
+      end
+
+      context 'with 24 workers previously enabled and reset to default worker_count' do
+        let(:facts) {
+          override_facts(
+            os_facts,
+            processors: {count: 24},
+            pulpcore_workers: (1..24).map { |i| "pulpcore-worker@#{i}.service" }
+          )
+        }
+
+        it 'enables 8 pulpcore workers' do
+          (1..8).each do |i|
+            is_expected.to contain_service("pulpcore-worker@#{i}.service")
+              .with_ensure(true)
+              .with_enable(true)
+          end
+        end
+
+        it 'disables existing workers from previous configuration' do
+          (9..24).each do |i|
+            is_expected.to contain_service("pulpcore-worker@#{i}.service")
+              .with_ensure(false)
+              .with_enable(false)
+          end
+        end
+      end
+
+      context 'configured for 12 workers and only 8 processors' do
+        let(:facts) { override_facts(os_facts, processors: { count: 8 }) }
+
+        let(:params) do
+          {
+            worker_count: 12
+          }
+        end
+
+        it 'allows configuring more workers than processors' do
+          (1..12).each do |i|
+            is_expected.to contain_service("pulpcore-worker@#{i}.service")
+              .with_ensure(true)
+              .with_enable(true)
+          end
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/acceptance/reduce_worker_count_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'spec_helper_acceptance'
```
ehelms, line 1: Perhaps `worker_count_spec.rb` is more apt? Since this tests increasing and decreasing it?

wbclark: the acceptance test only reduces it (which was the more difficult part to implement, relying on a custom fact to "remember" the previous worker count after the param has been changed)

I could add an acceptance test to increase it again if you see value in having that, either in this file or as a separate `increase_worker_count_spec.rb`. I suppose idempotence means either approach is sufficient, while increasing it again in the same test would be more computationally efficient

spec/classes/pulpcore_spec.rb
```diff
@@ -481,6 +481,66 @@
             ])
         end
       end
+
+      context 'with 24 processors from facts and default worker_count' do
+        let(:facts) { override_facts(os_facts, processors: {count: 24}) }
+
+        it 'enables only 8 pulpcore workers' do
+          (1..8).to_a.each do |i|
```
ekohl, line 489: Do you need to use `.to_a` here? I think this also works
```suggestion
          (1..8).each do |i|
```
(same below)

spec/acceptance/reduce_worker_count_spec.rb
ekohl: I think I agree with @ehelms here and make it really a file about worker counts. Then you can also include the CPU-based tests in the file.

spec/classes/pulpcore_spec.rb
wbclark: Thanks! I tested this and it works just as well.

It's been a while since I was first exposed to the concept of "Duck Typing" in Ruby, which I guess usually means using `object.method if object.responds_to?(method)` instead of `object.method if object.is_a?(specific_class_that_implements_method)`.

But under that same concept I would also include an understanding that if multiple classes implement the "same" method, and an object can be converted back and forth between the two classes without any loss of data (or if the conversion can be performed one-way without any loss of data which is relevant to that particular method, i.e. if the conversion is an "injective morphism" to borrow a term from math) then broadly speaking you should expect the same result whether calling the method on the initial object or converting the object and then calling the method. And in that case, there's no point in performing the conversion.... just call the method on the first object.

This is all an unnecessarily long winded way of saying: thanks for the suggestion, I found this to be not only useful for improving this code, but also generally kind of neat and led me to an understanding which should make me a better programmer in the longer term.

spec/acceptance/reduce_worker_count_spec.rb
wbclark: I currently have the CPU based tests implemented in the unit tests but not in the acceptance tests.

I don't think we've used `override_facts` in acceptance tests before -- pretty neat if that works. (I think you are the original author of `override_facts`? so I assume you made this recommendation knowing that it works)

I will give this a try... thanks for the suggestions

wbclark:
Pushed a partial update. Still needs a rebase and I have several remaining items to tackle:

1. Expand `spec/acceptance/worker_count_spec.rb` to include additional tests like increasing worker count, test behavior with different CPU counts from facts.

2. Eliminate multiple top-level describe blocks in `spec/acceptance/basic_spec.rb` (another one was introduced with the content caching feature, this should be split out into its own file)

3. Implement cleanup of the test environment between tests as necessary (perhaps using something like https://github.com/theforeman/puppet-foreman/blob/master/spec/support/acceptance/purge.rb )

4. Improve maintainability of tests and readbility of test output by testing feature areas instead of component areas. Instead of having a single block for the main configuration file under default parameters, have separate blocks for each feature like 'enable rq tasking system', 'enable content caching' and test only the relevant pieces of the configuration file for that feature. This makes the tests easier to update when we make future changes to features, and makes the test output more readable since failure output shouldn't include a bunch of general things that are irrelevant to the specific feature that got broken.

5. Give titles to `it` blocks where it makes sense to do so: `it do` -> `it 'does something meaningful to the reader' do`

6. be more consistent in use of `//` vs. `%r{}` for regular expressions

7. rebase on redis PR and solve merge conflicts

ehelms:
Needs a rebase with all the latest changes going in

wbclark:
I've rebased this and will have some further updates to it soon after breaking for dinner....

ekohl (approved):

PR #212, opened 2021-07-08
Fixes #32968: Add ability to modify log level

ehelms:
Comes from this request by Pulp developers to help with debugging: https://bugzilla.redhat.com/show_bug.cgi?id=1977693

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -166,6 +166,9 @@
 # @param cache_expires_ttl
 #   The number of seconds that content should be cached for. Specify 'None' to never expire the cache.
 #
+# @param log_level
+#   Sets the log level.
+#
 # @example Default configuration
 #   include pulpcore
 #
@@ -218,6 +221,7 @@ class pulpcore (
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},
   Boolean $cache_enabled = false,
   Optional[Variant[Integer[1], Enum['None']]] $cache_expires_ttl = undef,
+  Enum['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG'] $log_level = 'INFO',
 ) {
   $settings_file = "${config_dir}/settings.py"
 
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -23,6 +23,7 @@ describe 'pulpcore' do
             .with_content(%r{ALLOWED_EXPORT_PATHS = \[\]})
             .with_content(%r{ALLOWED_IMPORT_PATHS = \["/var/lib/pulp/sync_imports"\]})
             .with_content(%r{ALLOWED_CONTENT_CHECKSUMS = \["sha224", "sha256", "sha384", "sha512"\]})
+            .with_content(%r{\s'level': 'INFO',})
             .without_content(%r{sslmode})
           is_expected.to contain_file('/etc/pulp')
           is_expected.to contain_file('/var/lib/pulp')
@@ -552,6 +553,19 @@ CONTENT
             .with_content(%r{REDIS_HOST})
             .with_content(%r{REDIS_PORT})
             .with_content(%r{REDIS_DB})
+        end
+      end
+
+      context 'can change the log level to DEBUG' do
+        let :params do
+          {
+            log_level: 'DEBUG'
+          }
+        end
+
+        it do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{\s'level': 'DEBUG',})
         end
       end
     end
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -55,3 +55,13 @@ CACHE_SETTINGS = {
     'EXPIRES_TTL': <%= scope['pulpcore::cache_expires_ttl'] %>,
 }
 <% end -%>
+
+LOGGING = {
+    "dynaconf_merge": True,
+    "loggers": {
+        '': {
+            'handlers': ['console'],
+            'level': '<%= scope['pulpcore::log_level'] %>',
+        }
+    }
+}


wbclark (commented):
Thanks @mdellweg for the suggestion and @ehelms for building it

Could you create and link a Redmine so this PR and the BZ are easily discoverable from one another?

wbclark:
Normally we manage log levels centrally with hammer:

```
# hammer admin logging --list
-----------|-------------------------------------|-------------------------------------
COMPONENT  | AUTO-DETECTED BY EXISTENCE OF       | DESTINATIONS                        
-----------|-------------------------------------|-------------------------------------
postgresql | /var/lib/pgsql/data/postgresql.conf | syslog /var/lib/pgsql/data/pg_log/  
rails      | /etc/foreman/settings.yaml          | /var/log/foreman/production.log     
proxy      | /etc/foreman-proxy/settings.yml     | /var/log/foreman-proxy/proxy.log    
puppet     | /etc/puppet/puppet.conf             | /var/log/puppet/masterhttp.log      
dhcpd      | /etc/dhcp/dhcpd.conf                | syslog /var/log/dhcpd-debug.log     
named      | /etc/named.conf                     | syslog                              
tftp       | /etc/xinetd.d/tftp                  | syslog                              
qpidd      | /etc/qpid/qpidd.conf                | syslog                              
tomcat     | /etc/candlepin/candlepin.conf       | /var/log/candlepin/ /var/log/tomcat/
pulp       | /etc/pulp/server.conf               | syslog /var/log/pulp-debug.log      
virt-who   | /etc/sysconfig/virt-who             | syslog                              
-----------|-------------------------------------|-------------------------------------
```

We can see Pulp2 was in the component list that could be managed this way.

A commonly used command to set the level of all components to debug is `hammer admin logging -a -d` and for a specific component like Pulp only `hammer admin logging -c pulp -d`. These require a service restart to take effect.

While I don't see a harm in allowing this module and/or foreman installer to set the configured log level (and keep resetting it at every installer run) isn't it better to provide the ability to set the currently running log level without requiring a full installer run and using the same tool which is already used to change these in the debugging process?

ehelms:
> While I don't see a harm in allowing this module and/or foreman installer to set the configured log level (and keep resetting it at every installer run) isn't it better to provide the ability to set the currently running log level without requiring a full installer run and using the same tool which is already used to manage log level for other components?

I get your point. However, modifying the logging outside of the thing that manages configuration seems like a bad idea. You are effectively making a one off change that can then be wiped out. If `hammer admin` is changing managed configuration files to modify the log levels, while useful, seems like a bad design choice. The installer is the thing that manages how the system is configured and I think we should adhere to that pattern as a core tenet.

wbclark:
> > While I don't see a harm in allowing this module and/or foreman installer to set the configured log level (and keep resetting it at every installer run) isn't it better to provide the ability to set the currently running log level without requiring a full installer run and using the same tool which is already used to manage log level for other components?
> 
> I get your point. However, modifying the logging outside of the thing that manages configuration seems like a bad idea. You are effectively making a one off change that can then be wiped out. If `hammer admin` is changing managed configuration files to modify the log levels, while useful, seems like a bad design choice. The installer is the thing that manages how the system is configured and I think we should adhere to that pattern as a core tenet.

I don't think `hammer admin` is used to persistently configure logging though. Instead it's used during troubleshooting to temporarily change log levels, followed by changing them back at the end of the troubleshooting session.

Installer runs can be lengthy sometimes and I can say with absolutely certainty, drawing on my past experience, that support engineers NEED the ability to change log levels doing remote sessions without having to run the installer.

I would say that having it persistently configured by the installer is fine, and indeed having the configuration in place makes it much easier for hammer to modify the configuration temporarily. But this alone shouldn't be considered sufficient to satisfy the need described by the above BZ. `hammer admin` is the standard way to make these changes temporarily during a troubleshooting session.

wbclark:
Following up, even when not on a remote session, a common asynchronous troubleshooting request (i.e. via email) from a support engineer to a customer would be like:

```shell
# hammer admin logging -a -d             # set log level to debug
# foreman-maintain service restart      # restart services to apply above change

# foreman-tail | tee >/tmp/tail.out         # leave this running in a separate terminal, cancel process and upload /tmp/tail.out when finished

# some steps here to reproduce the problem such as running hammer commands, sending request to API, clicking on WebUI, running playbook

# hammer admin logging -a -d             # set logging back to production levels
# foreman-maintain service restart      # restart services to apply change
```

This type of request is very standard and common in support, and asking support engineers (or in turn, end users) to run the installer twice during this process just to modify the configured log level would be poorly received.

So I think where I'm landing is, including the configuration for logging in `templates/settings.py.erb` is a good change. I'm not sure whether it really needs to be configurable here or if there would be any use case for that, or if it should simply be added as a static snippet that sets the level `INFO`. And either way, `hammer admin` needs to be updated to support modifying this configuration.

ekohl:
This is literally the first time I find out that can can configure logging.

wbclark:
> This is literally the first time I find out that can can configure logging.

See, this is why you guys needed someone with a support background 😂

ehelms:
I know hammer admin was created a long time ago, and breaks the hammer mold. That feels like something that ought to be in foreman-maintain's realm from a concept stand point. I am bike shedding a bit off topic :)

ehelms:
I connected a Redmine.

wbclark:
> I know hammer admin was created a long time ago, and breaks the hammer mold. That feels like something that ought to be in foreman-maintain's realm from a concept stand point. I am bike shedding a bit off topic :)

That makes sense to me, and I would agree with setting a separate goal to move short term changes to log level for all services from `hammer admin` to `foreman-maintain`. I think unless that's something we're willing to prioritize and deliver quickly, we still need to add Pulpcore to the services which can have logging configured by `hammer admin` so that users can configure it from a single place using the same tool which is already used for all other services.

wbclark (changes requested):
Regarding this PR specifically, I don't really see a reason to _persistently_ configure Pulpcore to run with a log level other than `INFO`. And I definitely don't want to ask users to run the installer _twice_ to change the log level and then change it back again.

So I think this PR should add the log level `INFO` as a static configuration in `templates/settings.py.erb` without any parameter to change that. Running the installer would then always reset the configured log level to `INFO` and I think that is the proper behavior.

That is still the first step in solving the problem mentioned by the associated BZ.

Then I think for the next step, there are options, and each has benefits, drawbacks, and varying levels of feasibility given the nearness of the upcoming release:

1. Move short-term management of logging level for all services from `hammer admin` to `foreman-maintain`. This is the "cleanest" solution, and we can discuss with @upadhyeammit if this is something we can deliver quickly in foreman-maintain (but my gut says realistically I don't think there is any way we could deliver this by the next release).

2. Implement the functionality in `hammer admin` so that users can continue using the same tool to manage log level for services for now (not the most efficient to add code that we are planning to remove soon, but I think it's not likely to require too much development effort, and given the overwhelming likelihood that we can't deliver the feature in foreman-maintain in time for the release, this approach is the only one that doesn't deliver significantly degraded user experience and potentially harm user perception of the 'readiness' of pulp3. also even if this requires less effort, it still may not be feasible to deliver it before the release at this point)

3. do not implement the feature immediately in `hammer admin` (because it will be removed anyway, so skipping it saves time) but do not implement the feature immediately in `foreman-maintain` either (because there isn't time to do so). for now, users must change log levels by using the `hammer admin` command for all other servers, and additionally making a manual edit to `/etc/pulp/settings.py`, followed by restarting services (as they already do when using `hammer admin`). _This is still preferable to requiring a full run of the installer twice in order to change the log level and then change it back_.

Of these approaches, I think the first is not realistically an option given this timeline, so it probably comes down to the 2nd vs. the 3rd. I think the 3rd option could be harmful to user first impressions of Pulp3 and for that reason the 2nd approach is my preference. With that said, the 3rd option might be the only one that is realistic at this point.

wbclark:
NOTE:

it turns out that `tfm-rubygem-hammer_cli_foreman_admin` is not installed by default upstream

after installing the RPM on my box (it was available in foreman-plugins repository), this feature now appears in hammer

wbclark:
https://github.com/theforeman/hammer-cli-foreman-admin/pull/12 works for me to change between `DEBUG` and `INFO`, once the necessary configuration is added to /etc/pulp/settings.py

with the combination of these 2 changes, users can continue to use this existing workflow to change log levels during troubleshooting

BTW, another benefit of using foreman-maintain for this would be that it works on both katello and content-proxy server. Overall :+1: to making that change, but I felt motivated to cover the gap in the meantime until we get there.

ehelms:
> Regarding this PR specifically, I don't really see a reason to _persistently_ configure Pulpcore to run with a log level other than `INFO`. And I definitely don't want to ask users to run the installer once to change the log level, let alone _twice_ to change the log level and then change it back again.

This PR does not expose this as a parameter to the installer, thus you would have to configure this via heira without a puppet-foreman_proxy_content PR. If a user wants to modify the log levels persistently due to how they like to gather logs and store them in an external server, should we stop them? We allow users to configure Foreman logging today. If we need to first have the broad discussion on our design principles then lets sit on this and have that conversation.

wbclark:
> > Regarding this PR specifically, I don't really see a reason to _persistently_ configure Pulpcore to run with a log level other than `INFO`. And I definitely don't want to ask users to run the installer once to change the log level, let alone _twice_ to change the log level and then change it back again.
> 
> This PR does not expose this as a parameter to the installer, thus you would have to configure this via heira without a puppet-foreman_proxy_content PR. If a user wants to modify the log levels persistently due to how they like to gather logs and store them in an external server, should we stop them? We allow users to configure Foreman logging today. If we need to first have the broad discussion on our design principles then lets sit on this and have that conversation.

That's fair -- I just can't imagine where else this is going as it is currently written.

Realistically, do users want or need the ability to persistently configure the log level for the application to any level other than `INFO`?

Could we start with adding the logging configuration statically to the template, and if there is a need to configure the level via a parameter, add that functionality as a follow up?

wbclark (approved):
After some IRC discussion:

I still don't believe this is a good way for installer users to modify configuration for logging, so I wouldn't want to expose this parameter in puppet-FPC for example, but this is a good feature to have for users of the standalone Puppet module.

PR #214, opened 2021-07-12
acceptance: rq info should return connection refused when redis is no…

wbclark:
…t installed

The test was enforcing incorrect behavior. This commit fixes the typo

### DIFF
--- a/spec/acceptance/basic_spec.rb
+++ b/spec/acceptance/basic_spec.rb
@@ -79,10 +79,19 @@ describe 'basic installation' do
     its(:exit_status) { is_expected.to eq 0 }
   end
 
-  describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py /usr/libexec/pulpcore/rq info -c pulpcore.rqconfig") do
-    its(:stdout) { is_expected.to match(/^0 workers, /) }
-    its(:stdout) { is_expected.not_to match(/^resource-manager /) }
-    its(:exit_status) { is_expected.to eq 0 }
+  describe service('rh-redis5-redis'), if: %w[centos redhat].include?(os[:family]) && os[:release].to_i == 7 do
+    it { is_expected.not_to be_running }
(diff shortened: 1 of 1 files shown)


spec/acceptance/basic_spec.rb
```diff
@@ -90,7 +90,7 @@ class { 'pulpcore':
   end
 
   describe command("DJANGO_SETTINGS_MODULE=pulpcore.app.settings PULP_SETTINGS=/etc/pulp/settings.py rq info -c pulpcore.rqconfig") do
-    its(:stdout) { is_expected.not_to match(/Connection refused/) }
+    its(:stdout) { is_expected.to match(/Connection refused/) }
```
ehelms, line 93: If this passed previously, I would expect this to now fail. Be interesting to see waht the results are.

wbclark: I think lack of cleanup / test ordering is impacting the results on GH.

When I run it on EL8 Vagrant image locally:

```
Finished in 5 minutes 25 seconds (files took 1 minute 34.24 seconds to load)
62 examples, 0 failures
```

My theory is that on GH, Redis is left installed and the service is running, but the connection isn't configured in settings.py

ekohl: You should indeed ensure Redis is absent before starting the test. The order is not really defined that well and you can't assume it. Another example is where you rerun tests manually.

ehelms:
Needs a rebase

ekohl (commented):
The `/pulp/api/v3/status/` endpoint also contains a key `redis_connection`. Perhaps that's better than to call `rq`. In https://github.com/pulp/pulpcore/commit/8ab8afb45675cc9c420eb1872c0dfcfd973d657b I changed its content to be optional, depending on whether Redis is used but right now in 3.14 it's always there. On line 54 we already query it so you can expand it to check the content.

Perhaps that's an even better thing to query for more things anyway.

ehelms:
Needs a rebase

PR #216, opened 2021-07-13
Contain Redis within database

ekohl:
Without this, the Redis class may not be complete. That means Redis doesn't have to be running by the time the services are started.

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -27,6 +27,6 @@ class pulpcore::database(
     require     => Pulpcore::Admin['migrate --noinput'],
   }
 
-  include redis
+  contain redis
 
 }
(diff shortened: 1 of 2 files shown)


evgeni:
While I think this is a good change, I don't think it will fix the issue.

The error I've been seeing is *after* the installer has finished (and I'd expect at that point Redis to run, regardless of ordering in Puppet). The interesting thing is: I've not seen a redis service unit in the sosreport of the proxy at all :(

ekohl:
> While I think this is a good change, I don't think it will fix the issue.

You're right: this can fix issues but not the one we're running into.

ehelms (approved):

PR #220, opened 2021-07-13
Prefer the Redis unix socket if available

ekohl:
Unix sockets typically have lower overhead and also allows setting stricter permissions. While iptables can be used to limit access using users, file permissions are much easier to manage.

Includes https://github.com/theforeman/puppet-pulpcore/pull/218.

Currently a draft, but I'm exploring the option of deploying a dedicated Redis instance just for Pulp (if needed). This means Redis can be tuned different. Foreman needs Redis to be in persistent mode but Pulp doesn't always:

| Tasking | Caching | Redis mode |
|---------+---------+------------+
| RQ      | Yes     | Persistent |
| RQ      | No      | Persistent |
| PG      | Yes     | Empheral   |
| PG      | No      | Off        |

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -17,6 +17,15 @@ class pulpcore::config {
     ensure_newline => true,
   }
 
+  if $redis::unixsocket != '' {
+    $redis_url = "redis+unix://${redis::unixsocket}?db=${pulpcore::redis_db}"
+  } elsif $redis::port != 0 {
+    # TODO: this assumes $redis::bind at least has localhost in it
+    $redis_url = "redis://localhost:${redis::port}/${pulpcore::redis_db}"
+  } else {
+    fail('Unable to determine Redis URL')
+  }
+
   concat::fragment { 'base':
     target  => 'pulpcore settings',
     content => template('pulpcore/settings.py.erb'),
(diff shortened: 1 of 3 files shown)


ehelms:
> Currently a draft, but I'm exploring the option of deploying a dedicated Redis instance just for Pulp (if needed). This means Redis can be tuned different. Foreman needs Redis to be in persistent mode but Pulp doesn't always:

Since Foreman may also want to do caching potentially with a Redis, would we be looking at:

 * 3 instances, one for Dynflow, one for Pulp cache, one for Foreman cache
 * 2 instances, one for persistent workloads (e.g. Dynflow) and one for caching (Foreman or Pulp)

And does that change any of the design?

manifests/config.pp
```diff
@@ -17,6 +17,15 @@
     ensure_newline => true,
   }
 
+  if $redis::unixsocket != '' {
```
ehelms, line 20: What would be configuring this given we only `include redis` ?

ekohl: Technically it could have been changed via hiera. Outside of the installer this can also happen directly via manifests.

ekohl:
> Since Foreman may also want to do caching potentially with a Redis, would we be looking at:
> 
>     * 3 instances, one for Dynflow, one for Pulp cache, one for Foreman cache
> 
>     * 2 instances, one for persistent workloads (e.g. Dynflow) and one for caching (Foreman or Pulp)
> 
> 
> And does that change any of the design?

Possibly. At this point I want to find out if we can even use a unix socket in the first place. We may run into problems like permissions, maybe something inside Pulp itself and probably SELinux policies.

At the same time, should start to investigate if Foreman and Dynflow can.

ekohl:
Rebased, but it's now split into two commits. First there's https://github.com/theforeman/puppet-pulpcore/pull/331 which makes the Redis URL configurable (which is useful for remote Redis servers). Then this expands it to use the unix socket if available.

PR #222, opened 2021-07-19
Release 5.0.0

ehelms:
Potential open PRs:

~~- [ ] https://github.com/theforeman/puppet-pulpcore/pull/210~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## [5.0.0](https://github.com/theforeman/puppet-pulpcore/tree/5.0.0) (2021-07-27)
+
+[Full Changelog](https://github.com/theforeman/puppet-pulpcore/compare/4.0.1...5.0.0)
+
+**Breaking changes:**
+
+- Drop Puppet 5 support [\#197](https://github.com/theforeman/puppet-pulpcore/pull/197) ([ekohl](https://github.com/ekohl))
+
+**Implemented enhancements:**
+
+- Use REDIS\_URL as a configuration option [\#218](https://github.com/theforeman/puppet-pulpcore/pull/218) ([ekohl](https://github.com/ekohl))
+
+**Fixed bugs:**
+
+- Contain Redis within database [\#216](https://github.com/theforeman/puppet-pulpcore/pull/216) ([ekohl](https://github.com/ekohl))
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -44,6 +60,10 @@
 
 - Fixes [\#32622](https://projects.theforeman.org/issues/32622): Include StdEnvVars, ExportCertData SSL options in Apache [\#193](https://github.com/theforeman/puppet-pulpcore/pull/193) ([ehelms](https://github.com/ehelms))
 
+## [2.2.3](https://github.com/theforeman/puppet-pulpcore/tree/2.2.3) (2021-05-19)
+
+[Full Changelog](https://github.com/theforeman/puppet-pulpcore/compare/3.3.0...2.2.3)
```
ekohl, line 65: Should we skip this?

wbclark (commented):
I am still hoping to get #210 merged before this release.

As it is currently written, it simply works which you can verify by running the unit test, or using the basic acceptance test or any puppet apply and observe the configuration file contents.

This adds value now by warning the user not to make manual edits to `settings.py`, and it's particularly valuable to include that change now while we are adding features like tasking system choice and content caching, which are neither simple to configure nor particularly easy to find information about, and could easily break the installation with any manual edits.

In the future I'd like to expand that concept out into a separate puppet module for powerful, flexible, file headers that are reusable across multiple projects but can be overridden in any particular one. We can then use that across our modules to provide targeted hints and documentation links, and it should be generally useful for configuration file headers and particularly when it makes sense to customize or override the header when modules are nested.

I see that as something that could be done later which is not precluded by adding the feature as-is today, in the same way that we have previously written specialized logic directly in these modules to serve a need in the short term, before taking it further upstream e.g. voxpupuli-acceptance.

I understand however that the hesitance could be due to the complexity of some of the logic as well as limited time to review. For that reason, I'll add a few additional tests that should lend some confidence about the robustness of the solution.

wbclark (commented):
The changelog here otherwise looks good.

If it's absolutely necessary to move forward without #210 , I'd like to at least consider whether some less fully-features short-term alternative can be included.

ekohl (approved):

ehelms:
I am going to go ahead and release this. We can always pick up the additional features.

PR #223, opened 2021-07-26
Add warning not to directly edit settings.py

wbclark:
Short term static alternative to #210


templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
+################################################################################
+# !!! WARNING: DO NOT EDIT THIS FILE !!!                                       #
+################################################################################
+# File managed by Puppet.                                                      #
+# Module: pulpcore                                                             #
```
ekohl, line 5:
```suggestion
# Module: <%= scope['module_name'] %>                                                             #
```

templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
...
+################################################################################
+# Not only are your edits likely to be overwritten, there is also a strong     #
+# possibility of breaking your system if you change configuration here without #
+# making required changes elsewhere. Refer to the documentation you used to    #
+# install Pulpcore to determine the safe and persistent way to modify the      #
+# configuration.                                                               #
+################################################################################
```
ekohl, line 12: Is this part really needed? The user may not have explicitly thought about it so how are they going to link it back?

ekohl (commented):
I think this is a very long header and I generally dislike those. In puppet-foreman we have https://github.com/theforeman/puppet-foreman/blob/master/templates/_header.erb and IMHO that's plenty long.

templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
...
+# !!! WARNING: DO NOT EDIT THIS FILE !!!                                       #
+################################################################################
+# File managed by Puppet.                                                      #
+# Module: pulpcore                                                             #
+################################################################################
+# Not only are your edits likely to be overwritten, there is also a strong     #
+# possibility of breaking your system if you change configuration here without #
+# making required changes elsewhere. Refer to the documentation you used to    #
+# install Pulpcore to determine the safe and persistent way to modify the      #
+# configuration.                                                               #
```
ehelms, line 11: My only question is if we should avoid `you` pronouns here. So `Not only are edits likely`, `possibility of breaking the system`, `if changes are made`, `Refer to the documentation used to install`

templates/settings.py.erb
wbclark: Would `$module_name` ever take on another value? If not, then why not simply hard code it?

Note in the generic version I intend to build, it would be dynamic so that it could be used by other modules. That would also have an overrideable `$installer_name` so that it can say, e.g. `File managed by foreman-installer` one line above.

templates/settings.py.erb
wbclark: I think it's important to provide some explanation to the user so that they understand why they are asked to not make edits.

If there is a warning not to make edits, but no explanation, a confident user might say "oh, I already understand it won't be persistent, so this warning doesn't apply to me. I just want to try out that other tasking system..."

I agree it's difficult to phrase it in such a way that is equally applicable to users of puppet, foreman-installer, satellite-installer, orcharhino, etc. Necessarily, the explanation will be vague, and this is the problem that I intend to solve by allowing custom strings in hiera.

But in the meantime, with the upcoming release introducing several new config file entries that can easily be broken in unexpected ways, I argue that a vague explanation is still better than no explanation at all.

templates/settings.py.erb
wbclark: Updated to use passive voice consistently / remove 2nd person pronouns. Thanks

templates/settings.py.erb
ehelms: Nah, I think it will always be the module name based on the metadata and thus no worry that it might change for right now.

ehelms (approved):
I think this is a fair start, and the systematic change being worked on will be a better long term solution. For now this achieves the primary objective of helping users understand they should not be editing. I suppose we will learn over time if it actually works.

templates/settings.py.erb
ekohl: However, if you do want to go to a generic header it makes sense to think about it. For example, the closing `#` that you add now becomes a lot harder. That's why I'd advise to leave that part out. I should have mentioned that in my initial review.

My reasoning is that a change now will lead to a service restart. Restarting all services just for a new comment in a config file is a bit of a waste. So I'd prefer to get it right the first time.

ehelms: I think I see what you are saying... the trailing `#` is harder to calculate than if there was none. So if that part was dropped, when the generic solution is implemented it matches this closer and is easier to implement. We do not necessarily need the aesthetic.

wbclark: Understood. With that said, I already solved that issue in my proof of concept for the more general case:

https://github.com/theforeman/puppet-pulpcore/pull/210/files#diff-893d976ae6fa01e560858f0185fdcfad47f8550e71e8a510fc7a8135ca8e03e6R16

templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
+################################################################################
+# !!! WARNING: DO NOT EDIT THIS FILE !!!
+################################################################################
```
ekohl, line 3: I'd drop this line. The "file managed by Puppet" is actually the explanation for the warning.
```suggestion
```

templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
...
+# File managed by Puppet.
+# Module: <%= scope['module_name'] %>
```
ekohl, line 5: Can we combine these two?
```suggestion
# File managed by Puppet module: <%= scope['module_name'] %>
```

templates/settings.py.erb
```diff
@@ -1,3 +1,16 @@
...
+################################################################################
+# Not only are edits likely to be overwritten, there is also a strong
```
ekohl, line 7: They're not just likely, I think we can safely say they will be overwritten at some point.

wbclark:
> If that's your goal, why not copy what we have in dhcp?

Thanks, that is neat.

Quite a bit simpler, if not as robust, as what I had in https://github.com/theforeman/puppet-pulpcore/pull/210

PR #226, opened 2021-09-09
Fixes #33445: Increase default API gunicorn worker count

ehelms:
This increases the default API gunicorn worker count to be similar
to that of content worker count but with a smaller minimum. There
are different workloads that can require higher throughput through
the API and the current default of 1 is too low of a starting point.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -215,7 +215,7 @@ class pulpcore (
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
-  Integer[0] $api_service_worker_count = 1,
+  Integer[0] $api_service_worker_count = (2*min(4, $facts['processors']['count']) + 1),
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
   Hash[String[1], String[1]] $api_client_auth_cn_map = {},


manifests/init.pp
```diff
@@ -215,7 +215,7 @@
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
-  Integer[0] $api_service_worker_count = 1,
+  Integer[0] $api_service_worker_count = (2*min(4, $facts['processors']['count']) + 1),
```
jlsherrill, line 218: wouldn't that default to 9 or 10  as a minimum (2* 4 + 1)?  i think we probably want to drop the 2*, and just have:

min(4, $facts['processors']['count'])

although maybe we should set a max on it too (but maybe thats another bug as content_service_worker_count doesn'th ave a max.  Maybe don't worry about that here

wbclark (commented):
We document the dynamic minimum as default for the worker count parameter:

```
# @param worker_count
#   Number of pulpcore workers. Defaults to 8 or the number of CPU cores, whichever is smaller. Enabling more than 8 workers, even with additional CPU cores
```

Whatever default value we settle on for here `$api_service_worker_count` should be similarly documented, i.e. with Justin's recommendation something like:

```
# @param api_service_worker_count
#   Number of pulpcore-api service workers for gunicorn to use. Defaults to 4 or the number of CPU cores, whichever is smaller.
#   Modifying this parameter should be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
```

wbclark:
From the Redmine: "Generally this is fine, but we've seen some conditions where this isn't enough and can lead to errors."

I am definitely curious to know what conditions those are if you can provide some more info!

manifests/init.pp
ehelms: This was inspired by the content worker count calculation, and that comes from this https://docs.gunicorn.org/en/stable/design.html#how-many-workers so for what our default is here, I am not sure what to apply either. But note this is a default and users can tune this to their needs if need be.

So I suppose the question is still, what do we think our default should be and should it be dynamic to allow for user system resources to scale this value as we typically associate system resources with scale of deployment.

jlsherrill: I think it should scale with the # of processors, i was more stating that i think 2x is probably too much and a more 1 to 1 scaling probably makes more sense

jlsherrill:
@wbclark there have been a couple cases:

1.  a bz or 2 where connections to pulp timed out when a lot was going on and there was a lot of communication to the pulp server, it was not reproducible and was closed.  I can probably dig it up if you want to see it
2. @ianballou was working on a pr recently that uncovered a performance issue in pulp, in that a request to copy content to a new content view version was slow and when we did multiple at once it resulted in a timeout.  This error is being fixed by the pulp team, but given this is the 2nd or 3rd time i've seen this it seemed worth to bump up the worker count

jlsherrill:
Here's one such bz: https://bugzilla.redhat.com/show_bug.cgi?id=1995920

manifests/init.pp
evgeni: I would like to raise the fact that such computations are often for "this machine just runs my service", which is not true for most of the pulpcore installs we're looking at. We even have another `gunicorn` cluster next to this one for the content service.

So yeah, I would also probably look at a lower starting value, where `min(4, $facts['processors']['count']) + 1` sounds reasonable?

jlsherrill: since ehelms is out, i opened a new version here:  https://github.com/theforeman/puppet-pulpcore/pull/231

ekohl:
We merged https://github.com/theforeman/puppet-pulpcore/pull/231.

PR #229, opened 2021-09-29
DRAFT: #33446 with alternate test style

wbclark:
Continuing work from https://github.com/theforeman/puppet-pulpcore/pull/227

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -133,7 +133,8 @@
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
 # @param worker_ttl
-#   The number of seconds before a pulpcore worker should be considered lost.
+#   The number of seconds before a pulpcore worker should be considered lost. If undefined, pulpcore will use its default setting which is 30 seconds as of
+#   pulpcore version 3.14. You should not need to modify this setting unless the application reports workers timing out while they are busy completing tasks.
 #
 # @param use_rq_tasking_system
 #   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -132,6 +132,9 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param worker_ttl
+#   The number of seconds before a pulpcore worker should be considered lost.
+#
 # @param use_rq_tasking_system
 #   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
 #   Any benchmarking you did to optimize worker_count or other tasking related parameters will no longer be accurate after changing the tasking system.
@@ -211,6 +214,7 @@ class pulpcore (
   Pulpcore::ChecksumTypes $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
   String[1] $remote_user_environ_name = 'HTTP_REMOTE_USER',
   Integer[0] $worker_count = min(8, $facts['processors']['count']),
+  Optional[Integer[0]] $worker_ttl = undef,
   Boolean $use_rq_tasking_system = false,
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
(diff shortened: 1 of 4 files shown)


wbclark:
NOTE: I updated the parameter documentation to be clearer for users, and also tried a new style for the tests which should make it easier for developers to interpret test output.

spec/acceptance/basic_spec.rb
```diff
@@ -74,9 +74,20 @@ class { 'pulpcore':
     its(:exit_status) { is_expected.to eq 0 }
   end
 
-  describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
-    its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = True/) }
-    its(:exit_status) { is_expected.to eq 0 }
+  describe "Pulpcore Settings" do
+    let(:settings) { command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") }
```
wbclark, line 78: As written, this *works* and should give a more descriptive output in the case of test failures (I will break a test shortly to confirm).

That said, I want to make sure really I understand correctly how this behaves -- would the block argument here get evaluated in the `let` block, or does it later get evaluated three times when calling `settings.exit_status`, `settings.stdout`, etc, causing inefficient code execution?

What I'm trying to do may or may not still be possible with `describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do` instead. I am not sure if `its(:stdout) "will a string work here?" do` would function as I intend it do.

From looking at a lot of rspec examples, `expect(...)` seems to be preferred to `it { is_expected.to ... }` which is why I decided to try this. That said, maybe it is proper rspec style in general but not right for serverspec specifically.

wbclark (commented):
Thinking about this some more, what I really want to do is deliver a version of this PR which delivers the feature as needed, adhering to the current test style though it may be flawed.

I thought the tests would work in this style, and indeed they do, but I'm simply not confident enough yet with it to rule out drawbacks.

Therefore I will rename this PR and mark it as a draft, and shortly open another version that delivers the same functionality with the current style tests.

wbclark:
OK, the real version of this is https://github.com/theforeman/puppet-pulpcore/pull/230

spec/acceptance/basic_spec.rb
evgeni: I like the new grouping of things, and I think `let` should be evaluated once.

As for `expect` vs `is_expected.to` -- we use the later *everywhere*, so I'd prefer to keep that for consistency (and if change, change globally).

manifests/init.pp
```diff
@@ -132,6 +132,10 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param worker_ttl
+#   The number of seconds before a pulpcore worker should be considered lost. If undefined, pulpcore will use its default setting which is 30 seconds as of
+#   pulpcore version 3.14. You should not need to modify this setting unless the application reports workers timing out while they are busy completing tasks.
```
evgeni, line 137: Reading "You should not need to modify this setting unless the application reports workers timing out while they are busy completing tasks." one could think "Yeah, let's set it to 9000 then!", which is also not what we want the users to do.

For `worker_count` we say " Modifying this parameter should be done incrementally with benchmarking at each step to determine an optimal value for your deployment.", should something similar be added here?

spec/acceptance/basic_spec.rb
ekohl: What you can do is:
```ruby
describe 'Pulpcore settings' do
  subject { command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") }

end
```

This is effectively the same as what was there before.

wbclark: I think I recall trying something like this in the past which failed:

```ruby
describe "Pulpcore settings" do
  subject { command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") }

  its(:exit_status) "exits without error" { is_expected.to eq 0 }
end
```

I will try again to confirm

wbclark: The issue was

```
Failure/Error: __send__(method, file)

SyntaxError:
  /home/wclark/Projects/puppet-pulpcore/spec/acceptance/basic_spec.rb:80: syntax error, unexpected string literal, expecting `end'
      its(:exit_status) "retrieves pulpcore settings" ...
```

So `it` has an optional string argument before the block argument, but `its(:symbol)` can receive only the block.

PR #230, opened 2021-09-30
Fixes #33446 - Allow configuring Pulpcore worker timeout

wbclark:


spec/classes/pulpcore_spec.rb
```diff
@@ -513,6 +513,17 @@
...
       end
+
+      context 'with parameter worker_ttl = 60' do
+        let :params do
+          { worker_ttl: 60 }
+        end
+
+        it 'configures pulpcore with setting WORKER_TTL = 60' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{\sWORKER_TTL = 60})
```
ekohl, line 524:
```suggestion
            .with_content(%r{^WORKER_TTL = 60$})
```

ekohl (commented):
As noted in the offline chat: I'd keep init.pp small and rather add more instructions to ` README.md` about how to use `diffsettings`.

spec/acceptance/basic_spec.rb
```diff
@@ -76,6 +76,7 @@ class { 'pulpcore':
 
   describe command("PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings") do
     its(:stdout) { is_expected.to match(/^USE_NEW_WORKER_TYPE = True/) }
+    its(:stdout) { is_expected.to match(/^WORKER_TTL = 30/) }
```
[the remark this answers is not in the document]
wbclark: The point of the test was to catch when Pulp's default changes so that Pulp's default in the param doc could be updated. If we don't document Pulp's default then there is no point to the test, so I'll remove it.

templates/settings.py.erb
```diff
@@ -23,6 +23,10 @@ REDIS_URL = "redis://localhost:<%= scope['redis::port'] %>/<%= scope['pulpcore::
 
 USE_NEW_WORKER_TYPE = <%= scope['pulpcore::use_rq_tasking_system'] ? "False" : "True" %>
 
+<% if scope['pulpcore::worker_ttl'] -%>
+WORKER_TTL = <%= scope['pulpcore::worker_ttl'] %>
+<% end -%>
+
```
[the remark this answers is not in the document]
wbclark: Thanks. Looking at it a bit more, arguably the newline above should be removed so that `USE_NEW_WORKER_TYPE` and `WORKER_TTL` exist in the same block. I'll take that approach.

wbclark:
@ekohl @evgeni this is now updated with all feedback from this and the other PR.

wbclark:
force pushing to trigger CI

manifests/init.pp
```diff
@@ -132,6 +132,12 @@
 #   available, likely results in performance degradation due to I/O blocking and is not recommended in most cases. Modifying this parameter should
 #   be done incrementally with benchmarking at each step to determine an optimal value for your deployment.
 #
+# @param worker_ttl
+#   The number of seconds before a pulpcore worker should be considered lost. If undefined, pulpcore will use its default setting.
```
ekohl, line 136: I would be very explicit in that it's a setting and which setting it is.
```suggestion
#   Configure Pulpcore's WORKER_TTL setting. This is the number of seconds before a Pulpcore worker should be considered lost. If undefined, Pulpcore will use its default setting.
```

README.md
```diff
@@ -45,6 +45,22 @@ This results into the following structure, using `tree -pug`:
...
+PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings
+```
+
+For example, to check the current value of a Pulpcore setting such as `WORKER_TTL`:
+
+```shell
+PULP_SETTINGS=/etc/pulp/settings.py pulpcore-manager diffsettings | grep WORKER_TTL
+```
+
+This is useful in cases such as this module's `$worker_ttl` parameter which configures `WORKER_TTL` but is `undef` by default, in which case `WORKER_TTL` is omitted from `/etc/pulp/settings.py` and therefore Pulpcore's default is used.
```
ekohl, line 62: This is a very long sentence. How about something like:
```suggestion
This is useful in cases where parameters that configure settings have an `undef` default, such as this `$worker_ttl`. By default it's omitted from `settings.py` and therefore Pulpcore's default is used.
```

spec/classes/pulpcore_spec.rb
```diff
@@ -513,6 +513,17 @@
...
       end
+
+      context 'with parameter worker_ttl = 60' do
+        let :params do
+          { worker_ttl: 60 }
+        end
+
+        it 'configures pulpcore with setting WORKER_TTL = 60' do
+          is_expected.to contain_concat__fragment('base')
+            .with_content(%r{^WORKER_TTL = 60$})
```
ekohl, line 524: Please add the opposite of this to the base test. So around line 29 you should add `.without_content(%r{WORKER_TTL})`

ekohl (changes requested):

manifests/init.pp
wbclark: Great change, TY

README.md
wbclark: I like it -- it reads a lot more naturally.

wbclark:
@ekohl thank you for the feedback, I updated with the requested changes.

While the tests passed for me locally, for some reason CI has not run on my recent pushes. Any idea why that is?

ekohl (approved):
No, I wanted to wait for CI to run.

PR #231, opened 2021-10-04
Fixes #33445: Increase default API gunicorn worker count

jlsherrill:
This increases the default API gunicorn worker count to be similar
to that of content worker count but with a smaller minimum. There
are different workloads that can require higher throughput through
the API and the current default of 1 is too low of a starting point.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -215,7 +215,7 @@ class pulpcore (
   Boolean $service_enable = true,
   Boolean $service_ensure = true,
   Integer[0] $content_service_worker_count = (2*min(8, $facts['processors']['count']) + 1),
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
:+1: for this in general, but could you this line:
https://github.com/theforeman/puppet-pulpcore/blob/13c10a32989eccfc7c8b5b09045314b2360e6e3b/spec/classes/pulpcore_spec.rb#L128
```ruby
is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{-w 2})
```

That should verify that when there's 1 processor (see line 6) it has 2 workers.

jlsherrill:
updated with test!

ekohl (approved):

PR #234, opened 2021-10-20
notify the socket service when the DB changes

evgeni:


manifests/database.pp
```diff
@@ -13,6 +13,8 @@
       locale   => 'en_US.utf8',
       before   => Pulpcore::Admin['migrate --noinput'],
     }
+
+    Class['postgresql::server::service'] ~> Pulpcore::Socket_service['pulpcore-content']
```
ekohl, line 17: I'd prefer to only restart the `.service` part and not the .socket part.
```suggestion
    Class['postgresql::server::service'] ~> Service['pulpcore-content.service']
```

PR #235, opened 2021-10-20
Fixes #33733 - generate key for db encryption

jlsherrill:


manifests/config.pp
```diff
@@ -51,4 +51,17 @@
     mode   => '0770',
   }
 
+  exec { 'Create database symmetric key':
+    path    => ['/bin', '/usr/bin'],
+    command => "openssl rand -base64 32    | tr '+/' '-_' > /etc/pulp/certs/database_fields.symmetric.key",
```
evgeni, line 56: This should use the variable for the key path

manifests/init.pp
```diff
@@ -232,6 +232,8 @@
   Enum['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG'] $log_level = 'INFO',
 ) {
   $settings_file = "${config_dir}/settings.py"
+  $certs_dir = "${config_dir}/certs/"
```
ekohl, line 235: I don't think it's a problem, but usually we don't have trailing slashes. It also results in a double forward slash in the path
```suggestion
  $certs_dir = "${config_dir}/certs"
```

wbclark:
I previously took a stab at this with https://github.com/theforeman/puppet-pulpcore/pull/205 , using a puppet function to wrap Ruby's `SecureRandom.urlsafe_base64(32)` rather than relying on an exec and openssl. Longer term if we want to avoid the exec, that idea may be worth revisiting.

wbclark:
Does the file resource require `show_diff => false` ? Since it manages only attributes other than the file contents

Is the key visible in puppet logs when the exec resource is evaluated?

PR #238, opened 2021-10-21
Fixes #33766 - Support Pulpcore 3.15

ekohl:


manifests/plugin/certguard.pp
```diff
@@ -1,5 +1,6 @@
 # @summary Pulp Certguard plugin
 class pulpcore::plugin::certguard {
+  # TODO: this breaks - b79786c561ea56054c03fe83527cb79db84c295a should have addressed this
```
ekohl, line 3: @evgeni said on IRC that certguard in 3.15 depends on python3(8)-rhsm now so this should no longer be needed. The question is: do we still need to ensure it's installed on Pulpcore 3.14?

manifests/init.pp
```diff
@@ -140,7 +140,7 @@
 #   Modification should be performed incrementally to determine the least value that prevents false positive worker timeouts.
 #
 # @param use_rq_tasking_system
-#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14.
+#   Use the older RQ workers tasking system instead of the newer PostgreSQL tasking system introduced in Pulpcore 3.14. This is deprecated in 3.15.
```
ekohl, line 143: And will be removed in 3.16.

ekohl (commented):
I hear 3.16 is being worked on as well, but 3.15 is built today and is SCL'ed. On EL8 it also uses modules for Python 3.8. Having this PR allows us to verify we actually work.

manifests/plugin/certguard.pp
```diff
@@ -1,8 +1,14 @@
 # @summary Pulp Certguard plugin
 class pulpcore::plugin::certguard {
-  package { 'python3-subscription-manager-rhsm':
-    ensure => present,
+  # TODO: this breaks 3.15 - b79786c561ea56054c03fe83527cb79db84c295a should have addressed this
+  # The pulpcore_version fact is not always available - needs a better solution.
+  if fact('pulpcore_version') == '3.14' {
```
ehelms, line 5: This walling off will need to be reflected in the unit tests -- https://github.com/theforeman/puppet-pulpcore/blob/master/spec/classes/plugin_certguard_spec.rb#L9

PR #241, opened 2021-11-04
Clean up HTTPS Vhost when no longer used

wbclark:
https://github.com/theforeman/puppet-foreman_proxy_content/pull/390 transitions to deploying the Pulpcore Vhost within puppet-FPC

This PR ensures the Vhost that was previously configured here is cleaned up on upgrades

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -137,6 +137,13 @@ class pulpcore::apache (
           'directories' => [$content_directory, $api_directory],
           'proxy_pass'  => [$proxy_pass_static],
       })
+
+      include apache
+      apache::vhost { 'pulpcore-https':
+        ensure   => 'absent',
+        priority => '10',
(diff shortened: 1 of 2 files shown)


wbclark:
/sigh

TIL `rake validate` does not include `rake lint`

ekohl (commented):
Normally we rely on Apache to purge configs. I'm not sure if we want to support this. If we do, then we should also purge the HTTP vhost.

wbclark:
> Normally we rely on Apache to purge configs.

Interesting. Could you point me towards where I could read about this?

ekohl:
https://forge.puppet.com/modules/puppetlabs/apache/reference#purge_configs & https://forge.puppet.com/modules/puppetlabs/apache/reference#purge_vhost_dir

wbclark:
`purge_configs` looks `true` by default so I'll close this as it should not be needed. Thanks!

PR #242, opened 2022-01-21
Fixes #34298 - support KEEP_CHANGELOG_LIMIT option

wbclark:
This was added to Pulpcore with https://github.com/pulp/pulp_rpm/pull/2333

Backport to 3.14 is underway

### DIFF
--- a/manifests/plugin/rpm.pp
+++ b/manifests/plugin/rpm.pp
@@ -1,8 +1,14 @@
 # @summary Pulp RPM plugin
 # @param use_pulp2_content_route
 #   Whether to redirect the legacy (Pulp 2) URLs to the content server
+#
+# @param keep_changelog_limit
+#   Pulpcore's KEEP_CHANGELOG_LIMIT setting. Uses Pulpcore's default when
+#   undefined. Increasing this limit will cause pulpcore workers to use more
+#   memory when more changelogs are available in the repo metadata.
 class pulpcore::plugin::rpm (
   Boolean $use_pulp2_content_route = false,
+  Optional[Integer[0]] $keep_changelog_limit = undef,
 ) {
   if $use_pulp2_content_route {
     $context = {
@@ -28,8 +34,15 @@ class pulpcore::plugin::rpm (
     $content = undef
   }
 
+  if $keep_changelog_limit {
+    $rpm_plugin_config = "KEEP_CHANGELOG_LIMIT = ${keep_changelog_limit}"
+  } else {
+    $rpm_plugin_config = undef
+  }
+
   pulpcore::plugin { 'rpm':
     http_content  => $content,
     https_content => $content,
+    config        => $rpm_plugin_config,
   }
 }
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -216,6 +219,7 @@
   Array[Stdlib::Absolutepath] $allowed_import_path = ['/var/lib/pulp/sync_imports'],
   Array[Stdlib::Absolutepath] $allowed_export_path = [],
   Pulpcore::ChecksumTypes $allowed_content_checksums = ['sha224', 'sha256', 'sha384', 'sha512'],
+  Integer[1] $keep_changelog_limit = undef,
```
ekohl, line 222: This is not a valid type. It should be
```suggestion
  Optional[Integer[1]] $keep_changelog_limit = undef,
```

That said, the feature is explicit in that `0` should be the lowest value. See https://github.com/pulp/pulp_rpm/commit/cbc5e98770da6469bed936070b4f1c07fb5835c8#

ekohl (changes requested):
This feature belongs to the RPM plugin, so everything should be happening in [pulpcore::plugin::rpm](https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/rpm.pp).

Here's an example where we set config for a plugin:
https://github.com/theforeman/puppet-pulpcore/blob/d347abdac8091ed725a532427c3aa380a63bcbc9/manifests/plugin/container.pp#L37

wbclark:
I pushed a commit which addresses all feedback and moves the feature to pulpcore::plugin::rpm

Will squash it after running all tests

manifests/plugin/rpm.pp
```diff
@@ -1,8 +1,12 @@
 # @summary Pulp RPM plugin
 # @param use_pulp2_content_route
 #   Whether to redirect the legacy (Pulp 2) URLs to the content server
+#
+# @param keep_changelog_limit
+#   Pulpcore's KEEP_CHANGELOG_LIMIT setting. Uses Pulpcore's default when undefined. Increasing this limit will cause pulpcore workers to use more memory when more changelogs are available in the repo metadata.
```
ekohl, line 6: I usually limit the line length, like this:
https://github.com/theforeman/puppet-pulpcore/blob/d347abdac8091ed725a532427c3aa380a63bcbc9/manifests/init.pp#L30-L33

I use vim to do that (shift v -> gq wraps).

ekohl (commented):
Very small nits, but this is indeed what I expected. Please address them and squash the commit.

One thing that we can't do anything about, but I wonder if the Pulp setting should have been prefixed so it's `RPM_KEEP_CHANGELOG_LIMIT`. Or will pulp_deb also respect it? Just thinking out loud now.

manifests/plugin/rpm.pp
wbclark: Hugely useful. Thanks!

ekohl (approved):
Thanks!

PR #244, opened 2022-02-02
Fixes #34379 - Create the Pulp group as a system group

ekohl:
Similar to 5a7991ab90a25aa3add176c1714fc416053a007a but for the group instead of the user.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -24,5 +24,6 @@ class pulpcore::install {
 
   group { $pulpcore::group:
     ensure => present,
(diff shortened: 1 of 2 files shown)


ekohl:
I wanted to verify this but couldn't find a good way to specify `uid < 1000` so I opened https://github.com/mizzy/serverspec/pull/617.

ekohl:
Since that was merged and released, I've written some tests.

evgeni (approved):

PR #250, opened 2022-03-16
Fixes #34631 - Recurse into artifacts folder to ensure owner is proper

ianballou:
The artifacts folder can have the wrong owner (apache) if a user's system configuration allowed for Pulp 2 to create hardlinks during the migration process.  This change tells the installer to recurse into the artifacts folder to correct the artifact file ownership. I checked with the pulp team and 750 should be just fine for artifacts.

To test:
1) On Katello 4.3+, run `chown -hR apache.pulp /var/lib/pulp/media/artifact`
2) Try to complete sync a repo. It will fail.
3) Run the installer with this change. Syncing should now work.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -30,11 +30,19 @@ class pulpcore::config {
     mode   => '0775',
   }
 
-  file { [$pulpcore::cache_dir, $pulpcore::media_root]:
-    ensure => directory,
-    owner  => $pulpcore::user,
-    group  => $pulpcore::group,
-    mode   => '0750',
+  file { $pulpcore::cache_dir:
+    ensure  => directory,
+    owner   => $pulpcore::user,
+    group   => $pulpcore::group,
+    mode    => '0750',
+  }
+
+  file { $pulpcore::media_root:
+    ensure  => directory,
+    owner   => $pulpcore::user,
+    group   => $pulpcore::group,
+    mode    => '0750',
+    recurse => true,
   }
 
   file { $pulpcore::allowed_import_path:


evgeni:
I am concerned about the performance impact here. The pulp content storage can become rather big and defining it here with recurse true, means that every puppet run will need to scan *all* files and check their permissions/ownership.

Additionally, creating files as 750 means they are executable, which *can* be a security concern (if there would be a bug in pulp core that would allow execution of existing files, one could upload a specially crafted artifact first, and then execute it).

ekohl (changes requested):
> I am concerned about the performance impact here. The pulp content storage can become rather big and defining it here with recurse true, means that every puppet run will need to scan all files and check their permissions/ownership.

This. It's very expensive and generally speaking a bad idea. At most this is a workaround. It is much better to find the reason why these files were created with incorrect permissions.

Looking at the linked resources, this stems from the problem where Pulp 2 runs as the Apache user. That means it may be better to solve this in the final conversion from Pulp 2 to Pulp 3. Once a user is on Pulp 3 only, this should never happen.

> Additionally, creating files as 750 means they are executable, which can be a security concern (if there would be a bug in pulp core that would allow execution of existing files, one could upload a specially crafted artifact first, and then execute it).

AFAIK you can use mode `0640` and Puppet will be smart enough to make that `0750` on directories.

ianballou:
Okay, so perhaps this should go in foreman maintain since it would be a one time thing. We could run it automatically too when users remove Pulp 2.

evgeni:
My memory tells me this was detected *after* the user has migrated to p3 and removed p2, thus the remove step wouldn't be executed again? At least not for Katello (I think satellite updates force the remove step these days)

Is there a good way to detect migrated setups? Then we could (in installer hooks or maintain) detect it, run a recursive chown once, lay down a flag that tells the next run "we're done here, carry on".

ekohl:
We have this hook: https://github.com/theforeman/foreman-installer/blob/develop/hooks/pre/33-pulpcore_assets_permissions.rb
Perhaps we need something similar for the media?

Also, in maintain there's this:
https://github.com/theforeman/foreman_maintain/blob/d031a7dcbed9518cf10de9d3f2b4cab869ae0c99/definitions/procedures/prep_6_10_upgrade.rb#L11-L21

Perhaps that's not executed in upstream?

jlsherrill:
> Also, in maintain there's this: https://github.com/theforeman/foreman_maintain/blob/d031a7dcbed9518cf10de9d3f2b4cab869ae0c99/definitions/procedures/prep_6_10_upgrade.rb#L11-L21
> 
> Perhaps that's not executed in upstream?
It is, but its a manual process:  https://theforeman.org/plugins/katello/3.18/upgrade/index.html

ianballou:
Here's the new PR for foreman-maintain: https://github.com/theforeman/foreman_maintain/pull/596

I'll switch the issue and BZ over to foreman-maintain as well.

ekohl:
That it happens even after Pulp 2 removal is concerning. Are people by any chance running `pulpcore-manager` as root instead of `pulp`?

ianballou:
@ekohl the issue has to do with Pulp 3 creating its hardlinks, Grant has some explanation here: https://community.theforeman.org/t/katello-4-3-repo-sync-error-errno1-operation-not-permitted/27262/14

I suppose it could also be that users migrated to Pulp 3 at a time when the permissions where not set correctly.

ianballou:
@ekohl I agree on all points there. Since the issue is fixed by a one-time `chown` command, I figured the best place for a fix script would be foreman-maintain.  It's easy for users to use and low-risk.

I personally haven't been able to reproduce the issue when trying upgrades.

ekohl:
I think a reproducer is:
* Run preparation step
* Sync content with Pulp 2 (make sure there is new content!)
* Switch over to Pulp 3
* Clean up Pulp 3

I suspect the problem is that the switch over to Pulp 3 does NOT do the chown while it should. Instead I suspect it's only in the preparation step.

ianballou:
I thought that would reproduce it too but no dice.  The prep-6.10-upgrade only does permissions stuff on `/var/lib/pulp/content` rather than `/var/lib/pulp/media`.  I think it that did reproduce it, more people would be having problems.  I changed file ownership to apache on Katello 4.1 for a test and it instantly caused sync errors.

PR #249, opened 2022-03-09
Drop Pulpcore 3.14 & 3.15, move to 3.16 and 3.17

ekohl:

PR #251, opened 2022-03-17
test that the API is correctly protected with cert auth

evgeni:


ekohl:
I included the test in https://github.com/theforeman/puppet-pulpcore/pull/249.

PR #252, opened 2022-03-28
Fixes #34684 - install pulp-cli

evgeni:


spec/acceptance/cli_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    its(:exit_status) { is_expected.to eq 0 }
+    its(:stdout) { is_expected.to match(/valid pulp-cli config/) }
+  end
+
+  describe command("REQUESTS_CA_BUNDLE=/etc/pulpcore-certs/ca-cert.pem pulp status") do
+    its(:exit_status) { is_expected.to eq 0 }
+    its(:stdout) { is_expected.to match(/versions/) }
+    # currently this contains a warning:
+    # SubjectAltNameWarning: Certificate for centos8-64.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818.
+    # its(:stderr) { is_expected.to eq '' }
```
evgeni, line 44: I couldn't find a good way to fiddle this code to make it work.
https://github.com/theforeman/puppet-pulpcore/blob/3dd9f3796a52b6599279ca409078755aa65780f8/spec/setup_acceptance_node.pp#L53-L58

There are a few hints here on SO: https://security.stackexchange.com/questions/74345/provide-subjectaltname-to-openssl-directly-on-the-command-line

evgeni:
EL7 currently fails, as there is no `/usr/bin/pulp` thanks to the SCL, https://github.com/theforeman/pulpcore-packaging/pull/425 is fixing it.

manifests/cli/params.pp
```diff
@@ -0,0 +1,12 @@
+# Parameters for Pulpcore CLI class
```
ekohl, line 1: You don't need a params file. It's fine to have them inline in `cli.pp`

manifests/cli.pp
```diff
@@ -0,0 +1,63 @@
...
+  Boolean $dry_run = $pulpcore::cli::params::dry_run,
+) inherits pulpcore::cli::params {
+  package { 'pulp-cli':
+    ensure => $version,
+  }
+  -> file { '/etc/pulp/cli.toml':
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0644',
```
ekohl, line 46: There's a password in there. Should it really be world readable?

spec/acceptance/cli_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+end
+
+describe 'installation of server with cli' do
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      include pulpcore
+      class { 'pulpcore::cli':
+        pulpcore_url => "https://#{host_inventory['fqdn']}/",
```
ekohl, line 24: Could also be written using native Puppet:
```suggestion
        pulpcore_url => "https://${facts['fqdn']}/",
```

manifests/cli.pp
evgeni: That's one of the things I wanted to talk about. This file *can* have a PW in it, but in the Katello scenario it won't -- we use certs there, and those are properly protected already (only `root` and members of `foreman` can read those).

So I'd prefer to leave that file readable as that'll make it easier for non-root-but-foreman-members to use it.

But not in the case when there is a PW in it :see_no_evil: 

One possibility is just not to allow user/passwords at all (as we don't need them in Katello). The other is making the mode "smart" (but I don't like smart code).

Or, well, bite the bullet and make the file stricter owned.

manifests/cli.pp
```diff
@@ -0,0 +1,63 @@
...
+  Optional[String] $cert = undef,
+  Optional[String] $key = undef,
+  Optional[String] $api_root = undef,
+  Boolean $verify_ssl = true,
+  Boolean $dry_run = true,
+) {
+  package { 'pulp-cli':
+    ensure => $version,
+  }
+  -> file { '/etc/pulp/cli.toml':
```
evgeni, line 42: Do we also want to manage `/root/.config/pulp/cli.toml` like we do for Hammer? Given the default to cert auth, I'd vote no?

manifests/cli.pp
evgeni: If I'm reading https://github.com/pulp/pulp-cli/blob/e7fe445e605bbcfc0a665ba1c53a42add719a191/pulpcore/cli/common/__init__.py#L49-L59 correctly, it does merge.

Maybe @mdellweg wants to chime in?

manifests/cli.pp
evgeni: The certs are `root:foreman`, so certainly usable by others than `root`, thus it felt "OK" to put them in the global config?

spec/acceptance/cli_spec.rb
ekohl: In the past I also struggled with this. In https://github.com/ekohl/ownca/commit/2034411c5819ac62a41bcc27ffe82f1e81ac9573 I reverted to the config because `-addext` isn't an add: it replaces all extensions from the config. However, there is no config here so I'm guessing it should work.

manifests/cli.pp
mdellweg: Yes, that is the "level 1 merge". If you specify
```toml
[cli]
username = "bob"
password = "not my password"
```
in one and
```toml
[cli]
password = "most secret"
```
in the other, the `[cli]` sections will be combined by overwriting individual settings.

manifests/cli.pp
evgeni: Turns out, if you specify both certs (in global config) and password (in user config) things explode. So I think I'll move the certs to the user config after all.

spec/acceptance/cli_spec.rb
ekohl: https://github.com/theforeman/puppet-pulpcore/pull/253

manifests/cli.pp
```diff
@@ -0,0 +1,92 @@
...
+#
+# $verify_ssl::         Verify SSL connection
+#
+# $dry_run::            Trace commands without performing any unsafe HTTP calls
+#
+# $version::            pulp-cli package version, it's passed to ensure parameter of package resource
+#                       can be set to specific version number, 'latest', 'present' etc.
+#
+class pulpcore::cli (
+  Optional[Stdlib::HTTPUrl] $pulpcore_url = undef,
```
evgeni, line 31: should this default to `https://${fqdn}`?

ekohl (commented):
This looks good. I do wonder how we're going to present this to users. One thing to consider is to automatically inherit from Foreman. So one solution is how `foreman::cli` does it: https://github.com/theforeman/puppet-foreman/blob/9e483426c511e7a7a2648c2eaba18f50ac7df8f9/manifests/cli.pp#L39-L50=

That's probably the easiest if we take Kafo into consideration. It should be noted that this approach is compile order dependent (sadly it does exist in Puppet and here you can notice it). That is why the order is defined in the installer config (https://github.com/theforeman/foreman-installer/blob/8d7fb8ef47ffaf1023ffe8ade3db581ed5b1b655/config/foreman.yaml#L34-L37=).

ehelms:
> I do wonder how we're going to present this to users.

That is a solid question, cause we view this as a bit of advanced functionality when it's deployed in a Katello scenario. I, for example, would be hesistent to make it easy for a user to configure the dry-run (https://github.com/theforeman/puppet-pulpcore/pull/252/files#diff-751b60988fac3d7c7c119b3a70b6f70d6ef0c3facfaf6b9c0c0d44fc564cacb1R25) parameter as this can have severe implications to the system as a whole opened up.

evgeni:
The Katello integration of Pulpcore happens via `foreman_proxy` and `foreman_proxy_content` modules, this module is not exposed to the users at all (so if they want to modify things, they *have* to use `custom-hiera`).

Now, the integration, I think, should be one of these modules including `pulpcore::cli` and setting the Katello-wanted parameters. Something like:

```puppet
class { 'pulpcore::cli':
    pulpcore_url => foreman_proxy::plugin::pulp::pulpcore_api_url,
    dry_run => true,
    cert => …,
    key => …,
}
````

File: manifests/cli.pp [hunk 7]
```diff
@@ -0,0 +1,92 @@
+# = Pulpcore command line interface
+#
+# This class installs the Pulpcore command line interface (CLI).
+#
+# === Parameters:
+#
+# $pulpcore_url::       URL on which Pulpcore runs
```
evgeni, line 7: in other places (`foreman_proxy`, `foreman_proxy_content`) we call this `pulpcore_api_url`.

pulp-cli itself calls it `base_url`.

:bike: shed?

evgeni:
Okay, why does `status` work but `user list` raise an SSL error?!

spec/acceptance/cli_spec.rb
```diff
@@ -0,0 +1,88 @@
+require 'spec_helper_acceptance'
+
+describe 'installation of server with cli' do
+  it_behaves_like 'an idempotent resource' do
+    let(:manifest) do
+      <<-PUPPET
+      include pulpcore
```
ehelms, line 7: If you want to use certificates for client auth, won't you also need to configure Pulp with SSL certificates so Apache is set up that way? https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/init.pp#L22-L30

manifests/cli.pp
```diff
@@ -0,0 +1,92 @@
...
+      {
+        base_url   => $pulpcore_url,
+        api_root   => $api_root,
+        verify_ssl => $verify_ssl,
+        dry_run    => $dry_run,
+      }
+    ),
+  }
+
+  if $manage_root_config and (($username and $password) or ($cert and $key)) {
```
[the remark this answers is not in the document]
evgeni: So in `pulp-cli` the validation is a bit more involved: https://github.com/pulp/pulp-cli/blob/aa50bc63e949e7f413bce6bc10d323d119e61b09/pulpcore/cli/common/openapi.py#L49-L62

I think for this module, it would be sufficient to check:
* (username OR password) AND (cert OR key) → error
* password AND !username → error
* key AND !cert → error

Everything else is imho valid (like setting username via config, but using CLI `--password` for the password, or providing a cert/key combo in the cert parameter).

spec/acceptance/cli_spec.rb
evgeni: Maybe? Don't look at me like I'd understand all nuances of this module when used outside fpc :stuck_out_tongue_closed_eyes:

evgeni: In retrospect, no, or "maybe, but…": I'm including the basic example, which does the same request via `curl` and there it works:

https://github.com/theforeman/puppet-pulpcore/blob/3dd9f3796a52b6599279ca409078755aa65780f8/spec/support/acceptance/examples.rb#L57-L81

ehelms: Touche. I don't know then. I will try to poke at it locally.

ehelms: I think this is the problem?

```
python38-pulp-cli.noarch                        0.12.0-2.el8                             @pulpcore 
```

https://github.com/pulp/pulp-cli/blob/e567714d063d2b1449b1d54869acaf11fa56b4e7/docs/CHANGES.md#0130-2021-12-16

This would indicate you need at least 0.13+ in order to specify a CA cert and thus use verify_ssl. I would also ask if we could configure the ca-cert in the TOML file? And if not, use PULP_CA_BUNDLE.

evgeni: `REQUESTS_CA_BUNDLE` *should* work with earlier releases, and `pulp status` fails without setting it with the same error we're seeing now with `user list`

evgeni: This is some kind of race condition. Once I reverse the order of `pulp status` and `pulp user list` the former starts to fail…

spec/acceptance/cli_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    its(:stdout) { is_expected.to match(/valid pulp-cli config/) }
+    its(:stderr) { is_expected.to eq '' }
+  end
+
+  describe command("REQUESTS_CA_BUNDLE=/etc/pulpcore-certs/ca-cert.pem pulp status") do
+    its(:exit_status) { is_expected.to eq 0 }
+    its(:stdout) { is_expected.to match(/versions/) }
+    its(:stderr) { is_expected.not_to match(/Error/) }
+    # currently this contains a warning:
+    # SubjectAltNameWarning: Certificate for centos8-64.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818.
```
ekohl, line 30: We fixed it for 8:
```suggestion
    # SubjectAltNameWarning: Certificate for centos7-64.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818.
```

spec/acceptance/cli_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+      its(:stderr) { is_expected.to eq '' }
+    end
+  end
+
+  describe command("REQUESTS_CA_BUNDLE=/etc/pulpcore-certs/ca-cert.pem pulp user list") do
+    its(:exit_status) { is_expected.to eq 0 }
+    its(:stdout) { is_expected.to match(/admin/) }
+    its(:stderr) { is_expected.not_to match(/Error/) }
+    # currently this contains a warning:
+    # SubjectAltNameWarning: Certificate for centos8-64.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818.
```
ekohl, line 41: And here
```suggestion
    # SubjectAltNameWarning: Certificate for centos7-64.example.com has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818.
```

spec/acceptance/cli_spec.rb
ekohl: Anything we can do about this?

evgeni: I have no idea where to start looking, to be honest.

We could limit testing to `pulp user list` (as that ensures proper certificate authentication) and try to fix the underlying issue *after* this PR?

ekohl: Some testing suggests it doesn't read the certificate at all.
```console
# export REQUESTS_CA_BUNDLE=/etc/pulpcore-certs/ca-cert.pem
# strace -o pulp pulp user list 
Error: HTTPSConnectionPool(host='centos8-64-1', port=443): Max retries exceeded with url: /pulp/api/v3/users/?offset=0&limit=25 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1125)')))
# grep pem pulp
# rm -f ~/.cache/squeezer/https___centos8-64-1_/api.json
# strace -o pulp pulp user list 
Error: {"detail":"Authentication credentials were not provided."}
# grep pem pulp
stat("/etc/pulpcore-certs/ca-cert.pem", {st_mode=S_IFREG|0640, st_size=1155, ...}) = 0
stat("/etc/pulpcore-certs/ca-cert.pem", {st_mode=S_IFREG|0640, st_size=1155, ...}) = 0
openat(AT_FDCWD, "/etc/pulpcore-certs/ca-cert.pem", O_RDONLY) = 4
```

So this does appear to be a bug in pulp-cli: it doesn't read the CA file if the cache is already present.

Also, username/password auth doesn't appear to work in our setup but I'm less worried about that.

ekohl: Manually testing the builds from https://github.com/theforeman/pulpcore-packaging/pull/434 show that 0.14 at least works. I suspect https://github.com/pulp/pulp-cli/commit/65de92aba7f33676490e58e498179ca2ad0c1e6a is the fix.

evgeni: Thanks for the debugging. Packaging is updated and this is now green!

manifests/cli.pp
ekohl: Given it's not exposed to users, I don't care that much. Let's start with this.

manifests/cli.pp
ekohl: It's not exposed to users so I think this is fine.

ekohl (approved):

evgeni:
@ehelms any further comments, or are you happy with it as it is too?

ehelms (approved):

PR #253, opened 2022-03-29
Generate the acceptance test CA with SAN

ekohl:
Modern tooling prefers the subjectAltName extension and the CN on a certificate is considered deprecated. Some tools even complain about this. However, the -addext command is unavailable on EL7 because openssl is too old.

### DIFF
--- a/spec/setup_acceptance_node.pp
+++ b/spec/setup_acceptance_node.pp
@@ -45,13 +45,20 @@ $client_csr = "${directory}/client-csr.pem"
 $client_cert = "${directory}/client-cert.pem"
 $client_key = "${directory}/client-key.pem"
 
+# EL7 lacks openssl -addext
+if $facts['os']['release']['major'] == '7' {
+  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
+} else {
+  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -addext 'subjectAltName = DNS:${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
+}
+
 exec { 'Create certificate directory':
   command => "mkdir -p ${directory}",
   path    => ['/bin', '/usr/bin'],
   creates => $directory,
 }
 -> exec { 'Generate certificate':
-  command   => "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365",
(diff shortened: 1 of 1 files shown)


spec/setup_acceptance_node.pp
```diff
@@ -45,13 +45,20 @@
 $client_cert = "${directory}/client-cert.pem"
 $client_key = "${directory}/client-key.pem"
 
+# EL7 lacks openssl -addext
+if $facts['os']['release']['major'] == '7' {
+  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365",
+} else {
+  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -addext 'subjectAltName = DNS:${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365",
+}
```
evgeni, line 53:
```suggestion
if $facts['os']['release']['major'] == '7' {
  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
} else {
  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -addext 'subjectAltName = DNS:${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
}
```

spec/setup_acceptance_node.pp
```diff
@@ -45,13 +45,20 @@
 $client_cert = "${directory}/client-cert.pem"
 $client_key = "${directory}/client-key.pem"
 
+# EL7 lacks openssl -addext
+if $facts['os']['release']['major'] == '7' {
+  $ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
```
ehelms, line 50: You could opt to deploy a file with extensions defined in it similar to https://github.com/theforeman/foreman-installer/blob/develop/spec/fixtures/katello-certs-check/extensions.txt

And then point at the extensions when running this command similar to https://github.com/theforeman/foreman-installer/blob/develop/spec/fixtures/katello-certs-check/create_cert.sh#L36

ekohl: I considered that, but how long do we plan on supporting EL7? It is mostly needed in the pulp-cli work from #252 but IMHO there's it's sufficient to only check stderr on EL8.

evgeni: #252 passes with this change picked on EL8, so I'm happy enough.

(On the other hand, we could consider limiting the number of ways we generate certs in our CI…)

evgeni (approved):

PR #256, opened 2022-05-26
Add Pulpcore 3.18 support

ianballou:
Looking at the last [Pulpcore version bump](https://github.com/theforeman/puppet-pulpcore/commit/6f09c2ff328fac9fcaf1388d9112a5130db3e2d7), it seems this is all we need to support Pulpcore 3.18.

3.18 is still built for EL7 so we shouldn't have to drop any of the EL7 related bits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
         setfile: ${{fromJson(needs.setup_matrix.outputs.beaker_setfiles)}}
         puppet: ${{fromJson(needs.setup_matrix.outputs.puppet_major_versions)}}
         pulpcore_version:
+          - '3.18'
           - '3.17'
           - '3.16'
     name: Acceptance / ${{ matrix.puppet.name }} - ${{ matrix.setfile.name }} - Pulp ${{ matrix.pulpcore_version }}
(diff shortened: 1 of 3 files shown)


ehelms:
Thanks @ianballou. You are performing two actions here: adding 3.18 and dropping 3.16. Are the two required to happen together? If 3.18 can be added without dropping 3.16, I would go that route so as to avoid conflating the two, and to avoid having to necessarily do a major version bump of the module right away.

As we state in the README: "Whenever a Pulpcore version is dropped, the module's major version is increased."

That isn't to say we should not drop 3.16 support, more so that we should do that separately from 3.18 support unless the two are required together. That will allow us to:

 * Bump the minor version of the module with 3.18 support and get a release out
 * Follow up by dropping 3.16 and releasing a module with the major version bumped

ianballou:
> Thanks @ianballou. You are performing two actions here: adding 3.18 and dropping 3.16. Are the two required to happen together? If 3.18 can be added without dropping 3.16, I would go that route so as to avoid conflating the two, and to avoid having to necessarily do a major version bump of the module right away.
> 
> As we state in the README: "Whenever a Pulpcore version is dropped, the module's major version is increased."
> 
> That isn't to say we should not drop 3.16 support, more so that we should do that separately from 3.18 support unless the two are required together. That will allow us to:
> 
>     * Bump the minor version of the module with 3.18 support and get a release out
> 
>     * Follow up by dropping 3.16 and releasing a module with the major version bumped

That all sounds fine to me, we can drop 3.16 later.

ianballou:
This was discussed on IRC, but it seems the test errors are due to pulp_cli being out of date.  https://github.com/theforeman/pulpcore-packaging/pull/513

ehelms (approved):

PR #257, opened 2022-06-09
Drop EL7 support

ehelms:


ekohl (commented):
You can also drop https://github.com/theforeman/puppet-pulpcore/blob/c1331be3387362d3dd8e1d3a28f103d9b38a5fc7/spec/acceptance/hieradata/os/RedHat/7.yaml

Technically this mostly saves CI resources. I wonder if it makes sense to keep it in this module until we stop supporting EL7 altogether, including releases for cherry picks. Or until the next major version bump of this module.

ekohl:
I wonder if the tests will pass after a rebase.

ehelms:
Tests are passing

PR #260, opened 2022-08-02
Fix Datatypes after udated Apache

cocker-cc:


ekohl:
I think this is the wrong solution. It really is a regression in puppetlabs-apache: https://github.com/puppetlabs/puppetlabs-apache/pull/2275 IMHO.

PR #263, opened 2022-09-07
Always ignore missing package indices for APT repos within Katello

quba42:
For Katello users this is the sane/expected default behaviour. It allows for the synchronization of partial Ubuntu mirrors, a very common use case in the wild. Users often stumble over the error that is thrown for such mirrors without this setting.


quba42:
Should I create a Katello issue to go with this change?

templates/settings.py.erb
```diff
@@ -72,3 +72,6 @@ LOGGING = {
         }
     }
 }
+
+# pulp_deb plugin default behaviour:
+FORCE_IGNORE_MISSING_PACKAGE_INDICES = True
```
ekohl, line 77: This is only relevant for pulp_deb, and as such it must be in a fragment.

In practice it means https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/deb.pp must ensure the content. See  https://github.com/theforeman/puppet-pulpcore/commit/5858c93e91a29f910aa9eab0e0fb6bd34a69e071 for a similar change.

If you add it, I'd also consider adding a class parameter to change the value.

ekohl (changes requested):

ekohl:
> Should I create a Katello issue to go with this change?

It would be an installer issue. I always appreciate those since it helps with cherry picks.

templates/settings.py.erb
quba42: Thanks for the pointers!

manifests/plugin/deb.pp
```diff
@@ -27,7 +32,14 @@
     $content = undef
   }
 
+  if $force_ignore_missing_package_indices {
+    $deb_plugin_config_fimpi = 'True'
+  } else {
+    $deb_plugin_config_fimpi = 'False'
+  }
```
ekohl, line 39: You can also write it as:
```suggestion
  $deb_plugin_config_fimpi = if $force_ignore_missing_package_indices { 'True' } else { 'False' }
```

Or, if you raise the lower bound of the stdlib dependency to 8.3.0 (in metadata.json):
```suggestion
  $deb_plugin_config_fimpi = to_python($force_ignore_missing_package_indices)
```

ekohl (commented):
Thanks. This is indeed the correct place.

You can also add a test. Either check for the concat fragment, like this:
https://github.com/theforeman/puppet-pulpcore/blob/5858c93e91a29f910aa9eab0e0fb6bd34a69e071/spec/classes/plugin_ansible_spec.rb#L26-L28

Or add `.with_config('FORCE... = True')` to `contain_pulpcore__plugin('deb')` here:
https://github.com/theforeman/puppet-pulpcore/blob/5858c93e91a29f910aa9eab0e0fb6bd34a69e071/spec/classes/plugin_deb_spec.rb#L19-L21

Looking at it now I'd go with the latter since it's easier and you're testing less for internal details.

ekohl:
@quba42 we're about to release the module so we're ready for RC1. Any plan to finish this? Ideally speaking we'd release this module this week.

quba42:
Ugh totally forgot about this. I will have a look if I can quickly finish it now...

quba42:
@ekohl Added your suggestions and a test (that I think makes sense)...

ekohl (approved):
Thanks! Now I can do the release for Foreman 3.5 :)

PR #264, opened 2022-09-20
Add pulpcore deprecation and django_guid loggers

ehelms:


manifests/init.pp
```diff
@@ -224,6 +230,8 @@
   Boolean $cache_enabled = false,
   Optional[Variant[Integer[1], Enum['None']]] $cache_expires_ttl = undef,
   Enum['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG'] $log_level = 'INFO',
+  Enum['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG'] $log_level_deprecation = 'ERROR',
+  Enum['CRITICAL', 'ERROR', 'WARNING', 'INFO', 'DEBUG'] $log_level_correlation_id = 'WARNING',
```
ekohl, line 234: In foreman we expose `loggers` as a hash. Rather than adding parameters for every logger, should we also expose a hash so we can easily add additional loggers? I can imagine this being useful for testing too if we can set it via Hiera.

ekohl: I've been thinking about merging hashes, but wondered if it worked. So I wrote https://github.com/theforeman/puppet-pulpcore/pull/265. It's more complex than it needs to be (adds a new concat fragment to ease testing, adds type aliases), but they made it easier to verify.

This should address the problem that users need to know the built in values, but also gives them freedom. Let me know what you think.

PR #266, opened 2022-10-05
Add pulpcore 3.21 support

sjha4:
Adding pulpcore 3.21 since we have that packaged. Working on other bits to enable this in katello and gems packaging.

### DIFF
--- a/spec/support/acceptance/examples.rb
+++ b/spec/support/acceptance/examples.rb
@@ -56,7 +56,11 @@ shared_examples 'the default pulpcore application' do
 
   describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/", cacert: "#{certdir}/ca-cert.pem") do
     # Requires authentication: https://github.com/pulp/pulpcore/issues/2340
-    its(:response_code) { is_expected.to eq(403) }
+    if(!['3.16', '3.17', '3.18', '3.19'].include?(ENV['BEAKER_FACTER_PULPCORE_VERSION']))
+      its(:response_code) { is_expected.to eq(200) }
+    else
+      its(:response_code) { is_expected.to eq(403) }
+    end
     its(:exit_status) { is_expected.to eq 0 }
   end
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,6 +66,7 @@ jobs:
         setfile: ${{fromJson(needs.setup_matrix.outputs.beaker_setfiles)}}
         puppet: ${{fromJson(needs.setup_matrix.outputs.puppet_major_versions)}}
         pulpcore_version:
+          - '3.21'
           - '3.18'
           - '3.17'
           - '3.16'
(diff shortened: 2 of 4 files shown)


ekohl (changes requested):
Looks good, but if you look at the failures (most visible in annotations @ https://github.com/theforeman/puppet-pulpcore/pull/266/files) you'll see that this issue has been resolved and now properly returns HTTP 200 again:
https://github.com/theforeman/puppet-pulpcore/blob/87aa4c412ee879d1a15927f24a0ecf2d6a429c7c/spec/support/acceptance/examples.rb#L58-L59

I'd say that the easiest way to check it is `ENV['BEAKER_FACTER_PULPCORE_VERSION']`.

spec/support/acceptance/examples.rb
```diff
@@ -56,7 +56,11 @@
 
   describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/", cacert: "#{certdir}/ca-cert.pem") do
     # Requires authentication: https://github.com/pulp/pulpcore/issues/2340
-    its(:response_code) { is_expected.to eq(403) }
+    if(ENV['BEAKER_FACTER_PULPCORE_VERSION'] == '3.21')
```
ekohl, line 59: I'd prefer to be forward compatible and check if it isn't 3.17/3.18. Also because when you run it outside of CI (where we don't set the variable) it also ends up as 3.21 because that's the default in `pulpcore::repo`.

ekohl (changes requested):

wbclark:
Tangential comment -- I took a look at Pulp changelogs to see what shiny new features we get after your work here is done.

In particular, some users will definitely want to disable https://github.com/pulp/pulpcore/issues/3115 . I took at a stab at that with https://github.com/theforeman/puppet-pulpcore/pull/267

https://github.com/pulp/pulpcore/issues/2329 was added previously in 3.19 and also looks like something that users may want to configure somewhat persistently for certain medium or longer term troubleshooting or monitoring scenarios.

spec/support/acceptance/examples.rb
```diff
@@ -56,7 +56,11 @@
 
   describe curl_command("https://#{host_inventory['fqdn']}/pulp/api/v3/", cacert: "#{certdir}/ca-cert.pem") do
     # Requires authentication: https://github.com/pulp/pulpcore/issues/2340
-    its(:response_code) { is_expected.to eq(403) }
+    if(ENV['BEAKER_FACTER_PULPCORE_VERSION'] !~ /3.1+/)
```
ekohl, line 59: The `+` is kind of redundant since it means match `3.1` but also `3.1111`. It's a partial match so it ends up working. A stricter match is:
```suggestion
    if(ENV['BEAKER_FACTER_PULPCORE_VERSION'] !~ /^3.1[6-9]$/)
```

ekohl (approved):
Minor suggestion.

spec/support/acceptance/examples.rb
evgeni: Or you use the good old `['3.16', '3.17', '3.18', '3.19'].include?(…)` ;)

(I don't mind either way, just don't like regular expressions too much 🙃)

sjha4: > The `+` is kind of redundant since it means match `3.1` but also `3.1111`. It's a partial match so it ends up working. A stricter match is:

I love how this is a bug someone would have to find when pulpcore 3.100 is out.😄 
I think I like ['3.16', '3.17', '3.18', '3.19'].include?(…). Keeps it simple. Will update.

evgeni: The regex also would match `3116` ;)

evgeni (approved):

evgeni:
Thanks @sjha4!

PR #268, opened 2022-11-01
Refs #35607 - convert booleans from lowercase ruby to uppercase pytho…

wbclark:
…n in settings.py

### DIFF
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -14,7 +14,7 @@
 <% if scope['pulpcore::telemetry'].nil? -%>
 # TELEMETRY = False
 <% else -%>
-TELEMETRY = <%= scope['pulpcore::telemetry'] %>
+TELEMETRY = <%= scope.call_function('to_python', [scope['pulpcore::telemetry']]) %>
 <% end -%>
 
 CONTENT_HOST = "<%= scope['pulpcore::servername'] %>"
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
I'd also rephrase the message to `convert telemetry value to Python` or something like that.

templates/settings.py.erb
```diff
@@ -14,7 +14,7 @@
 <% if scope['pulpcore::telemetry'].nil? -%>
 # TELEMETRY = False
 <% else -%>
-TELEMETRY = <%= scope['pulpcore::telemetry'] %>
+TELEMETRY = <%= scope['pulpcore::telemetry'].to_s.capitalize %>
```
evgeni, line 17: `puppetlabs-stdlib` also has [`to_python`](https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/to_python.rb) since 8.0.0, but not sure how we feel about bumping that dep?

@ekohl?

ekohl: I made the same comment. We already raised it in 5e2bfd7e335c63998f4abb8d75647cd9837cc2d0 when we started to use `to_python` in the Debian plugin.

ekohl (approved):

wbclark:
Thanks @ekohl @evgeni for the suggestions.

I tried it as `Stdlib::to_python` first but I believe it needs to be `stdlib::to_python`, giving that another try

PR #269, opened 2022-11-01
Expose Apache server aliases as a parameter

ekohl:
This allows setting a CNAME.

Passing the aliases to Apache is especially important when there are multiple vhosts or when using mod_md. While today we don't do either, having the option is good. It can also help with debugging since sosreport does collect the vhosts, but not the certificates. If ServerAliases matches what's on the certificate, that can give a hint.

Currently still working on tests.

### DIFF
--- a/spec/classes/pulpcore_spec.rb
+++ b/spec/classes/pulpcore_spec.rb
@@ -79,7 +79,6 @@ describe 'pulpcore' do
         it 'configures apache' do
           is_expected.to contain_class('pulpcore::apache')
           is_expected.to contain_apache__vhost('pulpcore')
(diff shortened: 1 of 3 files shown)


spec/classes/pulpcore_spec.rb
```diff
@@ -79,7 +79,7 @@
         it 'configures apache' do
           is_expected.to contain_class('pulpcore::apache')
           is_expected.to contain_apache__vhost('pulpcore')
-          is_expected.not_to contain_apache__vhost__fragment('pulpcore-http-pulpcore')
```
ekohl, line 82: The bugs you find reading code. Turns out I was the one to blame.

PR #274, opened 2023-01-31
Add param for unsafe advisory conflict resolution

m-bucher:
https://community.theforeman.org/t/katello-incoming-and-existing-advisories-have-the-same-id-but-different-timestamps/23954/8

### DIFF
--- a/manifests/plugin/rpm.pp
+++ b/manifests/plugin/rpm.pp
@@ -6,9 +6,14 @@
 #   Pulpcore's KEEP_CHANGELOG_LIMIT setting. Uses Pulpcore's default when
 #   undefined. Increasing this limit will cause pulpcore workers to use more
 #   memory when more changelogs are available in the repo metadata.
+#
+# @param allow_automatic_unsafe_advisory_conflict_resolution
+#   Allow resolving of conflicts due to duplicate advisory ids with different creation dates
+#   https://docs.pulpproject.org/pulp_rpm/settings.html#allow-automatic-unsafe-advisory-conflict-resolution
 class pulpcore::plugin::rpm (
   Boolean $use_pulp2_content_route = false,
   Optional[Integer[0]] $keep_changelog_limit = undef,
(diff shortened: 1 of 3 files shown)


templates/settings.py.erb
```diff
@@ -60,6 +60,7 @@ REST_FRAMEWORK__DEFAULT_AUTHENTICATION_CLASSES = (
 ALLOWED_IMPORT_PATHS = <%= scope['pulpcore::allowed_import_path'] %>
 ALLOWED_EXPORT_PATHS = <%= scope['pulpcore::allowed_export_path'] %>
 ALLOWED_CONTENT_CHECKSUMS = <%= scope['pulpcore::allowed_content_checksums'] %>
+ALLOW_AUTOMATIC_UNSAFE_ADVISORY_CONFLICT_RESOLUTION = <%= scope['pulpcore::allow_automatic_unsafe_advisory_conflict_resolution'] %>
```
ekohl, line 63: You should convert this to Python booleans. See a few lines below for `CACHE_ENABLED`, or use `to_python()`.

ekohl (changes requested):

ekohl (changes requested):
I should have checked this earlier, but this comes from `pulp_rpm` so it should also live in [pulpcore::plugin::rpm](https://github.com/theforeman/puppet-pulpcore/blob/master/manifests/plugin/rpm.pp). See 5e2bfd7e335c63998f4abb8d75647cd9837cc2d0 for an example of how a plugin-specific config option was added earlier.

In this case there is already some inline content so it probably makes sense to move it to a template. While most templates are ERB, I'd like to move to EPP. In 2e2d3ecf486ed2b333e405d883b1bb2c59978911 I added an EPP template. Note how it uses `epp()` (with a context) instead of `template()` and a slightly different context.

ekohl (approved):
From a quick glance this looks ok and the test coverage should be sufficient.

m-bucher:
Thanks @ekohl for merging this.
I just remembered that I actually intended to set the default to `true`. What do you think?

PR #277, opened 2023-03-08
Drop Pulpcore 3.16 - 3.18, add EL 9 support

ekohl:
While it doesn't modify any code, does reduce the testing matrix. It's also chosen because 3.21 and 3.22 have EL9 and that keeps the tesing matrix easier.

Currently untested, going to rely on CI.

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -20,13 +20,16 @@ class pulpcore::repo (
     notify  => Anchor['pulpcore::repo'],
   }
 
-  package { 'pulpcore-dnf-module':
-    ensure      => $dist_tag,
-    name        => 'pulpcore',
-    enable_only => true,
-    provider    => 'dnfmodule',
-    require     => File['/etc/yum.repos.d/pulpcore.repo'],
-    notify      => Anchor['pulpcore::repo'],
+  # Only EL8 has DNF modules
+  if $dist_tag == 'el8' {
+    package { 'pulpcore-dnf-module':
+      ensure      => $dist_tag,
+      name        => 'pulpcore',
+      enable_only => true,
+      provider    => 'dnfmodule',
+      require     => File['/etc/yum.repos.d/pulpcore.repo'],
+      notify      => Anchor['pulpcore::repo'],
+    }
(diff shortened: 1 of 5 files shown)


evgeni:
```
  /bin/sh: line 1: hostname: command not found
```

seems we need a hack here (again?)

manifests/repo.pp
```diff
@@ -20,13 +20,16 @@
...
 
-  package { 'pulpcore-dnf-module':
-    ensure      => $dist_tag,
-    name        => 'pulpcore',
-    enable_only => true,
-    provider    => 'dnfmodule',
-    require     => File['/etc/yum.repos.d/pulpcore.repo'],
-    notify      => Anchor['pulpcore::repo'],
+  # Only EL8 has DNF modules
+  if $dist_tag == 'el8' {
```
evgeni, line 24: the tests still expect this on el9…

ekohl: Yes, I thought about fixing the other issues while I was at it but that it was filled with rabbits just to get an EL9 box within Vagrant.

spec/classes/repo_spec.rb
```diff
@@ -13,14 +13,18 @@
...
 
-      it 'configures the pulpcore module' do
-        is_expected.to contain_package('pulpcore-dnf-module')
-          .with_name('pulpcore')
-          .with_ensure(/^el\d+/)
-          .with_enable_only(true)
-          .with_provider('dnfmodule')
-          .that_requires('File[/etc/yum.repos.d/pulpcore.repo]')
-          .that_notifies('Anchor[pulpcore::repo]')
+      if os_facts[:op]['release']['major'] == '8'
```
evgeni, line 16:
```suggestion
      if os_facts[:os]['release']['major'] == '8'
```

;)

ekohl: I'll just leave this here until the other parts have been sorted.

ekohl:
This is now green, except on Puppet 6.

ehelms (approved):

PR #288, opened 2023-05-25
Fixes #36437 - preload pulpcore API and content code

evgeni:
this reduces the overall memory usage by keeping only one copy of the code, instead of N (= number of workers).

unscientific benchmark says that with 5 API and 17 content workers, this reduces the memory footprint from 500MiB to 125MiB and from 1.5GiB to 500MiB respectively.

those numbers are obviously before any actual work is executed and memory consumed by the app itself.


ekohl (approved):
Having a parameter we can toggle via Hiera does make it easier to revert, but I'm fine either way. No parameter also removes the need for tests and that does simplify things.

PR #289, opened 2023-05-25
Fixes #36438 - configure API request limit to avoid memory leaks

evgeni:


spec/classes/pulpcore_spec.rb
```diff
@@ -162,6 +164,21 @@
...
+          {
+            api_service_worker_max_requests: 100,
+            api_service_worker_max_requests_jitter: 10
+          }
+        end
+
+        it do
+          is_expected.to compile.with_all_deps
+          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests 100})
+          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests-jitter 10})
```
ekohl, line 178: You can chain them:
```suggestion
          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests 100}).with_content(%r{--max-requests-jitter 10})
```

spec/classes/pulpcore_spec.rb
```diff
@@ -150,6 +150,8 @@
           is_expected.to contain_pulpcore__socket_service('pulpcore-api')
           is_expected.to contain_systemd__unit_file('pulpcore-api.socket')
           is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--workers 2})
+          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests 50})
+          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests-jitter 30})
```
ekohl, line 154: You can chain them:
```suggestion
          is_expected.to contain_systemd__unit_file('pulpcore-api.service').with_content(%r{--max-requests 50}).with_content(%r{--max-requests-jitter 30})
```

ekohl (approved):

manifests/init.pp
```diff
@@ -233,6 +239,8 @@
   Integer[0] $api_service_worker_count = min(4, $facts['processors']['count']) + 1,
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
+  Integer[0] $api_service_worker_max_requests = 50,
```
dralley, line 242: @evgeni We discovered the "real" root cause now, so we can relax this number significantly.  I would back it off to a much higher value, like say 1000 for the limit and 200 for the jitter.

ehelms: Do you have a link to info about the real root cause?

dralley: See the discussion near the end on the BZ https://bugzilla.redhat.com/show_bug.cgi?id=2122872

PR #292, opened 2023-06-20
allow configuring HIDE_GUARDED_DISTRIBUTIONS setting

evgeni:


ekohl (commented):
According to the README we still support older versions. Do we care about those? They won't break, but I'm tempted to add a note to README that `hide_guarded_distributions` won't do anything.

evgeni:
We could add that to the help string or the readme, yes.

On the other hand, I doubt anyone would actually go try enabling this without reading pulpcore docs, that don't mention it on pre-3.22?

ekohl (approved):

PR #295, opened 2023-07-20
Relax max_requests

dralley:
The root cause of the memory leak has been established and resolved, so these aggressive values are no longer necessary


manifests/init.pp
```diff
@@ -243,8 +243,8 @@
   Integer[0] $api_service_worker_count = min(4, $facts['processors']['count']) + 1,
   Integer[0] $content_service_worker_timeout = 90,
   Integer[0] $api_service_worker_timeout = 90,
-  Integer[0] $api_service_worker_max_requests = 50,
-  Integer[0] $api_service_worker_max_requests_jitter = 30,
+  Integer[0] $api_service_worker_max_requests = 800,
+  Integer[0] $api_service_worker_max_requests_jitter = 100,
```
dralley, line 247: These seem reasonable to me, but they're just spitballs.  Probably 4,000 and 500 would be fine too.

ekohl:
> The root cause of the memory leak has been established and resolved, so these aggressive values are no longer necessary

Out of interest: can you share a link?

dralley:
https://github.com/pulp/pulpcore/pull/4090

dralley:
I believe the tests are testing changing those values rather than the current values, so they don't need to be changed.  Unless you mean something else.

dralley:
@ekohl Pushed

ekohl:
https://ci.theforeman.org/job/pulpcore-3.22-rpm-pipeline/ is still blocked (there was some issue in CentOS CI) and I'd like to wait for that to pass.

ekohl (approved):

PR #302, opened 2023-08-21
allow configuring IMPORT_WORKERS_PERCENT

evgeni:


ekohl (commented):
Looks like it's not documented @ https://docs.pulpproject.org/pulpcore/configuration/settings.html#pulp-settings so I wonder which version introduced this.

ekohl:
Looks like it was nominated for 3.21 and newer, so that's sufficient.

PR #304, opened 2023-08-30
Refs #36709 - Expose ANSIBLE_PERMISSION_CLASSES setting

ekohl:


ekohl:
The Redmine issue asks for it to be set to `[]` but I feel that configuration belongs in puppet-foreman_proxy_content since it's a non-default option.

evgeni (approved):

PR #315, opened 2023-11-08
Fixes #36902 - Support Pulpcore 3.39, drop older versions

ianballou:
Assumes that https://github.com/pulp/pulpcore/issues/4679 is in Pulpcore 3.39.

Also assumes that https://github.com/pulp/pulpcore/issues/4681 is going to be fixed for Pulpcore 3.39 in time. This doesn't affect the installer but we shouldn't release Katello with that Pulpcore bug.

I had to assume that the entry point will live in `/usr/bin` since, when installed via `pip`, it is installed to `/usr/local/bin`.

I verified these work (minus the jitter for now):

```
ExecStart=/usr/local/bin/pulpcore-api --preload --timeout 90 --workers 5 --max-requests 800 --access-logfile - --access-logformat 'pulp [%({correlation-id}o)s]: %(h)s %(l)s %(u)s %(t)s "%(r)s" %(s)s %(b)s "%(f)s" "%(a)s"'
```

and

```
ExecStart=/usr/local/bin/pulpcore-content  --preload --timeout 90  --workers 5 --access-logfile -
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,21 +10,9 @@ All supported versions are listed below. For every supported version, acceptance
 
 Supported operating systems are listed in `metadata.json` but individual releases can divert from that. For example, if Pulpcore x.y drops EL7, it will still be listed in metadata.json until all versions supported by the module have dropped it. Similarly, if x.z adds support for EL9, it'll be listed in `metadata.json` and all versions that don't support EL9 will have a note.
 
-### Pulpcore 3.28
+### Pulpcore 3.39
 
 Default recommended version.
-
-### Pulpcore 3.22
-
-Supported version.
-
-### Pulpcore 3.21
-
-Supported version. The parameter `$hide_guarded_distributions` doesn't work since it's a Pulp 3.22 feature.
-
-### Pulpcore 3.16 - 3.18
-
-Untested, but should work besides the `$hide_guarded_distributions` parameter. Not recommended.
 
 ## Installation layout
 
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -8,7 +8,7 @@
 #   An optional value for gpgkey to be used for yumrepo, instead of the default.
 #   If an empty string is passed, gpgcheck will be disabled.
 class pulpcore::repo (
-  Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $version = '3.28',
+  Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $version = '3.39',
   Optional[Stdlib::HTTPUrl] $baseurl = undef,
   Optional[String[0]] $gpgkey = undef,
 ) {
@@ -18,21 +18,9 @@ class pulpcore::repo (
     descr    => "Pulpcore ${version}",
     baseurl  => pick($baseurl, "https://yum.theforeman.org/pulpcore/${version}/${dist_tag}/\$basearch"),
     enabled  => '1',
-    gpgcheck => if $gpgkey == '' { '0' } else { '1' },
+    gpgcheck => if $gpgkey == '' or $version == 'nightly' { '0' } else { '1' },
     gpgkey   => pick($gpgkey, "https://yum.theforeman.org/pulpcore/${version}/GPG-RPM-KEY-pulpcore"),
     notify   => Anchor['pulpcore::repo'],
-  }
-
-  # Only EL8 has DNF modules
-  if $dist_tag == 'el8' {
-    package { 'pulpcore-dnf-module':
-      ensure      => $dist_tag,
-      name        => 'pulpcore',
-      enable_only => true,
-      provider    => 'dnfmodule',
-      require     => Yumrepo['pulpcore'],
-      notify      => Anchor['pulpcore::repo'],
-    }
   }
 
   # An anchor is used because it can be collected
--- a/spec/classes/repo_spec.rb
+++ b/spec/classes/repo_spec.rb
@@ -14,20 +14,6 @@ describe 'pulpcore::repo' do
           .with_gpgkey(%r{https://yum.theforeman.org/pulpcore/\d+\.\d+/GPG-RPM-KEY-pulpcore})
           .that_notifies('Anchor[pulpcore::repo]')
         }
-
-        if os_facts[:os]['release']['major'] == '8'
-          it 'configures the pulpcore module' do
-            is_expected.to contain_package('pulpcore-dnf-module')
-              .with_name('pulpcore')
-              .with_ensure(/^el\d+/)
-              .with_enable_only(true)
-              .with_provider('dnfmodule')
-              .that_requires('Yumrepo[pulpcore]')
-              .that_notifies('Anchor[pulpcore::repo]')
-          end
-        else
-          it { is_expected.not_to contain_package('pulpcore-dnf-module') }
-        end
       end
 
       describe "with nightly version" do
--- a/templates/pulpcore-api.service.erb
+++ b/templates/pulpcore-api.service.erb
@@ -11,7 +11,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=<%= scope['pulpcore::user_home'] %>
 RuntimeDirectory=pulpcore-api
-ExecStart=/usr/libexec/pulpcore/gunicorn pulpcore.app.wsgi:application \
+ExecStart=/usr/bin/pulpcore-api \
           --preload \
           --timeout <%= scope['pulpcore::api_service_worker_timeout'] %> \
           --workers <%= scope['pulpcore::api_service_worker_count'] %> \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,6 @@ jobs:
     uses: voxpupuli/gha-puppet/.github/workflows/beaker.yml@v2
     with:
       pidfile_workaround: 'false'
-      beaker_facter: 'pulpcore_version:Pulp:3.28,3.22,3.21'
+      beaker_facter: 'pulpcore_version:Pulp:nightly,3.39'
       rubocop: false
       cache-version: '1'
--- a/.sync.yml
+++ b/.sync.yml
@@ -1,3 +1,3 @@
 ---
 .github/workflows/ci.yml:
-  beaker_facter: 'pulpcore_version:Pulp:3.28,3.22,3.21'
+  beaker_facter: 'pulpcore_version:Pulp:nightly,3.39'
--- a/templates/pulpcore-content.service.erb
+++ b/templates/pulpcore-content.service.erb
@@ -11,10 +11,9 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=<%= scope['pulpcore::user_home'] %>
 RuntimeDirectory=pulpcore-content
-ExecStart=/usr/libexec/pulpcore/gunicorn pulpcore.content:server \
+ExecStart=/usr/bin/pulpcore-content \
           --preload \
           --timeout <%= scope['pulpcore::content_service_worker_timeout'] %> \
-          --worker-class 'aiohttp.GunicornWebWorker' \
           --workers <%= scope['pulpcore::content_service_worker_count'] %> \
           --access-logfile -
 ExecReload=/bin/kill -s HUP $MAINPID


ianballou:
I haven't touched the tests yet.

templates/pulpcore-api.service.erb
```diff
@@ -11,7 +11,7 @@ User=<%= scope['pulpcore::user'] %>
 Group=<%= scope['pulpcore::group'] %>
 WorkingDirectory=<%= scope['pulpcore::user_home'] %>
 RuntimeDirectory=pulpcore-api
-ExecStart=/usr/libexec/pulpcore/gunicorn pulpcore.app.wsgi:application \
+ExecStart=/usr/bin/pulpcore-api \
```
evgeni, line 14: This will require adjustments in https://github.com/theforeman/foreman/blob/1eb33c7eb13e6f547f8bcb21ccf5969e8b876e98/extras/pcp/foreman-hotproc.conf#L9-L10 to make PCP pick up the new processes.

(Just add the new names to the list, so the old ones keep working, I guess)

ianballou: https://github.com/theforeman/foreman/pull/9898

ianballou:
I suppose this is going to break all of the older Pulp tests?

ekohl (changes requested):
I agree this will break older releases, which is fine. Last time we did that was https://github.com/theforeman/puppet-pulpcore/commit/5db8f484474bf9d981ada639cf343cd171646bbb but it changed a bit since then. To update the supported versions:

https://github.com/theforeman/puppet-pulpcore/blob/101c3256339dd6363b9aef2772f65b345c30769a/.sync.yml#L3
https://github.com/theforeman/puppet-pulpcore/blob/101c3256339dd6363b9aef2772f65b345c30769a/.github/workflows/ci.yml#L18
https://github.com/theforeman/puppet-pulpcore/blob/101c3256339dd6363b9aef2772f65b345c30769a/manifests/repo.pp#L6
https://github.com/theforeman/puppet-pulpcore/blob/master/README.md#support-policy

Of course it will fail until we have repositories. We can also add nightly since that now exists.

ianballou:
@ekohl I updated all of those files including the README.

ianballou:
[test]

ianballou:
Actually I don't think we can test just yet, the packages are available but as nightly https://yum.theforeman.org/pulpcore/nightly/el8/

evgeni:
I've just merged https://github.com/theforeman/puppet-pulpcore/pull/318, if you rebase and add "nightly" to the matrix, at least that we should see working

evgeni:
```
  missing groups or modules: pulpcore:el8
```

this seems odd… cc @ehelms @odilhao

evgeni:
and on el9 it tries to fetch gpg keys, fails (obviously) and bails out

```
  Error: /Stage[main]/Pulpcore::Install/Package[pulpcore]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install pulpcore' returned 1: Status code: 404 for https://yum.theforeman.org/pulpcore/nightly/GPG-RPM-KEY-pulpcore (IP: 151.101.42.49)
```

ianballou:
I just found out that Python 11 doesn't work with module streams, so there will be no modules. This might cause some bigger issues, at the very least our docs will need updating...

evgeni:
Oooh, you're saying Pulpcore is also demodularized now?!

ianballou:
> Oooh, you're saying Pulpcore is also demodularized now?!

Seems like it, I just found out a couple minutes ago. Sounds like it's not by choice, but by how Python 3.11 works. Sounds like we could still ship the modular metadata if removing it is too troublesome.

evgeni:
Well, just because Python is not shipped in a module anymore doesn't mean we can't ship (no Android, not shit, I mean sure, in a way, but no) Pulpcore as a module. That being said, there is no reason for us to now. I think this is fine and we can just drop the module stuff from repo.pp.

ekohl:
I didn't review the last version, but I opened the question whether pulpcore needs modularity given Python 3.11 is just a regular package. That would needs updates, which makes the whole Pulp upgrade from a platform perspective an even larger risk than it already was, but cleaner

ianballou:
Would it only be the docs and this puppet-pulpcore repo that will need updates to remove the modular expectation?

evgeni:
That would be my naive assumption.

Maybe foreman operations collection, don't recall how we set up Pulpcore there.

ianballou:
I've pushed a change to drop the modules, let's see if I did it right.

evgeni:
https://github.com/search?q=org%3Atheforeman%20%22pulpcore%3Ael8%22&type=code
Agrees

https://github.com/search?q=org%3Atheforeman%20%22pulpcore%3A%22&type=code adds foreman maintain

So not a super crazy list.

The most interesting question will be: what does dnf think when a package (pulpcore-selinux, e.g.) came from a module (in  3.28) and now does not. I recall dnf having some very odd way of memorizing module metadata for repositories that are gone. Things to try out.

ianballou:
Also the nightly gpg keys live somewhere new: https://download.copr.fedorainfracloud.org/results/@theforeman/pulpcore-nightly-staging/pubkey.gpg

I guess we need a different repo spec for nightly since the GPG key is elsewhere?

Odilhao:
The GPG key and the repo name for 3.39 will be same as usual

ianballou:
Tests are now passing for nightly :)

evgeni:
It's something.jpg

Seems pulp cli is broken on el9. But it's a start.

README.md
```diff
@@ -10,21 +10,13 @@ All supported versions are listed below. For every supported version, acceptance
 
 Supported operating systems are listed in `metadata.json` but individual releases can divert from that. For example, if Pulpcore x.y drops EL7, it will still be listed in metadata.json until all versions supported by the module have dropped it. Similarly, if x.z adds support for EL9, it'll be listed in `metadata.json` and all versions that don't support EL9 will have a note.
 
-### Pulpcore 3.28
+### Pulpcore nightly
```
evgeni, line 13: I wonder if we should call out nightly at all in the README, it's not something we're going to *support* officially (as it's a moving target).

evgeni:
3.39 is obviously failing right now, but other than that this looks fine to me!

Odilhao:
> 3.39 is obviously failing right now, but other than that this looks fine to me!

It should be able to get released after theforeman/forklift#1731

evgeni:
> > 3.39 is obviously failing right now, but other than that this looks fine to me!
> 
> It should be able to get released after [theforeman/forklift#1731](https://github.com/theforeman/forklift/pull/1731)

I merged that, but centos ci is broken right now, so you won't be able to do a release pipe :(

manifests/repo.pp
```diff
@@ -23,18 +23,6 @@
     notify   => Anchor['pulpcore::repo'],
   }
 
-  # Only EL8 has DNF modules
-  if $dist_tag == 'el8' {
-    package { 'pulpcore-dnf-module':
-      ensure      => $dist_tag,
```
ekohl, line 29: Upgrade wise: I wonder what happens. I suspect the module will remain enabled, even if it doesn't exist anymore. Though we pretty much only use this class for CI, it will be relevant for when we consider real installations. Perhaps something the installer should do via a hook, perhaps something for foreman_maintain.

spec/classes/repo_spec.rb
```diff
@@ -14,20 +14,7 @@
...
-              .with_ensure(/^el\d+/)
-              .with_enable_only(true)
-              .with_provider('dnfmodule')
-              .that_requires('Yumrepo[pulpcore]')
-              .that_notifies('Anchor[pulpcore::repo]')
-          end
-        else
-          it { is_expected.not_to contain_package('pulpcore-dnf-module') }
-        end
+        it { is_expected.not_to contain_package('pulpcore-dnf-module') }
```
ekohl, line 17: These expect to/not to checks are useful for if-conditions but since we don't have this in the source anymore, I'd simply drop this.
```suggestion
```

ekohl (changes requested):

manifests/repo.pp
ianballou: It's looking like  the pulpcore module will need to be disabled -- without it the new pulpcore packages aren't being picked up for update by dnf.

ianballou: Actually I can't confirm this yet -- seems even without the pulpcore module `dnf` isn't picking up the new pulpcore packages. Potentially a packaging issue.

ianballou: After a chat, we decided that we'll be adding the module removal step to our documentation. It will indeed need to be disabled. Furthermore, adding a removal step to the foreman-installer is too late because users already `dnf update` before running the installer.

ianballou:
Nightly GPG key is an issue again: `Error: /Stage[main]/Pulpcore::Install/Package[pulpcore]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install pulpcore' returned 1: Status code: 404 for https://yum.theforeman.org/pulpcore/nightly/GPG-RPM-KEY-pulpcore (IP: 146.75.94.49)`

ianballou:
I'm attempting to remove the GPG key requirement for nightly...

ianballou:
Hmm still erroring `  Error: /Stage[main]/Pulpcore::Install/Package[pulpcore]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/dnf -d 0 -e 1 -y install pulpcore' returned 1: Status code: 404 for https://yum.theforeman.org/pulpcore/nightly/GPG-RPM-KEY-pulpcore`

manifests/repo.pp
```diff
@@ -18,23 +18,15 @@
     descr    => "Pulpcore ${version}",
     baseurl  => pick($baseurl, "https://yum.theforeman.org/pulpcore/${version}/${dist_tag}/\$basearch"),
     enabled  => '1',
-    gpgcheck => if $gpgkey == '' { '0' } else { '1' },
-    gpgkey   => pick($gpgkey, "https://yum.theforeman.org/pulpcore/${version}/GPG-RPM-KEY-pulpcore"),
+    gpgcheck => if $gpgkey == '' or version == 'nightly' { '0' } else { '1' },
+    gpgkey   => if version == 'nightly' {
```
ianballou, line 22: @ekohl what do you think about this method of setting the gpgkey for nightly? I've been doing some Puppet trial and error.

evgeni: It needs to be `$version` to pick up the variable ;)

ianballou: ah thank you!

evgeni: I am amazed it didn't yell at you that it has no friggin idea what `version` is.

ianballou:
Test-wise I think this is good to go -- the only failing bits are Pulp-CLI on EL9 with the 3.39 repo. Nightly is working on EL9, and everything EL8 works.

manifests/repo.pp
```diff
@@ -18,23 +18,11 @@
     descr    => "Pulpcore ${version}",
     baseurl  => pick($baseurl, "https://yum.theforeman.org/pulpcore/${version}/${dist_tag}/\$basearch"),
     enabled  => '1',
-    gpgcheck => if $gpgkey == '' { '0' } else { '1' },
-    gpgkey   => pick($gpgkey, "https://yum.theforeman.org/pulpcore/${version}/GPG-RPM-KEY-pulpcore"),
+    gpgcheck => if $gpgkey == '' or $version == 'nightly' { '0' } else { '1' },
+    gpgkey   => if $version == 'nightly' { undef } else { pick($gpgkey, "https://yum.theforeman.org/pulpcore/${version}/GPG-RPM-KEY-pulpcore") },
```
evgeni, line 22: I think you don't need to `undef` the key here, setting `gpgcheck` to `0` in the line above should be sufficient.

evgeni:
So I tried an upgrade by installing 3.28 with the module currently on Forge and then upgrading to 3.39 with this PR here.

Generally, things seem to work (well, a `pulp status` is happy, this was my sole test :D) after applying `pulpcore::repo` from this PR, running `dnf upgrade` and applying `pulpcore` and `pulcore::cli` afterwards.

manifests/repo.pp
```diff
@@ -18,23 +18,11 @@
     descr    => "Pulpcore ${version}",
     baseurl  => pick($baseurl, "https://yum.theforeman.org/pulpcore/${version}/${dist_tag}/\$basearch"),
     enabled  => '1',
-    gpgcheck => if $gpgkey == '' { '0' } else { '1' },
+    gpgcheck => if $gpgkey == '' or $version == 'nightly' { '0' } else { '1' },
     gpgkey   => pick($gpgkey, "https://yum.theforeman.org/pulpcore/${version}/GPG-RPM-KEY-pulpcore"),
     notify   => Anchor['pulpcore::repo'],
   }
 
-  # Only EL8 has DNF modules
```
evgeni, line 26: Removing this here, leaves a "failsafe" module in DNF, but seems not hinder the upgrade:

```console
[root@centos8-stream ~]# dnf module list |grep -B5 pulpcore
Last metadata expiration check: 0:04:07 ago on Thu 23 Nov 2023 07:26:27 AM UTC.
@modulefailsafe
Name                 Stream          Profiles                                 Summary                                                                                                                                                                                                                                   
pulpcore             el8 [e]                                                  Pulpcore module                                                                                                                                          
```

The only cleanup needed is to remove the `/etc/dnf/modules.d/pulpcore.module` file, which I think can be done in here?

(People who do not use `pulpcore::repo` still need to be informed via docs, but I think that's fine)

ekohl: > (People who do not use `pulpcore::repo` still need to be informed via docs to either delete that file or call `dnf module reset pulpcore`, but I think that's fine)

https://github.com/theforeman/foreman-documentation/pull/2597

ekohl (approved):
This is now green, so I'm going to merge this now.

ekohl:
I realize now that I only merged this part, which will end up in nightly but I can't merge https://github.com/Katello/katello/pull/10791 because I don't have permissions.

ekohl:
Everything related is now merged

Odilhao:
The Katello release failed for nightly, and it's related to the installer, probably it was not released yet on the nightly package:
```

[2023-11-24T20:42:14.474Z]     Error 1: Puppet Service resource 'pulpcore-api.service' failed. Logs:

[2023-11-24T20:42:14.474Z]       /Service[pulpcore-api.service]

[2023-11-24T20:42:14.474Z]         Starting to evaluate the resource (1386 of 1453)

[2023-11-24T20:42:14.474Z]         Skipping restart; service is not running

[2023-11-24T20:42:14.474Z]         Triggered 'refresh' from 3 events

[2023-11-24T20:42:14.474Z]         The container Systemd::Unit_file[pulpcore-api.service] will propagate my refresh event

[2023-11-24T20:42:14.474Z]         Evaluated in 1.14 seconds

[2023-11-24T20:42:14.474Z]       /Stage[main]/Pulpcore::Service/Pulpcore::Socket_service[pulpcore-api]/Systemd::Unit_file[pulpcore-api.service]/Service[pulpcore-api.service]/ensure

[2023-11-24T20:42:14.474Z]         change from 'stopped' to 'running' failed: Systemd start for pulpcore-api.service failed!

[2023-11-24T20:42:14.474Z]     journalctl log for pulpcore-api.service:

[2023-11-24T20:42:14.474Z]     -- Logs begin at Fri 2023-11-24 20:25:45 UTC, end at Fri 2023-11-24 20:42:05 UTC. --

[2023-11-24T20:42:14.474Z]     Nov 24 20:42:04 pipe-katello-server-nightly-centos8-stream.example.com systemd[1]: Starting Pulp API Server...

[2023-11-24T20:42:14.474Z]     Nov 24 20:42:05 pipe-katello-server-nightly-centos8-stream.example.com pulpcore-api[44109]: Error: This app must be executed using pulpcore-api entrypoint.

[2023-11-24T20:42:14.474Z]     Nov 24 20:42:05 pipe-katello-server-nightly-centos8-stream.example.com systemd[1]: pulpcore-api.service: Main process exited, code=exited, status=1/FAILURE

[2023-11-24T20:42:14.474Z]     Nov 24 20:42:05 pipe-katello-server-nightly-centos8-stream.example.com systemd[1]: pulpcore-api.service: Failed with result 'exit-code'.

[2023-11-24T20:42:14.474Z]     Nov 24 20:42:05 pipe-katello-server-nightly-centos8-stream.example.com systemd[1]: Failed to start Pulp API Server.
```

ekohl:
I investigated this and wrote it down in https://github.com/theforeman/jenkins-jobs/pull/382#issuecomment-1826124215. My initial theory was wrong, and it appears things aren't synced from staging to release.

evgeni:
But the pipes should be using staging anyway. So it shouldn't matter if staging>prod sync is broken?

evgeni:
https://ci.theforeman.org/job/foreman-nightly-rpm-pipeline/2099/consoleFull ran at 9:50PM and pushed the new installer but https://ci.theforeman.org/job/katello-nightly-rpm-pipeline/1837/ ran at 8:18PM and thus just didn't have the new installer (katello doesn't generate foreman staging, but the installer is in foreman).

but when you looked, staging was updated already and confused you :)

evgeni:
and now installs passed fine.

tests are failing tho, see https://community.theforeman.org/t/katello-nightly-rpm-pipeline-1838-failed/35929/2

PR #319, opened 2023-11-15
Allow configuring the yum repo baseurl for Pulp repo

ehelms:


ekohl (commented):
Do we still need the explicit file? I'd think we can switch back to `yumrepo` since we don't need `module_hotfixes` anymore. In https://github.com/theforeman/puppet-pulpcore/pull/320 I took this approach. I think it makes the implementation a bit easier.

ehelms:
Closing in favor of https://github.com/theforeman/puppet-pulpcore/pull/320

PR #323, opened 2023-11-22
Declare cli.toml files as managed via Puppet

quba42:
This would have saved us some confusion (though we did suspect foreman-installer had started overwriting our config pretty quickly).

I am not attached to the precise wording or formatting.

The idea is similar to https://github.com/theforeman/puppet-pulpcore/pull/210 but for the `cli.toml` files written by this module.

Not a Puppet expert so I am opening this as a draft PR in the expectation that it needs further changes to achieve best practice, tests, etc.

### DIFF
--- a/templates/cli-config.toml.epp
+++ b/templates/cli-config.toml.epp
@@ -8,6 +8,7 @@
   Optional[Boolean]         $verify_ssl = undef,
   Optional[Boolean]         $dry_run = undef,
 | -%>
+# This file is managed by puppet-pulpcore. Manual edits will be overwritten with the next run.
 [cli]
 <% if $base_url { -%>
 base_url = "<%= $base_url %>"


templates/cli-config.toml.epp
```diff
@@ -8,6 +8,7 @@
   Optional[Boolean]         $verify_ssl = undef,
   Optional[Boolean]         $dry_run = undef,
 | -%>
+# This file is managed by puppet-pulpcore. Manual edits will be overwritten with the next run.
```
ekohl, line 11: This may not be true. There's a condition on it being managed:
https://github.com/theforeman/puppet-pulpcore/blob/cb95de666838018aa88c5aba92cd39cb3835ec9b/manifests/cli.pp#L61

So what users can do is:
* Run `foreman-installer`
* Set `pulpcore::manage_root_config: false` in `/etc/foreman-installer/custom-hiera.yaml`
* Run `foreman-installer`

Now you have the file `/root/.config/pulp` on disk with this warning, but it's no longer managed.

I don't see a good way around this.

ekohl (changes requested):
Would it help if we exposed the `manage_root_config` parameter as an installer parameter?

templates/cli-config.toml.epp
quba42: Hmm, is it possible to have Puppet check for just the one line and remove it if `pulpcore::manage_root_config: false`?

quba42: I have amended the wording.

PR #326, opened 2024-01-18
Fixes #37062 - Rename pulpcore::telemetry to pulpcore::analytics

wbclark:
Pulp changed the setting name to ANALYTICS. The original name TELEMETRY is deprecated and could be removed in the future. Using pulp's preferred name will also remove the deprecation warning that appears in logs.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -183,11 +183,11 @@
 # @param loggers
 #   Configure additional loggers or override pre-defined logger configuration.
 #
-# @param telemetry
-#   Configure TELEMETRY in settings.yml, which controls the reporting of anonymous data to https://analytics.pulpproject.org/,
+# @param analytics
+#   Configure ANALYTICS in settings.yml, which controls the reporting of anonymous data to https://analytics.pulpproject.org/,
 #   starting with Pulpcore version 3.21.0, to guide Pulp project developers. Set this to false to opt out of this anonymous reporting;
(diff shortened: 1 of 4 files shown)


evgeni:
Which Pulpcore release is the first to support the new name? 3.39?

wbclark:
> Which Pulpcore release is the first to support the new name? 3.39?

It was first introduced as TELEMETRY in 3.21 and renamed to ANALYTICS already in 3.22

evgeni:
Splendid, then it is a breaking change because we change params, but it doesn't influence the supported Pulpcore versions.

wbclark:
Re-running the one test that failed, as the failure appeared transient.

ehelms (approved):

ekohl:
> It was first introduced as TELEMETRY in 3.21 and renamed to ANALYTICS already in 3.22

So we accidentally sent telemetry in 3.22, while we promised we didn't. That's not good.

ehelms:
> > It was first introduced as TELEMETRY in 3.21 and renamed to ANALYTICS already in 3.22
> 
> So we accidentally sent telemetry in 3.22, while we promised we didn't. That's not good.

It was deprecated and a new naming scheme introduced alongside it. The TELEMETRY setting exists still in the main branch (https://github.com/pulp/pulpcore/blob/main/pulpcore/app/util.py#L320-L322).

PR #329, opened 2024-02-15
Ensure glibc-langpack-en is always installed

evgeni:
Not all systems have this package pre-installed and without it creation of the database will fail.


evgeni:
Current example is Almalinux 9

ekohl (approved):
I wonder if we should do the same in our other modules where it's needed (Foreman & Candlepin).

PR #336, opened 2024-03-28
Remove Certguard and File installation as plugin

Odilhao:
It was now merged into pulpcore


manifests/config.pp
```diff
@@ -81,4 +81,28 @@
     mode    => '0640',
     require => Exec['Create database symmetric key'],
   }
+
+  if $use_pulp2_content_route {
```
ekohl, line 85: This needs to be fully qualified
```suggestion
  if $pulpcore::use_pulp2_file_content_route {
```

manifests/init.pp
```diff
@@ -257,6 +257,7 @@
   Optional[Boolean] $analytics = undef,
   Optional[Boolean] $hide_guarded_distributions = undef,
   Optional[Integer[1,100]] $import_workers_percent = undef,
+  Optional[Boolean] $use_pulp2_content_route = false,
```
ekohl, line 260: Other than the lint comments, I'd rename it to indicate it's the file part. Though perhaps it should be `isos` now that `file` is no longer a plugin?
```suggestion
  Optional[Boolean] $use_pulp2_file_content_route = false,
```

ekohl (changes requested):

manifests/plugin/rpm.pp
```diff
@@ -11,11 +11,11 @@
 #   Allow resolving of conflicts due to duplicate advisory ids with different creation dates
 #   https://docs.pulpproject.org/pulp_rpm/settings.html#allow-automatic-unsafe-advisory-conflict-resolution
 class pulpcore::plugin::rpm (
-  Boolean $use_pulp2_content_route = false,
+  Boolean $use_pulp2_file_content_route = false,
```
ekohl, line 14: There's no need to change this. This is **only** scoped to the RPM plugin

ekohl (changes requested):

manifests/plugin/deb.pp
```diff
@@ -1,14 +1,14 @@
 # @summary Pulp Deb plugin
-# @param use_pulp2_content_route
```
Odilhao, line 2: Should we also keep this @ekohl ?

ekohl: I'd imagine so. These are endpoints that Pulp 2 users can use continue using.

ekohl (changes requested):
I think you didn't really understand my comments, so taking a step back.

Previously we had a setting for each plugin. These are set [here](https://github.com/theforeman/puppet-foreman_proxy_content/blob/1c787b10b0d2f3bb4dbe1a7b7095feab03786a07/manifests/init.pp#L284-L298). These compatibility endpoints are not part of Pulp itself, but Katello chooses to keep them for now so users don't have to change URLs on clients. The reason we have booleans is that we can at some point start deprecating them, provide an option to turn them off, default to off and eventually drop them.

If you're dropping the file plugin because it's merged in core, we still want to provide that functionality. That means the parameter needs to move to core. We don't need to change the other plugins if we just move it to `init.pp` and scope it to just the isos endpoint.

Now we can consider a redesign where there is one global parameter in `pulpcore` (so `init.pp`) that plugins can respect, but I'm not sure that's needed.

Also note that once we merge this, we must also modify puppet-foreman_proxy_content.

And this should update `README` to describe the Pulpcore compatibility.

evgeni:
FWIW, we added `Provides` to `python-pulpcore`, so this is not blocking 3.49 anymore.
https://github.com/theforeman/pulpcore-packaging/pull/939

PR #344, opened 2024-05-29
don't add a trailing slash to pulpcore_url in cli tests

evgeni:
new cli is more strict and doesn't accept this


evgeni:
Yeah, I pinged @mdellweg about this, as I think it's too strict.

PR #299, opened 2023-08-10
Support Pulp 3.28

ekohl:
Right now this can't pass because the repositories have not yet been published.

PR #347, opened 2024-07-02
Update puppet_metadata to ~> 4.0 and voxpupuli-acceptance to ~> 3.0

archanaserver:


ekohl:
Looks like an update that starts to require PostgreSQL 13 was done and we don't set that up on EL8 (EL9 defaults to PG 13):
```
  Notice: /Stage[main]/Pulpcore::Database/Pulpcore::Admin[migrate --noinput]/Exec[pulpcore-manager migrate --noinput]/returns: django.db.utils.NotSupportedError: PostgreSQL 12 or later is required (found 10.23).
```

So that isn't related to this PR, but something we should address.

ekohl (changes requested):
Oh, I know why again: the Hiera hierarchy changed in the major version. It's now https://github.com/voxpupuli/voxpupuli-acceptance?tab=readme-ov-file#hiera so you'll need to move `spec/acceptance/hieradata/os`. Effectively reverting c16ddb8acd3c3b7209c122b4e5807cb0e22dafb3

archanaserver:
> PostgreSQL 13

For the new version of PostgreSQL 13, how we can do that? Do we have to define new variable based on OS release or how you preferred to do this @ekohl ?

spec/acceptance/hieradata/family/RedHat-8.yaml
ekohl, line 1: Please follow the hierarchy as specified in https://github.com/voxpupuli/voxpupuli-acceptance/?tab=readme-ov-file#hiera. Now it's not in the right path which means it isn't loaded and causing failures.

ekohl (changes requested):
Please also squash it into a single commit since this is a single change.

ehelms (approved):
LGTM. @ekohl ?

spec/spec_helper_acceptance.rb
```diff
@@ -14,4 +14,27 @@
   end
 end
 
+RSpec.configure do |c|
```
ekohl, line 17: This is not what I meant. I've already tried telling you in https://github.com/theforeman/puppet-pulpcore/pull/347#pullrequestreview-2157476682 that you need to move https://github.com/theforeman/puppet-pulpcore/blob/master/spec/acceptance/hieradata/os/RedHat/8.yaml because version 3 of `voxpupuli-acceptance` ends up using a different hierarchy. Rather than overriding the hierarchy to the old style, you should move the file.

In practice what you need to do is exactly the reverse of https://github.com/theforeman/puppet-pulpcore/commit/c16ddb8acd3c3b7209c122b4e5807cb0e22dafb3.

archanaserver: @ekohl I am not sure what I was missing out but isn't my this change https://github.com/archanaserver/puppet-pulpcore/commit/9f884fb511707c7721503bf9ebc03eb2890b1d21 was exactly reverting https://github.com/theforeman/puppet-pulpcore/pull/299/commits/4248c03610f38a30133daf9390a9fc2a52ba90e0 one?

Also I'm not sure if I did the right thing now, but i did revert this specific https://github.com/theforeman/puppet-pulpcore/commit/c16ddb8acd3c3b7209c122b4e5807cb0e22dafb3 and pushed it and I can see this now "Skipping EOL operating system CentOS 8"

ekohl:
As discussed elsewhere: https://github.com/theforeman/puppet-pulpcore/commit/c16ddb8acd3c3b7209c122b4e5807cb0e22dafb3 used a dash in `RedHat-8.yaml` but according to https://github.com/voxpupuli/voxpupuli-acceptance?tab=readme-ov-file#hiera it needs to be a file in the subdirectory (so a slash).

archanaserver:
@ekohl it is passing all the tests now. thanks!

PR #348, opened 2024-07-12
Move Hiera file back to family/RedHat-8.yaml

archanaserver:
Ref: https://github.com/theforeman/puppet-pulpcore/pull/347#pullrequestreview-2157476682


ekohl:
This is part of https://github.com/theforeman/puppet-pulpcore/pull/347 because of the gem version update. It must be in the same PR.

PR #352, opened 2024-08-14
explicitly support Pulpcore 3.49

evgeni:


manifests/repo.pp
```diff
@@ -8,7 +8,7 @@
 #   An optional value for gpgkey to be used for yumrepo, instead of the default.
 #   If an empty string is passed, gpgcheck will be disabled.
 class pulpcore::repo (
-  Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $version = '3.39',
+  Variant[Enum['nightly'], Pattern['^\d+\.\d+$']] $version = '3.49',
```
evgeni, line 11: does this make it a breaking change? :thinking:

Odilhao: I don't see why it would

evgeni: people using `include pulpcore::repo` will get upgraded to 3.49?

Odilhao: We should be able to safe drop 3.39 when the new katello get branched.

ekohl: I'd say it technically is a backwards incompatible change: it's declared as a public API and changes behavior in something that would be surprising to me as a user.

Short term I'd prefer to not change this line to avoid a major version bump. Then after that we can drop the default version. Then we may need to add some fallback here:
https://github.com/theforeman/puppet-pulpcore/blob/3258e7828678a1038392bdd246720ea34b8bb25c/spec/setup_acceptance_node.pp#L2
Because that's the biggest reason we have a default: so you can simply run beaker locally without specifying a version and still get some test suite to run. We could make that default to nightly.

evgeni: ack, done

ekohl (approved):

PR #354, opened 2024-08-27
Reuse headers from pulpcore::apache class

ekohl:
We have the exact same code in both pulpcore::plugin::container as in pulpcore::apache. This reuses the variables rather than duplicating the logic.

Fixes: 21aa39e28f19 ("Fixes #37308 - set REMOTE_USER properly for pulpcore registry")

### DIFF
--- a/manifests/plugin/container.pp
+++ b/manifests/plugin/container.pp
@@ -9,14 +9,6 @@ class pulpcore::plugin::container (
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
 ) {
-  $api_default_request_headers = [
-    "unset ${pulpcore::apache::remote_user_environ_header}",
-  ]
-
-  $api_additional_request_headers = $pulpcore::api_client_auth_cn_map.map |String $cn, String $pulp_user| {
-    "set ${pulpcore::apache::remote_user_environ_header} \"${pulp_user}\" \"expr=%{SSL_CLIENT_S_DN_CN} == '${cn}'\""
-  }
-
   $context = {
     'directories' => [
       {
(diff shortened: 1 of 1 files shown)


evgeni (approved):

manifests/plugin/container.pp
```diff
@@ -27,7 +19,7 @@
             'url' => "${pulpcore::apache::api_base_url}${registry_version_path}",
           },
         ],
-        'request_headers' => $api_default_request_headers + $api_additional_request_headers,
+        'request_headers' => $pulpcore::apache::api_default_request_headers + $pulpcore::apache::api_additional_request_headers,
```
ekohl, line 22: I debated introducing a new variable that has both these values, but decided against it to keep this smaller.

ekohl:
When adding tests I figured out that at least there it's somehow different.

Redoing it, I first wrote tests prior to this patch (e3d1ddb169713d6e2b68ff99abe0cf4b3fdd720c, based on @evgeni's work elsewhere). Then reapplied this (bd1dbecec0b47316a57b8d18aa17f72c2d666bce) and fixed tests the (66b97fed46a4ce506ec475ad37ee1ef38b2fe7d2).

Why does it result in an additional line when it should use the exact same variables? Still trying to figure that out.

evgeni:
> Why does it result in an additional line when it should use the exact same variables? Still trying to figure that out.

Because the content of the variables are not identical!

The "global" ones:
https://github.com/theforeman/puppet-pulpcore/blob/5b6d0885a44d5b5027f15c866de9d7c39e5f71b0/manifests/apache.pp#L48-L55

The ones you removed here:
https://github.com/theforeman/puppet-pulpcore/blob/c63cb654ab2ac78c3cc4828b1fe504aa2d3bb577/manifests/plugin/container.pp#L12-L18

the global ones, that we now use have an additional
```
"set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN", 
```

ekohl:
I was so blindly staring at `api_additional_request_headers` that I didn't look too carefully at `api_default_request_headers`.

ekohl:
https://github.com/theforeman/puppet-pulpcore/pull/356

PR #356, opened 2024-09-03
Correct container plugin apache header rewrite

ekohl:
In 5b6d0885a44d it was assumed that `$api_default_request_headers` was the same, but it was slightly different. This corrects it and adds tests, which should have been added in the first place. These tests pass prior to the application and after, making sure there's no change.

Fixes: 5b6d0885a44d ("Reuse headers from pulpcore::apache class")

### DIFF
--- a/manifests/plugin/container.pp
+++ b/manifests/plugin/container.pp
@@ -9,6 +9,11 @@ class pulpcore::plugin::container (
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
 ) {
+  # This is like pulpcore::apache's value, but slightly different
+  $api_default_request_headers = [
+    "unset ${pulpcore::apache::remote_user_environ_header}",
+  ]
+
   $context = {
     'directories' => [
       {
@@ -19,7 +24,7 @@ class pulpcore::plugin::container (
             'url' => "${pulpcore::apache::api_base_url}${registry_version_path}",
           },
         ],
-        'request_headers' => $pulpcore::apache::api_default_request_headers + $pulpcore::apache::api_additional_request_headers,
+        'request_headers' => $api_default_request_headers + $pulpcore::apache::api_additional_request_headers,
       },
     ],
     'proxy_pass'  => [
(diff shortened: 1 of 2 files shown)


manifests/plugin/container.pp
```diff
@@ -9,6 +9,11 @@
   String $location_prefix = '/pulpcore_registry',
   String $registry_version_path = '/v2/',
 ) {
+  # This is like pulpcore::apache's value, but slightly different
+  $api_default_request_headers = [
```
evgeni, line 13: Do we really need this?

https://httpd.apache.org/docs/current/mod/mod_headers.html#requestheader says

set
    The request header is set, replacing any previous header with this name

So while the content of the file is changed, the behavior should not?

(Unless the cn map is empty)

ekohl: The CN map contains "if" expressions. My understanding is the first line 100% certain clears it. Then IF the cert's CN equals something, it sets the header.

evgeni: Right, so the flow is:
* use `unset` to remove *any* header by that name, wherever it comes from
* use `set` to set it to the CN (if there is a CN!)
* use `set` to set it to `admin` (if the expression matches, resetting a possibly previously set CN-based name)

I honestly see no reason (ever) to set the CN-based remote user, as such user just doesn't exist in Pulp (in our deployment)

evgeni: It was added in ~~https://github.com/theforeman/puppet-pulpcore/pull/186~~ https://github.com/theforeman/puppet-pulpcore/pull/105, but I don't really see a reason for it.

evgeni: My vote: don't add this complexity (same variable name, different values) back in, it's confusing as hell.
Do the releases we need to do.

Drop CN-stuff in an 11.x

ekohl: Should we then include 11.x in the upcoming releases? I'd actually feel more comfortable if we only had the explicit unset and explicit mapping.

evgeni: I'm not sure…

evgeni: fwiw, I tested the current (two-`set`) config with https://github.com/theforeman/forklift/pull/1853 and it pushes containers fine, so :woman_shrugging:

ekohl: I'm leaning to merging this now as is because it keeps it minimal and then resolve it properly in the next release cycle.

evgeni (approved):

ekohl:
Skipping per https://github.com/theforeman/puppet-pulpcore/pull/358#discussion_r1743896983

PR #359, opened 2024-09-12
Add Wants=postgresql.service to Pulpcore service files

ekohl:
Wants is a weak requirement, so the service won't fail it can't be started but systemd will at least try it. That should work with help when postgresql is local.

This was reported in https://github.com/theforeman/foreman_maintain/issues/824. I don't know if this is sufficient or if foreman_maintain also needs to be smarter.

An open question: if we know PostgreSQL is managed, should we make it a hard `Requires=` instead?

### DIFF
--- a/templates/pulpcore-worker@.service.erb
+++ b/templates/pulpcore-worker@.service.erb
@@ -1,7 +1,7 @@
 [Unit]
 Description=Pulp Worker
 After=network-online.target
-Wants=network-online.target
+Wants=network-online.target postgresql.service
 
 [Service]
 Type=simple
--- a/templates/pulpcore-api.service.erb
+++ b/templates/pulpcore-api.service.erb
@@ -2,6 +2,7 @@
 Description=Pulp API Server
 After=network.target
 Requires=pulpcore-api.socket
+Wants=postgresql.service
 
 [Service]
 Type=notify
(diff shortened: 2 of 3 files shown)


evgeni (approved):

evgeni:
I'd not add a hard requires

templates/pulpcore-worker@.service.erb
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Pulp Worker
 After=network-online.target
-Wants=network-online.target
+Wants=network-online.target postgresql.service
```
ekohl, line 4: Reading this file I debated adding Redis as well, but wasn't quite sure about it.

evgeni: Makes sense (also to foreman and dynflow)

sbernhard (approved):
Make sense!

ianballou:
A user reported on Discourse that this wasn't sufficient to stop the coredumps when stopping Pulpcore: https://community.theforeman.org/t/pulpcore-coredumps-during-stopping-services/36512/6

ehelms:
We've identified this is still useful even if it does not address the particular problem discussed, which @evgeni has made headway with actually fixing through packaging.

PR #360, opened 2024-10-08
never set REMOTE_USER to the value of SSL_CLIENT_S_DN_CN

evgeni:
We only deploy a single user in Pulp: admin
And we do not give out certs with CN=admin, so there is no point in trying to obtain the REMOTE_USER from the CN.


manifests/apache.pp
```diff
@@ -50,7 +50,6 @@
   $api_default_request_headers = [
     "unset ${remote_user_environ_header}",
     "unset ${remote_user_environ_header_underscore}",
-    "set ${remote_user_environ_header} \"%{SSL_CLIENT_S_DN_CN}s\" env=SSL_CLIENT_S_DN_CN",
```
evgeni, line 53: Technically, this is a breaking change. At the same time, this never worked so I'd err on the side of "enhancement"?

evgeni: lol, OUR TESTS WERE USING THIS?!

spec/acceptance/basic_spec.rb
```diff
@@ -4,7 +4,9 @@
   it_behaves_like 'an idempotent resource' do
     let(:manifest) do
       <<-PUPPET
-      include pulpcore
+      class { 'pulpcore':
+        api_client_auth_cn_map => {${facts['networking']['fqdn']} => 'admin'},
```
evgeni, line 8: I wonder if we should set that in `spec/acceptance/hieradata/common.yaml` instead?

manifests/apache.pp
ekohl: I have a vague recollection that we used to create the certs with `cn=admin` and the acceptance tests copied that. When we changed it in certs it wasn't reflected here

evgeni:
Do I want to know why this fails everywhere but CS9?!

evgeni:
> > And we do not give out certs with CN=admin, so there is no point in trying to obtain the REMOTE_USER from the CN.
> 
> We used to, and when we introduced the current method we had to keep the old configuration around for compatibility due to N-1 Capsules.

But wasn't this for Pulp2 stuff only? At least that's how decipher https://projects.theforeman.org/issues/35004.

Anyway. Old enough to be dropped.

ehelms:
> Anyway. Old enough to be dropped.

^ this

manifests/apache.pp
ekohl: I did just mark it as backwards incompatible just so it's clear we're changing behavior. Recently we had some cherry picks with bumps and I don't want to accidentally include this in an existing release.

ekohl (approved):

spec/setup_acceptance_node.pp
```diff
@@ -34,14 +34,14 @@
...
-  command   => "openssl req -nodes -new -newkey rsa:2048 -subj '/CN=admin' -out '${client_csr}' -keyout '${client_key}'",
+  command   => "openssl req -nodes -new -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -addext 'subjectAltName = DNS:${facts['networking']['fqdn']}' -out '${client_csr}' -keyout '${client_key}'",
   path      => ['/bin', '/usr/bin'],
   creates   => $client_csr,
   logoutput => 'on_failure',
   umask     => '0022',
 }
 -> exec { 'Sign CSR':
-  command   => "openssl x509 -req -days 360 -in '${client_csr}' -CA '${ca_cert}' -CAkey '${ca_key}' -CAcreateserial -out '${client_cert}'",
+  command   => "openssl x509 -req -days 360 -in '${client_csr}' -CA '${ca_cert}' -CAkey '${ca_key}' -CAcreateserial -out '${client_cert}' -extfile <(echo 'subjectAltName = DNS:${facts['networking']['fqdn']}')",
```
evgeni, line 44: openssl (pre 3.2, which is what CS9 has, and that's what has been passing) doesn't copy extensions from the CSR, so the resulting cert ends up being invalid.

evgeni: and no, I have no idea why the `admin` one worked before…

spec/setup_acceptance_node.pp
```diff
@@ -19,7 +19,7 @@
 $client_cert = "${directory}/client-cert.pem"
 $client_key = "${directory}/client-key.pem"
 
-$ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']}' -addext 'subjectAltName = DNS:${facts['networking']['fqdn']}' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
+$ca_cmd = "openssl req -nodes -x509 -newkey rsa:2048 -subj '/CN=${facts['networking']['fqdn']} CA' -keyout '${ca_key}' -out '${ca_cert}' -days 365"
```
evgeni, line 22: the CA and the client cert can't have the same CN :fire:

spec/acceptance/hieradata/common.yaml
```diff
@@ -1,6 +1,8 @@
 ---
 apache::default_mods: false
-pulpcore::apache_https_cert: '/etc/pulpcore-certs/ca-cert.pem'
-pulpcore::apache_https_key: '/etc/pulpcore-certs/ca-key.pem'
+pulpcore::apache_https_cert: '/etc/pulpcore-certs/client-cert.pem'
+pulpcore::apache_https_key: '/etc/pulpcore-certs/client-key.pem'
```
evgeni, line 4: neither is truly right, but I was too lazy to crate a `server-*.pem`.

evgeni:
If y'all think it's beneficial, I could split out the setup changes into an own PR (and add a dedicated server-cert), and then this would/could be reduced to the original "drop the CN=admin handling"?

spec/setup_acceptance_node.pp
```diff
@@ -34,14 +35,21 @@
...
   logoutput => 'on_failure',
   umask     => '0022',
 }
+-> exec { 'Create extfile':
+  command   => "echo 'subjectAltName = DNS:${facts['networking']['fqdn']}' > '${client_ext}'",
+  path      => ['/bin', '/usr/bin'],
+  creates   => $client_ext,
+  logoutput => 'on_failure',
+  umask     => '0022',
+}
```
ekohl, line 50: Isn't this equal to
```suggestion
-> file { $client_ext:
  content   => "subjectAltName = DNS:${facts['networking']['fqdn']}",
  mode      => '0644',
}
```

spec/setup_acceptance_node.pp
evgeni: yeah, we can also do that

spec/setup_acceptance_node.pp
ekohl: I think you were too lazy to say that

ehelms (approved):

PR #361, opened 2024-10-16
properly escape quotes in passwords by calling to_python

evgeni:
database passwords can contain special characters, especially " and '
so we can't just print the value of the field enclosed by single quotes
as that would break whenever the user uses a literal ' in their password

### DIFF
--- a/templates/settings.py.erb
+++ b/templates/settings.py.erb
@@ -24,7 +24,7 @@ DATABASES = {
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': '<%= scope['pulpcore::postgresql_db_name'] %>',
         'USER': '<%= scope['pulpcore::postgresql_db_user'] %>',
-        'PASSWORD': '<%= scope['pulpcore::postgresql_db_password'] %>',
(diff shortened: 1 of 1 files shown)


templates/settings.py.erb
```diff
@@ -24,7 +24,7 @@ DATABASES = {
         'ENGINE': 'django.db.backends.postgresql',
         'NAME': '<%= scope['pulpcore::postgresql_db_name'] %>',
         'USER': '<%= scope['pulpcore::postgresql_db_user'] %>',
-        'PASSWORD': '<%= scope['pulpcore::postgresql_db_password'] %>',
+        'PASSWORD': <%= scope.call_function('to_python', [scope['pulpcore::postgresql_db_password']]) %>,
```
bastelfreak, line 27: is that the function from stdlib? Then maybe use the new namespaced version?
```suggestion
        'PASSWORD': <%= scope.call_function('stdlib::to_python', [scope['pulpcore::postgresql_db_password']]) %>,
```

evgeni: it is, the rest of the file is no using the namespaced version, so I didn't want to diverge

bastelfreak (approved):

ekohl (approved):

PR #364, opened 2024-12-17
always compare CNs as downcase

evgeni:
Sometimes people end up with certificates that have uppercase letters in
the CN, but pass lowercase in the auth map.

### DIFF
--- a/manifests/apache.pp
+++ b/manifests/apache.pp
@@ -53,7 +53,7 @@ class pulpcore::apache (
   ]
 
   $api_additional_request_headers = $pulpcore::api_client_auth_cn_map.map |String $cn, String $pulp_user| {
-    "set ${remote_user_environ_header} \"${pulp_user}\" \"expr=%{SSL_CLIENT_S_DN_CN} == '${cn}'\""
+    "set ${remote_user_environ_header} \"${pulp_user}\" \"expr=%{tolower:%{SSL_CLIENT_S_DN_CN}} == '${cn.downcase}'\""
   }
 
   $api_directory = {
--- a/spec/classes/plugin_container_spec.rb
+++ b/spec/classes/plugin_container_spec.rb
@@ -53,7 +53,7 @@ APACHE_CONFIG
   <Location "/pulpcore_registry">
     RequestHeader unset REMOTE-USER
     RequestHeader unset REMOTE_USER
-    RequestHeader set REMOTE-USER "admin" "expr=%{SSL_CLIENT_S_DN_CN} == 'foreman.example.com'"
+    RequestHeader set REMOTE-USER "admin" "expr=%{tolower:%{SSL_CLIENT_S_DN_CN}} == 'foreman.example.com'"
     ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api
     ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api
   </Location>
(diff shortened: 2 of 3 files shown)


manifests/apache.pp
```diff
@@ -53,7 +53,7 @@
   ]
 
   $api_additional_request_headers = $pulpcore::api_client_auth_cn_map.map |String $cn, String $pulp_user| {
-    "set ${remote_user_environ_header} \"${pulp_user}\" \"expr=%{SSL_CLIENT_S_DN_CN} == '${cn}'\""
+    "set ${remote_user_environ_header} \"${pulp_user}\" \"expr=%{tolower:SSL_CLIENT_S_DN_CN} == '${cn.downcase}'\""
```
evgeni, line 56: There is another occurence of this in https://github.com/theforeman/puppet-foreman_proxy_content/blob/e24f94194ebd7858f4ce9c34e5a671a30009873c/manifests/container.pp#L26-L27 which I don't understand where it's used from

ekohl (approved):
Yesterday I looked at the same problem and came to the same solution, just didn't have time to test it.

evgeni:
The problem is that in our Puppet code we (mostly) use the `networking.fqdn` fact, which will return whatever the system thinks about itself:
```console
# facter networking.fqdn
centos9-STREAM.tanso.example.com
```

However, in [some places](https://github.com/theforeman/foreman-installer/blob/d4a78ffa234286a1f595b5f47665f9fd74f581ee/katello_certs/hooks/boot/02-message-helpers.rb#L23-L27) we use `hostname -f` output, which will under certain circumstances DO SOMETHING COMPLETELY DIFFERENT:
```console
# hostname -f
centos9-stream.tanso.example.com
```

The circumstances being `/etc/hosts` (or DNS, I guess?) containing a different casing of the name:
```
# grep centos /etc/hosts 
127.0.1.1 centos9-StrEam.tanso.example.com centos9-stream

# hostname -f
centos9-StrEam.tanso.example.com
```

Now, if you have a system that thinks it's `MACHINE.example.com` and got a cert with `CN=MACHINE.example.com` but you use `--foreman-proxy-trusted-hosts machine.example.com` when setting up a proxy using that machine as the server… You're gonna have a bad time.

manifests/apache.pp
ekohl: That looks interesting. That code only looks correct for content proxies, not Foreman itself.

evgeni: container gateway is a proxy-only feature :)

ekohl: Oh yes, that isn't `pulpcore::plugin::container`. It's used here: https://github.com/theforeman/puppet-foreman_proxy_content/blob/e24f94194ebd7858f4ce9c34e5a671a30009873c/manifests/init.pp#L254-L256

ekohl (approved):
I'm debating a Redmine issue for proper bug tracking.

evgeni:
> I'm debating a Redmine issue for proper bug tracking.

I can certainly create one, sure.

Should we also try to fix up the installer to use networking.fqdn if possible?

evgeni:
FWIW, the "uppercase letter in fqdn" should be guarded by `checks/hostname.rb` in the installer, but old certs might still exist :see_no_evil:

ekohl:
> Should we also try to fix up the installer to use networking.fqdn if possible?

Probably a good idea.

> FWIW, the "uppercase letter in fqdn" should be guarded by `checks/hostname.rb` in the installer, but old certs might still exist 🙈

I still see that check as a workaround for known and unknown bugs. Ideally we'll fix cases like these over time and drop the check.

PR #372, opened 2025-01-22
Drop support for Pulpcore 3.49

evgeni:


evgeni:
I wasn't sure about this one, do we want to drop .49 already? *technically* we should, as we dropped EL8 and .49 supports EL8, but…

ekohl:
> I wasn't sure about this one, do we want to drop .49 already? _technically_ we should, as we dropped EL8 and .49 supports EL8, but…

I think we can say we're only EL9 and still 3.49 compatible if 3.49 works on EL9.

evgeni:
it's time…

PR #373, opened 2025-01-28
use the right username when computing the db password hash

evgeni:
some people use other names than `pulp` for their DB user

### DIFF
--- a/manifests/database.pp
+++ b/manifests/database.pp
@@ -11,7 +11,7 @@ class pulpcore::database (
     ensure_packages(['glibc-langpack-en'])
     postgresql::server::db { $pulpcore::postgresql_db_name:
       user     => $pulpcore::postgresql_db_user,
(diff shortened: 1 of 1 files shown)


ekohl (approved):

wbclark (commented):
Interesting, how did this come up?

Probably no point to add Refs #28901 now given how old that feature is. Is there some newer associated Redmine?

evgeni:
> Interesting, how did this come up?

user had external db deployed with `pulpcore` as the username, now they are moving back to internal and when setting `manage=true` they ended up with not being able to talk to the db anymore.

evgeni:
Oh, and to add a pinch of mystery to this: You need to use MD5-auth (so Foreman 3.10 or older) to *actually* run into a problem here.
With SCRAM the code here would still use the "wrong" salt, but that salt is properly stored in PostgreSQL (and can actually be *anything*!) and thus authentication works.

PR #379, opened 2025-03-31
Support s3 storage

ekohl:
Pulp supports multiple storage backends via the django-storages package. This exposes both the file storage (default) and s3.

No specific s3 options are exposed, but rather the whole options hash. This makes it easily extensible, both for specific s3 tuning and for future expansion.

It depends on the django-storages Python package in install. This is too specific now, but changing that would require a packaging change.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -41,6 +41,20 @@ class pulpcore::config {
     order   => '02',
   }
 
+  if $pulpcore::storage_backend == 's3' {
+    $storages = {
+      'default' => {
+        'BACKEND' => 'storages.backends.s3.S3Storage',
+        'OPTIONS' => $pulpcore::storage_options,
+      },
+    }
+    concat::fragment { 'storage':
+      target  => 'pulpcore settings',
+      content => "STORAGES = ${stdlib::to_python($storages)}\n",
+      order   => '03',
+    }
+  }
+
   file { $pulpcore::user_home:
     ensure => directory,
     owner  => $pulpcore::user,
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -12,6 +12,13 @@ class pulpcore::install {
     }
   }
 
+  if $pulpcore::storage_backend != 'file' {
+    # TODO: better virtual for this
+    package { 'python3.11-django-storages':
+      ensure => present,
+    }
+  }
+
   user { $pulpcore::user:
     ensure     => present,
     system     => true,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -203,6 +203,12 @@
 #   What percentage of available-workers will pulpcore use for import tasks at a time.
 #   By default, pulpcore will use all available workers.
 #
+# @param storage_backend
+#   Which storage backend to use
+#
+# @param storage_options
+#   Options to pass to the storage backend
+#
 # @example Default configuration
 #   include pulpcore
 #
@@ -268,6 +274,8 @@ class pulpcore (
   Optional[Boolean] $analytics = undef,
   Optional[Boolean] $hide_guarded_distributions = undef,
   Optional[Integer[1,100]] $import_workers_percent = undef,
+  Enum['file', 's3'] $storage_backend = 'file',
+  Hash[String[1], Any] $storage_options = {},
 ) {
   $settings_file = "${config_dir}/settings.py"
   $certs_dir = "${config_dir}/certs"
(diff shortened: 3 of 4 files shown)


manifests/install.pp
```diff
@@ -12,6 +12,13 @@
     }
   }
 
+  if $pulpcore::storage_backend != 'file' {
+    # TODO: better virtual for this
+    package { 'python3.11-django-storages':
```
ekohl, line 17: @Odilhao any thoughts on how we should tackle this in packaging? I thought about a `pulpcore-storages` subpackage or multiple to reflect the extras in https://github.com/pulp/pulpcore/blob/528efde666ef49b6a3f0815e6084419750ee2314/pyproject.toml#L74-L77.

ekohl: https://github.com/theforeman/pulpcore-packaging/pull/1760

Odilhao: We can use subpackages as we use for cachecontrol

```
%description -n python%{python3_pkgversion}-%{srcname}+filecache
This is a metapackage bringing in filecache extra requires for python%{python3_pkgversion}-%{srcname}
It contains no code, just makes sure the dependencies are installed.

%files -n python%{python3_pkgversion}-%{srcname}+filecache
%ghost %{python3_sitelib}/%{srcname}-%{version}.dist-info/
```

This code will generate the following package:
`Provides: python3.12-cachecontrol+filecache = 0.14.2-4.el9 python3.12dist(cachecontrol[filecache]) = 0.14.2`

We can use the rich decency package in this case `python3.12dist(cachecontrol[filecache])` or in pulpcore logic 
`python3.12dist(pulpcore[s3])`

For pulpcore I see the package like this:

```
%package -n python%{python3_pkgversion}-%{pypi_name}+s3
Summary: Metapackage for python%{python3_pkgversion}-%{pypi_name}: s3 extra
Requires: python%{python3_pkgversion}-django-storage+s3 >= 3.8.0

%description -n python%{python3_pkgversion}-%{pypi_name}+s3
This is a metapackage bringing in s3 extra requires for python%{python3_pkgversion}-%{pypi_name}
It contains no code, just makes sure the dependencies are installed.

%files -n python%{python3_pkgversion}-%{pypi_name}+s3
%ghost %{python3_sitelib}/%{pypi_name}-%{version}.dist-info/

```

This will also require us to add the extra package into django-storage.

1 - https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#Extras

Odilhao: I don't see a problem with `Provides: pulpcore-storage-s3` we can add that to the metapackage

manifests/config.pp
```diff
@@ -41,6 +41,20 @@
     order   => '02',
   }
 
+  if $pulpcore::storage_backend == 's3' {
+    $storages = {
+      'default' => {
+        'BACKEND' => 'storages.backends.s3.S3Storage',
+        'OPTIONS' => $pulpcore::storage_options,
+      },
```
alexjfisher, line 49: @ekohl Any plans to resurrect this PR?  I'd be interested.  I'm currently trying to get this working by taking what you've done and including it in my profile.

Currently failing when running `pulpcore-manager collectstatic --noinput` with...

```
django.core.files.storage.handler.InvalidStorageError: Could not find config for 'staticfiles' in settings.STORAGES.
```

Looks like a `staticfiles` config is also needed here?

From https://pulpproject.org/pulpcore/docs/admin/guides/configure-pulp/configure-storages/#example_1 it also appears that it might be necessary to set `MEDIA_ROOT` to an empty string, (and the top-level param doesn't allow this currently, but maybe best to just automatically do this in the template anyway)??

(_also_ https://pulpproject.org/pulpcore/docs/admin/reference/settings/#media_root suggests that maybe the setting will just be ignored???)

ekohl: I don't have an immediate plan to continue this. I consider it blocked on the packaging changes.

Edit: that's not to say I wouldn't love to see this finished. I just don't have the time for it now.

Odilhao: I can take a look this week on the packaging side PR, it was on my plate to clean up that backlog of old PRs.

alexjfisher: If that can be sorted, I can probably pick this PR for @ekohl .  I've also been working on a few types/providers for creating rpm repos, remotes etc.  They're not quite ready yet, but I'd like to contribute those when they are.

PR #384, opened 2025-06-25
Drop EnvironmentFile from pulpcore-worker@.service

evgeni:
We never shipped anything in these files and never supported them
officially.


evgeni:
Mostly a "why is it here" while looking at other things in that file (see yesterdays discussion in `#theforeman-dev`).
But yes, it'd be cool if our deployment methods would be aligned.

ekohl (approved):

PR #400, opened 2026-04-17
Add apache config support for /pypi

pavanshekar:
## Summary

Add Apache config support for `/pypi` in puppet-pulpcore

Fixes: [SAT-44474](https://redhat.atlassian.net/browse/SAT-44474)

## Changes                                                                                                                                   
  - Added `use_pypi_route` parameter to `pulpcore::plugin::python` (default: `true`)                                                           
  - Generates Apache `<Location "/pypi">` block that proxies to `pulpcore-api` service
  - Added comprehensive tests for the new parameter

## Generated Apache Config

When `use_pypi_route => true` (default), the module should generate:
```apache                                                                                                                                    
  <Location "/pypi">                                                                                                                           
    RequestHeader unset X-CLIENT-CERT                                                                                                          
    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT                                                                  
    RequestHeader set X-FORWARDED-PROTO expr=%{REQUEST_SCHEME}                                                                                 
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/pypi timeout=600
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/pypi                                                                    
  </Location>
```

### DIFF
--- a/manifests/plugin/python.pp
+++ b/manifests/plugin/python.pp
@@ -1,4 +1,31 @@
 # @summary Pulp Python plugin
 class pulpcore::plugin::python {
-  pulpcore::plugin { 'python': }
+  $pypi_path = '/pypi'
+  $pypi_url = "${pulpcore::apache::api_base_url}${pypi_path}"
+
+  $context = {
+    'directories' => [
+      {
+        'provider'        => 'location',
+        'path'            => $pypi_path,
+        'proxy_pass'      => [
+          {
+            'url'    => $pypi_url,
+            'params' => $pulpcore::apache::api_proxy_params,
+          },
+        ],
+        'request_headers' => [
+          'unset X-CLIENT-CERT',
+          'set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT',
+          'set X-FORWARDED-PROTO expr=%{REQUEST_SCHEME}',
+        ],
+      },
+    ],
+  }
+  $content = epp('pulpcore/apache-fragment.epp', $context)
+
+  pulpcore::plugin { 'python':
+    http_content  => $content,
+    https_content => $content,
+  }
 }
--- a/spec/classes/plugin_python_spec.rb
+++ b/spec/classes/plugin_python_spec.rb
@@ -4,20 +4,30 @@ describe 'pulpcore::plugin::python' do
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let(:facts) { os_facts }
+      let(:pre_condition) { 'include pulpcore' }
 
       it { is_expected.to compile.with_all_deps }
       it { is_expected.to contain_package('pulpcore-plugin(python)') }
       it { is_expected.to contain_pulpcore__plugin('python') }
 
-      context 'with pulpcore' do
-        let(:pre_condition) { 'include pulpcore' }
+      it do
+        is_expected.to contain_pulpcore__plugin('python')
+          .that_subscribes_to('Class[Pulpcore::Install]')
+          .that_notifies(['Class[Pulpcore::Database]', 'Class[Pulpcore::Service]'])
+      end
 
-        it do
-          is_expected.to compile.with_all_deps
-          is_expected.to contain_pulpcore__plugin('python')
-            .that_subscribes_to('Class[Pulpcore::Install]')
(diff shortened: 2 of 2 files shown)


manifests/plugin/python.pp
```diff
@@ -1,4 +1,38 @@
 # @summary Pulp Python plugin
-class pulpcore::plugin::python {
-  pulpcore::plugin { 'python': }
+# @param use_pypi_route
+class pulpcore::plugin::python (
+  Boolean $use_pypi_route = true,
```
ekohl: When would a user set this to `false`? In other places we have a boolean for legacy routes (for Pulp 2) but this looks more like core functionality.

pavanshekar: Thanks for pointing that out! I was following the deb plugin pattern but missed the distinction between legacy Pulp 2 routes and core Pulp 3 functionality. I've removed the parameter and the /pypi route is now always enabled.

spec/classes/plugin_python_spec.rb
```diff
@@ -4,20 +4,33 @@
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
       let(:facts) { os_facts }
+      let(:pre_condition) { 'include pulpcore' }
```
ekohl, line 7: Please don't move just this precondition. If the class no longer works without the precondition then a lot of assertions after become redundant and you can merge both contexts. Since you now always need it, I think merging both contexts is better.

spec/classes/plugin_python_spec.rb
```diff
@@ -4,20 +4,33 @@
...
         end
+
+        it do
+          is_expected.to contain_pulpcore__plugin('python')
+            .with_http_content(%r{<Location "/pypi">})
+            .with_https_content(%r{<Location "/pypi">})
+        end
+
+        it do
+          content = catalogue.resource('pulpcore::plugin', 'python').send(:parameters)[:http_content]
```
ekohl: I'm not a fan of this construct and above you already use the proper construct: `is_expected.to contain_pulpcore__plugin('python').with_http_content`. I think you can repeat it since internally it's just iterating a list of tuples, not a hash:
```ruby
          is_expected.to contain_pulpcore__plugin('python')
            .with_http_content(%r{<Location "/pypi">})
            .with_http_content(%r{RequestHeader unset X-CLIENT-CERT})
```

ekohl (changes requested):

spec/classes/plugin_python_spec.rb
pavanshekar: I've removed the nested context since the pre_condition is always needed now. All tests are at the same level.

spec/classes/plugin_python_spec.rb
pavanshekar: I've flattened the structure and moved everything to the top level since we always include pulpcore.

qcjames53 (approved):
I was (**finally**) able to get this to run as a part of foreman-installer on a clean dev environment. After pulling changes and re-running installer, the Apache config successfully includes the new endpoint:
```
[vagrant@centos9-pulp-upgrade ~]$ sudo cat /etc/httpd/conf.d/05-foreman-ssl.conf
...
  <Location "/pypi">
    RequestHeader unset X-CLIENT-CERT
    RequestHeader set X-CLIENT-CERT "%{SSL_CLIENT_CERT}s" env=SSL_CLIENT_CERT
    RequestHeader set X-FORWARDED-PROTO expr=%{REQUEST_SCHEME}
    ProxyPass unix:///run/pulpcore-api.sock|http://pulpcore-api/pypi timeout=600
    ProxyPassReverse unix:///run/pulpcore-api.sock|http://pulpcore-api/pypi
  </Location>
```

The unit test passes and looks fit for purpose. The source seems good to me as well. Approving changes. @ekohl how are things looking to you?

ekohl (approved):
Looks good. I've kicked off CI to see if it agrees.

pavanshekar:
CI is failing. What should I do to resolve this?

qcjames53:
The failure in test setup looks unrelated to me.

ekohl:
I think this is due to https://github.com/voxpupuli/gha-puppet/commit/fd171ad3be28029f16e54a8aaf039f906c0aaa04.

evgeni (approved):

PR #403, opened 2026-04-30
Set explicit gunicorn control socket path for API and content services

Odilhao:
## Summary

gunicorn 25.1.0+ introduces a control socket that defaults to `$HOME/gunicorn.ctl`. For the pulp user this resolves to `/var/lib/pulp/gunicorn.ctl`, placing a runtime socket inside the persistent data directory.

This PR explicitly sets `--control-socket` for both the API and content services, routing the socket to the `RuntimeDirectory`-managed paths (`/run/pulpcore-api/` and `/run/pulpcore-content/`) that systemd already creates before the service starts.

**Benefits:**
- Keeps runtime sockets in `/run/` where they belong, not in `/var/lib/pulp`
- Avoids SELinux AVC denials — `/run/pulpcore-{api,content}/` is already labeled `pulpcore_server_var_run_t` with existing `sock_file` transitions in the pulpcore SELinux policy
- Paths are exposed as class parameters (`api_control_socket_path`, `content_control_socket_path`) so operators can override if needed

## Test plan

- [ ] Deploy with puppet-pulpcore against gunicorn >= 25.1.0
- [ ] Confirm `/run/pulpcore-api/gunicorn.ctl` and `/run/pulpcore-content/gunicorn.ctl` are created on service start
- [ ] Confirm `/var/lib/pulp/gunicorn.ctl` does **not** appear
- [ ] Run `ausearch -m avc -ts recent` and confirm no denials for `pulpcore_server_t` socket operations
- [ ] Confirm services start and respond normally

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -48,6 +48,12 @@
 # @param content_socket_path
 #   Path where the Pulpcore Content service is listening. This is a unix socket.
 #
+# @param api_control_socket_path
+#   Path for the gunicorn control socket of the API service.
+#
+# @param content_control_socket_path
+#   Path for the gunicorn control socket of the Content service.
+#
 # @param config_dir
 #   Pulp configuration directory. The settings.py file is created under this
 #   directory.
@@ -229,6 +235,8 @@ class pulpcore (
   String[1] $apache_vhost_priority = '10',
   Stdlib::Absolutepath $api_socket_path = '/run/pulpcore-api.sock',
   Stdlib::Absolutepath $content_socket_path = '/run/pulpcore-content.sock',
+  Stdlib::Absolutepath $api_control_socket_path = '/run/pulpcore-api/gunicorn.ctl',
+  Stdlib::Absolutepath $content_control_socket_path = '/run/pulpcore-content/gunicorn.ctl',
   String $postgresql_db_name = 'pulpcore',
   String $postgresql_db_user = 'pulp',
   String $postgresql_db_password = extlib::cache_data('pulpcore_cache_data', 'db_password', extlib::random_password(32)),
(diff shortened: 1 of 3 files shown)


ehelms:
How long has `--control-socket` existed? Does this make puppet-pulpcore only compatible with certain versions?

Odilhao:
> How long has `--control-socket` existed? Does this make puppet-pulpcore only compatible with certain versions?

It started on https://gunicorn.org/2026-news/#2510-2026-02-13 25.1.0, we would need to cherrypick the fix if we ever update gunicorn on 3.85 or early versions.

ehelms:
> It started on https://gunicorn.org/2026-news/#2510-2026-02-13 25.1.0, we would need to cherrypick the fix if we ever update gunicorn on 3.85 or early versions.

Based on this, https://github.com/theforeman/puppet-pulpcore#support-policy needs to be updated and this is a backwards incompatible change for now.

Odilhao:
> > It started on https://gunicorn.org/2026-news/#2510-2026-02-13 25.1.0, we would need to cherrypick the fix if we ever update gunicorn on 3.85 or early versions.
> 
> Based on this, https://github.com/theforeman/puppet-pulpcore#support-policy needs to be updated and this is a backwards incompatible change for now.

Yes, that's why I'm adding as Optional https://github.com/theforeman/puppet-pulpcore/pull/403/changes#diff-c6c3a824d185ff01e063011517a1b13a604dc3d7b63d1c67a30451c10df02bcbR240 so we can prepare to support older version as well, we would need ensure that it's cherrypicked to older versions and that we also support the option on the installer side.

ehelms (approved):

PR #411, opened 2026-07-31
Add temporary CI job to test Pulpcore nightly on EL10

Odilhao:
## Summary
- Adds a standalone `acceptance-el10` job to `.github/workflows/ci.yml` that runs Beaker directly against CentOS Stream 10 with Pulpcore `nightly`, sourced from the [Copr staging repo](https://copr.fedorainfracloud.org/coprs/g/theforeman/pulpcore-nightly-staging/) since nightly isn't published on `yum.theforeman.org` for EL10 yet.
- This intentionally does **not** touch `metadata.json`/`operatingsystem_support`: `metadata2gha` crosses every declared OS with every `beaker_facter` value, so adding EL10 there today would also try (and fail) unsupported combinations like `EL10 + 3.63`.
- This is a temporary, standalone job to validate the module installs on EL10 with nightly Pulpcore packages. Real integration into the supported-OS matrix is tracked in theforeman/el10_rebuild#25.

## Test plan
- [ ] CI: `acceptance-el10` job passes, confirming `pulpcore::repo` + `pulpcore-selinux` install cleanly on CentOS Stream 10 against the nightly Copr staging repo
- [ ] Existing `puppet` job (EL9 matrix) remains unaffected

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,11 +23,15 @@ jobs:
       rubocop: false
 
   # TODO: fold this back into the `puppet` job's beaker_facter matrix once
-  # Pulpcore nightly (and later 3.105) is published for EL10 on
-  # yum.theforeman.org and CentOS/RedHat/AlmaLinux 10 are added to
-  # metadata.json. metadata2gha has no way to run a subset of
-  # beaker_facter values against a subset of operatingsystem_support, so
-  # until then EL10 is tested here against the Copr staging repo instead.
+  # CentOS/RedHat/AlmaLinux 10 are added to metadata.json. metadata2gha has
+  # no way to run a subset of beaker_facter values against a subset of
+  # operatingsystem_support (EL10 will only ever get nightly/3.105+, never
+  # 3.63/3.73/3.85), so until that's resolved EL10 is tested here as a
+  # standalone job instead. See theforeman/el10_rebuild#25.
+  #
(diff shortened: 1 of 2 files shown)


Odilhao:
Pulpcore nightly is now published for EL10 on https://yum.theforeman.org/pulpcore/nightly/el10/, so dropped the Copr staging `baseurl`/`gpgkey` overrides — `pulpcore::repo`'s defaults already resolve the right repo (`dist_tag` comes from `facts['os']['release']['major']`) and disable gpgcheck for `nightly` automatically.

Still keeping this as a standalone job rather than folding into the main `puppet` matrix — that part is still blocked on metadata2gha's lack of per-OS beaker_facter filtering (theforeman/el10_rebuild#25).

ekohl (commented):
If you now just add EL 10 to `metadata.json`, does it work to run things end-to-end? It will probably fail on the older Pulp versions which haven't been built, but we could reduce the supported versions to just nightly to work around that.

Odilhao:
> If you now just add EL 10 to `metadata.json`, does it work to run things end-to-end? It will probably fail on the older Pulp versions which haven't been built, but we could reduce the supported versions to just nightly to work around that.

No it fails, it will try to install all other version with EL10.

Odilhao:
Was created to validate that the module installs cleanly on EL10 with nightly Pulpcore packages, not intended to merge as a standalone job. Validation done — closing without merging. Ref: theforeman/el10_rebuild#25

PR #412, opened 2026-08-03
Assert valkey service on EL10 in redis acceptance test

Odilhao:
## Summary
- RHEL/CentOS Stream 10 dropped Redis in favor of Valkey (a Redis-protocol-compatible fork). `puppet-redis` already handles this correctly on EL10 by installing `Package[valkey]`/`Service[valkey]` instead of `Package[redis]`/`Service[redis]`.
- `spec/acceptance/basic_spec.rb` hardcoded the service name `redis`, which fails on EL10 even though the underlying install works correctly.
- This makes the assertion OS-conditional: `valkey` on RedHat-family OSes release >= 10, `redis` otherwise.

Split out from #411 so this fix can land independently of the temporary EL10 CI job in that PR.

## Test plan
- [ ] Existing EL9 acceptance runs continue to assert `service('redis')` unchanged
- [ ] EL10 acceptance runs (see #411) assert `service('valkey')` and pass

### DIFF
--- a/spec/acceptance/basic_spec.rb
+++ b/spec/acceptance/basic_spec.rb
@@ -33,7 +33,12 @@ describe 'with content cache enabled' do
     it { is_expected.to be_listening }
   end
 
-  describe service('redis') do
+  # RHEL/CentOS Stream 10 dropped Redis in favor of Valkey (a
+  # Redis-protocol-compatible fork); puppet-redis installs Package[valkey]/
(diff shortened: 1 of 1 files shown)


Odilhao:
Breaking this into another PR, so we can merge it for us to release with forklift and have the rpms published. 

The valkey test worked at https://github.com/theforeman/puppet-pulpcore/pull/411

ekohl (approved):

